PR #57, opened 2014-12-23
add a test to Trinary that numbers that are not valid trinary convert to...

stevejb71:
... 0, following the readme.


etrepum:
Thank you!

Issue #58, opened 2015-01-04
nucleotide-count tests don't enforce valid strand constraint


mambocab:
In [the tests for nucleotide-count](https://github.com/exercism/xhaskell/blob/master/nucleotide-count/nucleotide-count_test.hs), this test case:

```haskell
assertError "invalid nucleotide 'X'" $ count 'X' "GACT"
```

ensures that the implementation checks that the nucleotide being counted is valid. However, there is no such test for the strand whose nucleotides are being counted, for `count` or `nucleotideCounts`.

The [example implementation](https://github.com/exercism/xhaskell/blob/master/nucleotide-count/example.hs) does check for this, so I thought it might be a missing test case.

etrepum:
Sounds like a reasonable thing to do, want to submit a pull request?

mambocab:
Happy to.

PR #60, opened 2015-01-09
nucleotide count: adds strand validation to tests

mambocab:
This pull request addresses #58.

The previous tests asserted that, e.g., `count 'X' "ACGT"` would throw an error indicating that `'X'` is not a valid nucleotide. The example implementation performed the same error checking on the strand being counted and on the argument to `nucleotideCounds`, but the tests did not check for this behavior. This PR implements tests that do.


etrepum:
Looks great, thanks!

PR #75, opened 2015-04-16
Adds an implementation of POV

alexkalderimis:
See exercism/x-common#70


etrepum:
I think this might need to be rebased or something, the Merge pull request button is disabled.

kytrinyx:
Yeah, I would suggest rebasing onto master, and then squashing the commits so that we get an atomic change. If you need help with the details of that holler.

alexkalderimis:
@kytrinyx were you asking me to do that? If so, that's fine - I will see to it shortly.

kytrinyx:
@alexkalderimis -- it's easiest if you do it, since it's on your branch, but if you don't have time we're always happy to fix it by resolving it locally.

alexkalderimis:
Rebased as requested.

etrepum:
Thanks! :+1:

PR #78, opened 2015-04-29
Improved the tests by adding labels for each attack

alexkalderimis:
This pull request makes some minor improvements to the queen-attack tests by labelling each of the attacks so that failures can be diagnosed.

### DIFF
--- a/queen-attack/.gitignore
+++ b/queen-attack/.gitignore
@@ -1,0 +1,1 @@
+Queens.hs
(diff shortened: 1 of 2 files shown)


etrepum:
:+1:

Did you know that you can run tests for an individual exercise by specifying it as a parameter to `./_test/check-exercises.hs`?

https://github.com/exercism/xhaskell/blob/master/_test/check-exercises.hs#L62-L64

kytrinyx:
Maybe this would be worth mentioning in the SETUP.md.

etrepum:
It's only relevant to folks working on the repo, and it is documented in the README https://github.com/exercism/xhaskell#running-tests

kytrinyx:
Ah, right. Thanks!

PR #79, opened 2015-04-30
Improved tests to aid in debugging.

alexkalderimis:
Added labels for tests that are difficult to distinguish by result,
added examples from the readme so implementers can see how the readme
relates specifically, and added a top-level label.

These changes are solely to make it easier to identify failing conditions, the existing example passes.


etrepum:
Thanks!

PR #84, opened 2015-06-28
Add lens-person exercise

pminten:
This is a relatively easy exercise to play around with basic lenses. It's designed in such a way that lenses aren't actually required, so you can play around with different approaches and see what works best.

@etrepum I haven't made an x-common part for this yet since I wanted to get your opinion on this first. Do you think this is a good exercise for the Haskell track? It does kinda deviate from standard exercism practice by a) (strongly) suggesting the use of non-standard-library libraries and b) being clearly applicable only to Haskell.

### DIFF
--- a/_test/bootstrap.sh
+++ b/_test/bootstrap.sh
@@ -23,4 +23,5 @@ cabal install \
       vector \
       parallel \
       stm \
-      old-locale
+      old-locale \
+      lens
--- a/config.json
+++ b/config.json
@@ -72,6 +72,7 @@
     "go-counting",
     "zipper",
     "forth",
+    "lens-person",
     "pov"
   ],
   "deprecated": [
--- a/lens-person/Person.hs
+++ b/lens-person/Person.hs
@@ -1,0 +1,45 @@
+module Person where
+
+import           Data.Time.Calendar
+
+data Person = Person {
+      _name    :: Name,
+      _born    :: Born,
+      _address :: Address
+    }
+
+data Name = Name {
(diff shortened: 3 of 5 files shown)


kytrinyx:
I think language-specific exercises are fine (they should still have an entry in the x-common, though perhaps with something in the config yml that says that it was conceived specifically for the haskell track.

I'm going to be working with some students this summer who will hopefully start using the metadata in x-common to start showing a bit of information about exercises on the site itself.

pminten:
Build fails on lack of install lens lib. Anyone have objections to adding that to _test/bootstrap.sh?

pminten:
@kytrinyx That's great. My experience is that the exercises which are most useful are often the more language (or at least paradigm) specific ones. The "translate the exercise" approach, practical as it may be, does tend to result in shoehorning.

etrepum:
No objection to adding it to the bootstrap. I think having exercises designed for third party libraries is a good thing, particularly if they're de facto standards like lens. The Haskell world has changed in the past few years with cabal sandboxes, stackage and stack in a way that makes Haskell Platform much less relevant. We already depend on a number of packages that don't ship with GHC itself, so as long as it's clearly documented what the exercise needs I think it's fine.

pminten:
Ok then, I'll add this tomorrow (don't have time / the right computer right now).

etrepum:
:+1:

pminten:
Build fixed, once the x-common PR is merged this one is ready to be merged here as well.

etrepum:
Thanks! :+1:

PR #86, opened 2015-12-13
Add ABOUT.md, LEARNING.md, and RESOURCES.md

outkaj:
Adding information that was missing on [the Haskell section of the website] (http://exercism.io/languages/haskell).

### DIFF
--- a/docs/LEARNING.md
+++ b/docs/LEARNING.md
@@ -1,0 +1,1 @@
+Please see the "Useful Haskell Resources" tab. 
--- a/docs/RESOURCES.md
+++ b/docs/RESOURCES.md
@@ -1,0 +1,13 @@
+##Recommended Learning Resources
+
+* [The Haskell website](www.haskell.org) and [the Haskell wiki] (www.wiki.haskell.org) are great resources for getting started with Haskell as well as diving deeper into specific features of the language.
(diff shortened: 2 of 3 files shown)


docs/RESOURCES.md
```diff
@@ -0,0 +1,13 @@
+##Recommended Learning Resources
```
kytrinyx: The resources page already has a header, so this would make a duplicate.

docs/LEARNING.md
```diff
@@ -0,0 +1 @@
+Please see the "Useful Haskell Resources" tab.
```
kytrinyx: Learning should be about learning the language from scratch, whereas resources is more about tools, references, and resources that you might use while developing in Haskell.

kytrinyx:
I'd like to have something about what kinds of problems Haskell is used to solve in the `ABOUT.md`.

rbasso:
This PR is inactive for 6 months and most of the proposed content was, meanwhile, included in the repository by #93 and #97.

Considering that...

- any old changes would have to be completely rewritten to merge with the current *docs*
- the content proposed is similar to what we currently have
- half an year has elapsed without answer from the author.

... I'll close this PR.

PR #89, opened 2016-01-23
largest-series-product corner cases

petertseng:
As part of the work to have consistency between tracks as in https://github.com/exercism/todo/issues/104 and https://github.com/exercism/todo/issues/13. Comments are in individual commit messages.

### DIFF
--- a/largest-series-product/example.hs
+++ b/largest-series-product/example.hs
@@ -9,7 +9,7 @@ slices :: Integral a => Int -> String -> [[a]]
 slices n = go . digits
   where go xs = map (take n) $ take (length xs - pred n) (tails xs)
 
-largestProduct :: Integral a => Int -> String -> a
+largestProduct :: Integral a => Int -> String -> Maybe a
 largestProduct n text = case map product (slices n text) of
-  []       -> 1
-  products -> maximum products
+  []       -> Nothing
+  products -> Just $ maximum products
--- a/largest-series-product/largest-series-product_test.hs
+++ b/largest-series-product/largest-series-product_test.hs
@@ -49,6 +49,7 @@ seriesTests = map TestCase
   , int 1 @=? largestProduct 0 "123"
   , intNothing @=? largestProduct 1 ""
   , intNothing @=? largestProduct 4 "123"
-    -- edge case :)
+    -- if all spans contain zero, result is zero.
+  , int 0 @=? largestProduct 3 "99099"
   , int 0 @=? largestProduct 2 "00"
   ]
--- a/largest-series-product/largest-series-product_test.hs
+++ b/largest-series-product/largest-series-product_test.hs
@@ -18,8 +18,11 @@ main = exitProperly $ runTestTT $ TestList
 ints :: [Int] -> [Int]
 ints = id
 
-int :: Int -> Int
-int = id
+int :: Int -> Maybe Int
+int = Just
+
+intNothing :: Maybe Int
+intNothing = Nothing
 
 seriesTests :: [Test]
 seriesTests = map TestCase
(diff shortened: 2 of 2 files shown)


petertseng:
Ah, that's unfortunate. Tests are failing on `-Werror` because of those `Nothing`s. Going to think about how to deal with that

petertseng:
Yeah ran locally with -Werror to check, this should be good now. had to add a Nothing that was explicitly `Maybe Int`

petertseng:
Going to change that `int` to be `Int -> Maybe Int` to avoid all the `Just` business going on here, I think.

petertseng:
OK, that's better. Much less churn doing it this way.

kytrinyx:
I'm not sure who's around these days who knows Haskell. It looks like @abo64, @NobbZ and @stevejb71 have all commented recently in the Haskell track... would one of you have a moment to review this?

NobbZ:
I'll take a closer look into this in the next 36 hours.

Katrina Owen <notifications@github.com> schrieb am Fr., 29. Jan. 2016 19:58:

NobbZ:
As far as I can see, the code in the example is idiomatic, the choose of `Maybe` is reasonable. Only thing I fear is, that someone new to this exercise might fail miserably because he has to return `Just x` instead of just `x` ;) I am not aware of how often `Maybe` is used in the exercises before this already.

Also all the added tests (and the changes to make old ones pass with new code) seem reasonable.

petertseng:
Reasonble question. I originally tried to answer this question by seeing what has `Maybe` in it

```
$ git grep -l Maybe | cut -d/ -f1 | uniq
_test
atbash-cipher
bank-account
binary-search-tree
binary
connect
go-counting
hexadecimal
largest-series-product
linked-list
pov
queen-attack
saddle-points
say
scrabble-score
sgf-parsing
wordy
zipper
```

If someone follows the order in config.js, then the ones that come before: scrabble-score, binary, atbash-cipher, bank-account, queen-attack, binary-search-tree, hexadecimal

However, I actually realized that it's not necessarily the case that the tests are testing for `Maybe` -- it could just be an implementation detail of the examples. Indeed, this is true for a lot of them. So I manually look through the list and see which of them have tests involving `Maybe` (which means the student would have to write code involving `Maybe`)

* The queen-attack has a test where it passes in `Just` a board position or `Nothing`.
* The binary-search-tree problem needs to have a `bstLeft` and `bstRight` that return `Maybe (BST a)`.
* bank-account needs `Maybe Int` for its bank balance. It also seems to require using the `IO` monad, an interesting choice.

So someone on this exercise either has seen `Maybe` in a previous exercise (or is skipping around, but in that case they on their own anyway). So I think the specific question of "will someone be confused at the `Maybe` in here?" is "I think we are OK"

Separately, we may consider somewhere to discuss general questions that rise from the above comment. Possible questions I can think of:

* Should we have exercises designed to ease learners of the language into various concepts of the language? In other words, to what extent shall Exercism try to act as an explicit language tutorial of sorts (rather than just providing an impetus to apply what is learned from an external tutorial)?
* Should there be a limit to the number of new language concepts introduced in each problem?
* To what extent should our test suites test "the fiddly bits" (dealing with errors, etc.) rather than the algorithmic "meat" of the problem?

petertseng:
Y'all think anything else needs to be done here? Based on the investigation above, I think the `Maybe` thing is covered earlier so will not be an obstacle, and y'all seem to think everything else is good.

NobbZ:
I do also think that this can be merged. @kytrinyx

kytrinyx:
Thanks, merging!

PR #90, opened 2016-01-24
Add learning.MD

ahubers:
Edit text however you like. The key is linking to the Haskell learning Wiki, which (in my opinion) is already an excellent resource center for beginning Haskell.


kytrinyx:
Thank you!

PR #94, opened 2016-02-16
Added attribution link to the source of the quicksort example in ABOUT.md

suetanvil:
I realized after the fact that I'd gotten the Quicksort example from "Learn You A Haskell", so I added a link.


kytrinyx:
Thanks!

PR #96, opened 2016-03-03
gigasecond: fix use of deprecated package

daniloisr:
Apparently `System.Locale` get deprecated on haskell 7.10 and moved to the package `old-locale-1.0.0.7`.

And this change will use reuse the function provided by `Data.Time.Format`

Ref: https://github.com/exercism/xhaskell/issues/85#issuecomment-191382840


kytrinyx:
Thanks for catching this!

PR #98, opened 2016-03-13
largest-series-product: Do not test digits/slices

petertseng:
The digits and slices functions are internal implementation details and
thus the test case for the largest-series-product problem should not be
concerned with testing them. The series tests could potentially be moved
to the series exercise (already implemented by this track).

Their presence may cause students to falsely think that their solution
has to use these two functions, instead of the alternative
implementation of only iterating through the digits once.

If it is desired to give hints on how to approach this problem (which
was one advantage of having the digits and slices tests), then consider
including a hints file and/or directory in the largest-series-product
directory.

This PR arises from discussion in
https://github.com/exercism/x-common/issues/192

### DIFF
--- a/largest-series-product/example.hs
+++ b/largest-series-product/example.hs
@@ -1,4 +1,4 @@
-module Series (digits, slices, largestProduct) where
+module Series (largestProduct) where
 import Data.Char (digitToInt)
 import Data.List (tails)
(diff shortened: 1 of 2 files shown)


kytrinyx:
Thanks!

PR #101, opened 2016-03-23
sum-of-multiples: Explicitly pass [3, 5] instead of defaulting

### LINKED ISSUE #100: Pass explicit list of multiples in "Sum of Multiples" exercise rather than defaulting to 3 and 5
Hello, as part of https://github.com/exercism/x-common/issues/198 we'd like to make the sum of multiples exercise less confusing. Currently, the [README](https://github.com/exercism/x-common/blob/master/sum-of-multiples.md) specifies that if no multiples are given it should default to 3 and 5.

We'd like to remove this default, so that a list of multiples will always be specified by the caller. This makes the behavior explicit, avoiding surprising behavior and simplifying the problem.

**Please make sure this track's tests for the sum-of-multiples problem do not expect such a default**. Any tests that want to test behavior for multiples of [3, 5] should explicitly pass [3, 5] as the list of multiples.

After all tracks have completed this change, then https://github.com/exercism/x-common/pull/209 can be merged to remove the defaults from the README.

The reason we'd like this change to happen before changing the README is that it was very confusing for students to figure out the default behavior. It wasn't clear from simply looking at the tests that the default should be 3 and 5, as seen in https://github.com/exercism/exercism.io/issues/2654, so some had to resort to looking at the example solutions (which aren't served by `exercism fetch`, so they have to find it on GitHub). It was added to the README to fix this confusion, but now we'd like to be explicit so we can remove the default line from the README.

You can find the common test data at https://github.com/exercism/x-common/blob/master/sum-of-multiples.json, in case that is helpful.

petertseng:
Closes #100

### DIFF
--- a/sum-of-multiples/example.hs
+++ b/sum-of-multiples/example.hs
@@ -1,7 +1,4 @@
-module SumOfMultiples (sumOfMultiples, sumOfMultiplesDefault) where
-
-sumOfMultiplesDefault :: Int -> Int
-sumOfMultiplesDefault = sumOfMultiples [3, 5]
+module SumOfMultiples (sumOfMultiples) where
 
 sumOfMultiples :: [Int] -> Int -> Int
 sumOfMultiples targets upperBound = sum (filter f [1..upperBound-1])
--- a/sum-of-multiples/sum-of-multiples_test.hs
+++ b/sum-of-multiples/sum-of-multiples_test.hs
@@ -1,6 +1,6 @@
 import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
 import System.Exit (ExitCode(..), exitWith)
-import SumOfMultiples (sumOfMultiples, sumOfMultiplesDefault)
+import SumOfMultiples (sumOfMultiples)
 
 exitProperly :: IO Counts -> IO ()
 exitProperly m = do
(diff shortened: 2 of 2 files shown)


petertseng:
This does not bode well. The build fails for `cabal`-related reasons, reasons completely unrelated to my change (which works fine when tested locally). As you can see, the [build for the merge of #98](https://travis-ci.org/exercism/xhaskell/builds/117801650) failed, despite [the build for #98](https://travis-ci.org/exercism/xhaskell/builds/115607987) passing, and I'm pretty sure that merge would have been a fast-forward. Will need a cabal fix, then I can rebase to get that fix.

petertseng:
I think this is because [cabal packages are cached](https://github.com/exercism/xhaskell/blob/master/.travis.yml#L19) and [clearing them](https://docs.travis-ci.com/user/caching/) will solve this.

petertseng:
Definitely a cache-related, since https://travis-ci.org/exercism/xhaskell/builds/117969333 on https://github.com/exercism/xhaskell/pull/102/commits/77dcf3cc88c2f5d3930b306a875fa0fe2ae53715 succeeded

Let's see if I can't do something to help here...

petertseng:
Good, I can cause the cache to get cleared on this branch, so that makes tests pass for this PR. I suspect master will remain broken until the cache gets cleared.

kytrinyx:
What would the incantation be to clear the cache on master?

kytrinyx:
I've deleted the caches for master.

PR #102, opened 2016-03-23
DO NOT MERGE: maybe clear cabal cache

petertseng:


petertseng:
I probably should have done this on my own repo rather than a PR here. But actually I'm pretty sure my own repo doesn't use the same cache as this repo, and since I think I'm trying to deal with cache things here, maybe it makes sense.

petertseng:
I've seen what I needed to see.

PR #103, opened 2016-04-01
move exercises to exercises/ dir

### LINKED ISSUE #88: Move exercises to subdirectory
The problems api (x-api) now supports having exercises collected in a subdirectory
named `exercises`.

That is to say that instead of having a mix of `bin`, `docs`, and individual exercises,
we can have `bin`, `docs`, and `exercises` in the root of the repository, and all
the exercises collected in a subdirectory.

In other words, instead of this:

```
x{TRACK_ID}/
├── LICENSE
├── README.md
├── bin
│   └── fetch-configlet
├── bowling
│   ├── bowling_test.ext
│   └── example.ext
├── clock
│   ├── clock_test.ext
│   └── example.ext
├── config.json
└── docs
│   ├── ABOUT.md
│   └── img
... etc
```

we can have something like this:

```
x{TRACK_ID}/
├── LICENSE
├── README.md
├── bin
│   └── fetch-configlet
├── config.json
├── docs
│   ├── ABOUT.md
│   └── img
├── exercises
│   ├── bowling
│   │   ├── bowling_test.ext
│   │   └── example.ext
│   └── clock
│       ├── clock_test.ext
│       └── example.ext
... etc
```

This has already been deployed to production, so it's safe to make this change whenever you have time.

petertseng:
`ls -1 exercises | wc -l` outputs 70, which matches the current number
of Haskell exercises, so we have some confidence this is being done
correctly.

Closes #88


kytrinyx:
Indeed, and the tests are passing, so it looks like that fix is good also.

PR #107, opened 2016-04-02
[DO NOT MERGE] pov: Add alternate expected orderings for rerooted trees

petertseng:
See individual commit messages.

This is a possible solution but I feel a little uncomfortable explicitly listing out all the possibilities (and an issue I'm sweeping under the rug is that there are actually four possibilities for the `cousins` case because the reordering question aplies to both (`x` and its parents/children) and (`parent` and its parents/children).

So this is a stopgap solution if we can't come up with something better, but I think I may be more comfortable just comparing the sorted edge list as xgo does. I will explore that solution soon, so let's not merge this until I've completed that exploration. Feel free to review in the meantime though.


petertseng:
I prefer #108 over this so much that I'm confident #108 is the better solution and this should not be merged. Feel free to discuss if you disagree and if I am convinced I'm happy to reopen.

PR #108, opened 2016-04-02
pov: Test sorted edge list, not exact graph order

### LINKED ISSUE #92: pov: Consider making tests insensitive to child ordering
In the [pov tests](https://github.com/exercism/xhaskell/blob/master/pov/pov_test.hs), given a tree that must be reparented on node `x`, the tests ensure that the former parent of `x` comes *after* the former children of `x`.

I think it would be beneficial if the tests did not care about the ordering of the children, since trees probably don't care about the ordering of their children, and it almost certainly is valid to allow either order in the POV problem.

petertseng:
Given a tree that must be reparented on node `x`, the tests used to
ensure that the former parent of `x` comes *after* the former children
of `x`. But there's no reason this should be the case, since in trees we
don't particularly care about the ordering of the children. (There are
some specialised trees for which it is the case, but this is not one of
those trees)

By comparing the sorted edge list, we allow either ordering.

Tested by ensuring that the example solution passes (it puts former
parent after former children), and that it still passes when using the
opposite ordering by modifying `(children g) ++ [reparented]` to instead
be `reparented : children g`.

Also tested to ensure that solutions that always return `Nothing` and
`Just (Graph x [])` return reasonable error messages.

Always return `Nothing`:

    ### Failure in: 0:5:"reparenting flat: correct root"
    pov_test.hs:76
    expected: Just "x"
     but got: Nothing
    ### Failure in: 0:6:"reparenting flat: correct sorted edges"
    pov_test.hs:76
    expected: Just [("root","a"),("root","b"),("root","c"),("x","root")]
     but got: Nothing

Always return `Just (Graph x [])`:

    ### Failure in: 0:6:"reparenting flat: correct sorted edges"
    pov_test.hs:76
    expected: Just [("root","a"),("root","b"),("root","c"),("x","root")]
     but got: Just []

Fixes #92

### DIFF
--- a/exercises/pov/pov_test.hs
+++ b/exercises/pov/pov_test.hs
@@ -1,4 +1,5 @@
 import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
+import Data.List (sort)
 import System.Exit (ExitCode(..), exitWith)
 import POV(Graph(..), fromPOV, tracePathBetween)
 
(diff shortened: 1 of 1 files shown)


petertseng:
I prefer this over #107 by far.

Hold on though, need to add root check.

PR #109, opened 2016-04-02
rename point-mutations to hamming

### LINKED ISSUE #95: "point-mutations" is deprecated in favor of hamming
This happened a while back, and it was for really weird legacy reasons.

I've since fixed the underlying issues that caused the problem, but for consistency
it would be nice to rename point-mutation to hamming, so that all the tracks are using
the same exercise name.

Once the problem has been renamed, I can run a script on the website to point people's
existing point-mutations solutions to the new hamming exercise so that they'll be able
to review solutions to hamming, and people who solve the new hamming exercise can see
all the old ones.

petertseng:
point-mutations was deprecated in favor of hamming a while ago (to make
it more generic, since it really was just about Hamming distance rather
than specifically about biological applications)

Since the module is now named Hamming, the function hammingDistance is
renamed to distance so as to not be repetitive (Hamming.distance versus
Hamming.hammingDistance)

Fixes #95

### DIFF
--- a/exercises/hamming/example.hs
+++ b/exercises/hamming/example.hs
@@ -1,0 +1,4 @@
+module Hamming (distance) where
+
+distance :: String -> String -> Int
+distance a b = sum . map fromEnum $ zipWith (/=) a b
--- a/config.json
+++ b/config.json
@@ -13,7 +13,7 @@
     "bob",
     "sum-of-multiples",
     "strain",
-    "point-mutations",
+    "hamming",
     "space-age",
     "anagram",
     "nucleotide-count",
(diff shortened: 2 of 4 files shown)


exercises/hamming/example.hs
```diff
@@ -1,4 +1,4 @@
-module DNA (hammingDistance) where
+module Hamming (hammingDistance) where
```
kytrinyx: Should this be just `distance` instead of `hammingDistance` to avoid the echo with the module name?

petertseng: give me a bit to make sure this is something Haskell does, must research

petertseng: going to say yes to this. see how https://hackage.haskell.org/package/containers-0.5.7.1/docs/Data-Set.html and https://hackage.haskell.org/package/containers-0.5.7.1/docs/Data-Map-Lazy.html just have a `size` instead of, say, `setSize` or `mapSize` or something. Coming soon.

petertseng:
`hammingDistance` is hereby renamed to `distance`

kytrinyx:
Lovely!

PR #118, opened 2016-04-07
travis: 7.10.1 -> 7.10.3

### LINKED ISSUE #117: travis: 7.10.1 -> 7.10.3?
7.10.1 was in March 2015 https://www.haskell.org/ghc/download_ghc_7_10_1
Since then we've had:

* 7.10.2 https://www.haskell.org/ghc/download_ghc_7_10_2 (release notes at http://downloads.haskell.org/~ghc/7.10.2/docs/html/users_guide/release-7-10-2.html)
* 7.10.3 https://www.haskell.org/ghc/download_ghc_7_10_3 (release notes at http://downloads.haskell.org/~ghc/7.10.3/docs/html/users_guide/release-7-10-3.html)

My impression is that 7.10.1 was chosen because it was the newest at the time. Now that 7.10.3 is out I think we'd like to test that now. This will be the case unless we specifically care about some distro that stays at 7.10.1 and does not move to 7.10.3. That would be odd behavior since it's my impression that incrementing just the least significant number means it's a bugfix release (GHC seems to follow this convention at least)

I wonder if there's a way to specify 7.10.x to travis to always stay up to date... (current suspicion on that is "no" looking at some prominent haskell projects that test multi haskell versions, try lens for example)

petertseng:
Closes #117

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ addons:
     packages:
       - ghc-7.6.3
       - ghc-7.8.4
-      - ghc-7.10.1
+      - ghc-7.10.3
       - cabal-install-1.22
 env:
   global:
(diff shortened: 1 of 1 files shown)


petertseng:

PR #120, opened 2016-04-07
travis: Test ghc 8.0.1 (and avoid some compiler warnings on 8.0.1)

petertseng:
It's a prerelease version, but it's likely to be good to start testing
against it now.

Two commits are needed to avoid compiler warnings on 8.0.1 (and thus errors since we use `-Werror`). The errors that would appear are noted in the commit message.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ addons:
       - ghc-7.6.3
       - ghc-7.8.4
       - ghc-7.10.3
+      - ghc-8.0.1
       - cabal-install-1.22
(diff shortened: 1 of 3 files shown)


petertseng:

petertseng:
will rebase in next 24 hours. I can and should be prodded if not.

petertseng:
I think github automatically sends notifications when I push new commits to a PR branch now, so notifications should have been sent. let me know whether that's true so I know whether I can safely stop notifying via comment

PR #128, opened 2016-04-09
travis: Install only the GHC we are testing

### LINKED ISSUE #119: travis: all builds install all GHC versions, rather than just the one they're testing
For example, the 7.6 build will install the 7.8 ghc and the 7.10 ghc. This may unnecessarily increase install times.

This can be avoided by specifying addons in the matrix, rather than on the top level.

petertseng:
This will decrease install times

Closes #119

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,23 +6,35 @@
 language: c
 sudo: false
 addons:
-  apt:
+  apt: &apt
     sources:
       - hvr-ghc
-    packages:
-      - ghc-7.6.3
-      - ghc-7.8.4
-      - ghc-7.10.3
-      - ghc-8.0.1
-      - cabal-install-1.22
+cabal: &cabal cabal-install-1.22
 env:
   global:
     - CABALVER=1.22
-  matrix:
-    - GHCVER=7.6.3
-    - GHCVER=7.8.4
-    - GHCVER=7.10.3
-    - GHCVER=8.0.1
+matrix:
+  include:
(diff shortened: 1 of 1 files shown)


petertseng:

petertseng:
Hmm. Adds a bit of complexity (yaml references) to the yaml file. Saves about 8 seconds out of an 80 second build (if there's a cache hit) or a 5 minute build (if there's a cache miss). Though we run tests for every PR so I suppose it's good to save all the time we can. Plus, the savings are greater if we test more GHC versions (for example 8.0 as in #120, maybe even 7.4...)

petertseng:
will rebase in next 24 hours. I can and should be prodded if not.

petertseng:
I think github automatically sends notifications when I push new commits to a PR branch now, so notifications should have been sent. let me know whether that's true so I know whether I can safely stop notifying via comment

rbasso:
I received the two notifications from your messages.

But don't worry! I'm checking it frequently.

petertseng:
I'm just going to take a moment to say that I kept the cabal version at 1.22 across the board since this preserves existing behavior. I would note however that https://github.com/hvr/multi-ghc-travis/blob/master/make_travis_yml.hs has a map of GHC version to the cabal version at the time (well, excepting anything before 1.16, but that doesn't affect us) and usually tests the two together.

I don't have a strong opinion on what cabal version should be used, so I'll leave the existing behavior untouched unless there's a compelling reason to change.

PR #131, opened 2016-04-11
DO NOT MERGE: travis: MORE VERSIONS

petertseng:


petertseng:
7.4 currently broken. Dependencies install fine, but:

```
_test/check-exercises.hs:5:27:
    Module `System.Posix.Temp' does not export `mkdtemp'
```

It is possible we may be able to fix it by updating the `unix` package.

7.2 can't resolve deps:

```
cabal: Could not resolve dependencies:
trying: text-1.2.2.1 (user goal)
trying: binary-0.5.0.2/installed-f86... (dependency of text-1.2.2.1)
trying: attoparsec-0.13.0.1 (user goal)
trying: scientific-0.3.4.6 (dependency of attoparsec-0.13.0.1)
rejecting: scientific-0.3.4.6:+bytestring-builder (conflict: binary =>
bytestring==0.9.2.0/installed-fe0..., scientific-0.3.4.6:bytestring-builder =>
bytestring>=0.10 && <0.11)
rejecting: scientific-0.3.4.6:-bytestring-builder (manual flag can only be
changed explicitly)
Backjump limit reached (change with --max-backjumps)
```

7.0 has a similar error for lens that I'm not posting all of because it's quite long.

6.12.3 has a bit of a weird error message.


```
cabal: dependencies conflict: ghc-6.12.3 requires an installed instance of
bytestring ==0.9.1.7 however:
```

it just stops there?

petertseng:
A valiant effort to support 7.4 but ultimately unsuccessful. Feel free to rerun the experiments at a later date if we change our set of dependencies (notably how `check-exercises` creates temporary directories), but not a battle I'm fighting now.

PR #133, opened 2016-04-22
binary-search-tree: Allow empty trees

petertseng:
Since #59 mentioned that it was surprising that binary search trees in this exercise couldn't be empty (since `bstValue` was of type `BST a -> a` rather than `BST a -> Maybe a`) I explored what it would be like if it was allowed.

Unfortunately, now that `data BST` has two variants and is no longer a record, we no longer get "free" implementations of `bstValue`, `bstLeft`, and `bstRight`. However, I found this better than the alternatives:

* `newtype BST a = BST (Maybe (Node a))` means you have to litter `fromBST` and `BST` in various places to convert to/from the `newtype`. I didn't particularly like this, but you can examine this at [my bst-empty-newtype branch](https://github.com/exercism/xhaskell/compare/master...petertseng:bst-empty-newtype?expand=1)
* `type BST a = Maybe (Node a)` works interestingly, but then I can't do the `Foldable` thing done here since the `Foldable` for `Maybe` didn't make sense for the `BST`, but `type` makes them indistinguishable.

As alluded to, since #59 mentions monoids, I made the example tree be an instance of `Foldable` so that the implementation of `toList` comes for free (along with other functions like `sum` and `product`)

See individual commit messages for more detail.

### DIFF
--- a/exercises/binary-search-tree/binary-search-tree_test.hs
+++ b/exercises/binary-search-tree/binary-search-tree_test.hs
@@ -20,28 +20,28 @@ int4 = 4
 bstTests :: [Test]
 bstTests =
   [ testCase "data is retained" $
-    4 @=? bstValue (singleton int4)
+    Just 4 @=? bstValue (singleton int4)
   , testCase "inserting less" $ do
     let t = insert 2 (singleton int4)
-    4 @=? bstValue t
-    Just 2 @=? bstValue `fmap` bstLeft t
+    Just 4 @=? bstValue t
+    Just 2 @=? (bstLeft t >>= bstValue)
   , testCase "inserting same" $ do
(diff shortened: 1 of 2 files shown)


kytrinyx:
@towynlin You mentioned that you have both haskell experience and an interest in taking a look at things here on Exercism. Do you have a moment to review this?

towynlin:
Yep, I'll take a look later today. Thanks for the contribution @petertseng, and thanks for pinging me @kytrinyx.

towynlin:
Looks great to me! 👍 Thanks again for the detailed, thoughtful work @petertseng!

@kytrinyx I recommend merging the PR and closing #59.

kytrinyx:
Thanks @towynlin and @petertseng!

PR #134, opened 2016-04-25
Add edge case test for defaults in sum-of-multiples

petemcfarlane:
The README.md says 
> Write a program that can find the sum of the multiples of a given set of numbers. If no set of numbers is given, default to 3 and 5.

I have updated the tests (and example) to add this edge case.

### DIFF
--- a/exercises/sum-of-multiples/sum-of-multiples_test.hs
+++ b/exercises/sum-of-multiples/sum-of-multiples_test.hs
@@ -35,4 +35,8 @@ sumOfMultiplesTests =
     2203160 @=? sumOfMultiples [43, 47] 10000
   , testCase "[5, 25] 51" $
     275 @=? sumOfMultiples [5,25] 51
+  , testCase "[1] 100" $
+    4950 @=? sumOfMultiples [1] 100
+  , testCase "[] 10000" $
+    0 @=? sumOfMultiples [] 10000
   ]


petertseng:
Hello, and thanks for helping out! We're in the process of removing those defaults though, sorry! They were too confusing, so they were removed in #101, as part of the efforts of https://github.com/exercism/x-common/issues/198

petertseng:
Hello, and thanks for helping out! We're in the process of removing those defaults though, sorry! They were too confusing, so they were removed in #101, as part of the efforts of https://github.com/exercism/x-common/issues/198

Once we're done with that, then the README can be edited with https://github.com/exercism/x-common/pull/209.

What I think would be good here would be one of two things:

1. Check https://github.com/exercism/x-common/blob/master/sum-of-multiples.json to see if there are any cases in there that Haskell is missing. If there are, then it would be great to add them!
2. Check the issues linked-to from https://github.com/exercism/x-common/issues/198 to see which tracks haven't yet removed their defaults, and help them out by submitting PRs to no longer use defaults. Then we'll be able to remove the defaults sentence from the README.

petemcfarlane:
Hi @petertseng, sorry I hadn't seen that PR! I'll make a new PR to update the test cases.

kytrinyx:
Thanks @petemcfarlane -- and sorry about the confusion. Still trying to figure out how to manage things like this that affect a bunch of different repositories that exist on slightly different time scales.

PR #137, opened 2016-05-17
Remove errant .gitignore file

ozan:
I imagine this file is here accidentally. It caused confusion for me when working on the corresponding exercise as I have my solutions in a git repo.

PR #138, opened 2016-05-18
Remove two more errant .gitignore files

ozan:
Like #137, removes two more (the only two other) misplaced .gitignore files.

Sorry for not being thorough the first time :)


kytrinyx:
Not at all, thank you so much for catching these!

PR #139, opened 2016-05-28
gitignore: rm beam

petertseng:
beam is for Erlang, which has nothing to do with Haskell.

First discovered in
https://github.com/exercism/xrust/pull/129#discussion_r64982860

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-*.beam
 *.swp
 .DS_Store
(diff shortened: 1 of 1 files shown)


kytrinyx:
Agreed. I think that if people submit changes to the gitignore for files that are not directly relevant to the language in question we should encourage them to update their own global gitignore file.

PR #142, opened 2016-06-06
added test case for isosceles triangle

beane:
The existing tests missed a case where the two smaller sides were equal. A lot of solutions involved sorting the input, so this matters.


rbasso:
Sorry for making you wait, @beane !

PR #144, opened 2016-06-08
Add -fno-warn-type-defaults to supress unwanted warnings.

### LINKED ISSUE #143: Warning "Defaulting the following constraints to type `Integer` in leap_test.hs
The years in leap_test.hs have no explicit type, allowing solutions with differente type signatures:

```haskell
isLeapYear :: Int     -> Bool
isLeapYear :: Integer -> Bool
```

This is desirable, but has the unintended effect of displaying a warning...

```
Defaulting the following constraint(s) to type `Integer` arising from the use of `isLeapYear` at leap_test.hs ... arising from the literal 1996 ...
```

... when compiling `leap_test.hs` with the most general solution:

```haskell
isLeapYear :: Integral a => a -> Bool
```

It's important to supress this warning so we don't misguide newcomers in thinking that there is something wrong with their code, when the problem is in the test module. Also, using the most general type is usually encouraged in Haskell.

I don't think it's possible to write a test allowing those three distinct signatures without raising a warning, so I propose to suppress it with:

```haskell
{-# OPTIONS_GHC -fno-warn-type-defaults #-}
```

I could make a *pull request* for **leap** and maybe other exercises sharing the same problem, but I haven't checked yet how many they are.

Anyone against it?

rbasso:
In some problems, like `leap`, the test module doesn't declare explicitly some numeric types, allowing solutions with multiple types:

```haskell
isLeapYear :: Int     -> Bool
isLeapYear :: Integer -> Bool
```

This is desirable, but has the unintended effect of displaying the warning...

```
Defaulting the following constraint(s) to type `Integer` arising from the use of `isLeapYear` at leap_test.hs ... arising from the literal 1996 ...
```

... when compiling `leap_test.hs` with polymorphic solutions:

```haskell
isLeapYear :: Integral a => a -> Bool
```

It's important to suppress this warning so we don't misguide newcomers in thinking that there is something wrong with their code, when the problem is in the test module. Also, using the most general type is actually encouraged in Haskell.

It's desirable to have test modules that compile silently. This would benefit users and automated tests, that could be compiled with `-Wall -Werror`.

Affected problems:
- grains
- hamming
- leap
- pascals-triangle
- scrabble-score
- sieve
- sum-of-multiples

**Proposed solution**: Add `{-# OPTIONS_GHC -fno-warn-type-defaults #-}` to the affected modules.

Closes #143

### DIFF
--- a/exercises/grains/grains_test.hs
+++ b/exercises/grains/grains_test.hs
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
+
 import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
 import System.Exit (ExitCode(..), exitWith)
 import Grains (square, total)
--- a/exercises/hamming/hamming_test.hs
+++ b/exercises/hamming/hamming_test.hs
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
+
 import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
 import System.Exit (ExitCode(..), exitWith)
 import Hamming (distance)
--- a/exercises/leap/leap_test.hs
+++ b/exercises/leap/leap_test.hs
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
+
 import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
 import System.Exit (ExitCode(..), exitWith)
 import LeapYear (isLeapYear)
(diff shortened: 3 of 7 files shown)


rbasso:
Just to make it clear, the list of affected problems is probably not complete, as it would demand inspection of all exercises individually. These are the ones I identified running my last iteration of each exercise, but not all of my solution are polymorphic as to force exhibition of the warning.

Anyway, the solution for any new appearing warning of that type would be exactly the same.

This PR will probably conflict with #146 .

rbasso:
I think I solved the conflict cause by merge of #146 is this PR seems OK again.

petertseng:
I noted in #143 that this is desirable (for the reasons listed in that issue and in this PR description). I would even support merging this as-is before I get a chance to investigate. I say this because I honestly think we may be in a "your guess is as good as mine" situation so if you couldn't find anything it's possible I wouldn't either. If I do find a better solution it's always possible for me to make another PR, so not really necessary to hold this one up!

petertseng:
The thing I didn't check thoroughly would be "is there any exercise you missed", but I'm sure we can fix that up easily if that is the case. Seems to require going through the entire list of exercises, not an easy way to grep for something that would be indicative...

PR #152, opened 2016-06-17
sum-of-multiples: Update test cases to match x-common json file

rbasso:
Add test case { "factors": [3, 5]
              , "limit": 100
              , "expected": 2318 }

Swap cases { "factors" : [7, 13, 17]
           , "limit"   : 20
           , "expected": 51 }
       and { "factors" : [4, 6]
           , "limit"   : 15
           , "expected": 30 }

Closes #87


petertseng:
Confirmed, with this we match https://github.com/exercism/x-common/blob/master/sum-of-multiples.json exactly. 👍 from me

PR #154, opened 2016-06-18
check-exercises: Fix warning with ghc >= 7.10.

### LINKED ISSUE #122: check-exercises.hs redundant import warning after GHC 7.10.x
Travis keeps warning with:

```
_test/check-exercises.hs:18:1: warning: [-Wunused-imports]
    The import of ‘Control.Applicative’ is redundant
      except perhaps to import instances from ‘Control.Applicative’
    To import instances alone, use: import Control.Applicative()
```

Should check whether that's required. If not required, remove the line completely. If required, add the `()`.

rbasso:
This fixes the warning *"The import of Control.Applicative is redundant"* when compiling `check-exercises` caused by `Prelude` re-exporting `Control.Applicative.<$>` in `base >= 4.8`.

Removing the offending line is undesirable because it would break compiling if `ghc <= 7.10`.

Closes #122

### DIFF
--- a/_test/check-exercises.hs
+++ b/_test/check-exercises.hs
@@ -15,7 +15,10 @@ import System.Process (rawSystem)
 import Data.List (isPrefixOf, intercalate)
 import Data.Maybe (catMaybes)
 import Control.Monad (filterM)
-import Control.Applicative
+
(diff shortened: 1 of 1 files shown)


petertseng:
Seems good as I no longer see warning, so :+1:

PR #155, opened 2016-06-18
robot-simulator: Add test cases for turning.

### LINKED ISSUE #116: robot-simulator: "turning edge cases" are not necessarily edge cases
We have:

```haskell
  [ testCase "turning edge cases" $ do
    North @=? turnRight West
    West @=? turnLeft North
-- other tests omitted
   ]
```

These aren't necessarily edge cases. It's only an edge case for *the particular implementation* used in the example solution, where North is the first variant of the `data Bearing =`, and West is the last. Any student may have chosen a different ordering, at which point these aren't edge cases anymore.

rbasso:
Closes #116


petertseng:
yup :+1:

PR #156, opened 2016-06-19
clock: Update to match x-common/clock.json

### LINKED ISSUE #125: clock: canonical test data has been improved
The JSON file containing canonical inputs/outputs for the Clock exercise has gotten new data.

There are two situations that the original data didn't account for:

- Sometimes people perform computation/mutation in the display method instead of in add. This means that you might have two copies of clock that are identical, and if you add 1440 minutes to one and 2880 minutes to the other, they display the same value but are not equal.
- Sometimes people only account for one adjustment in either direction, meaning that if you add 1,000,000 minutes, then the clock would not end up with a valid display time.

If this track has a generator for the Clock exercise, go ahead and regenerate it now. If it doesn't, then please verify the implementation of the test suite against the new data. If any cases are missing, they should be added.

See https://github.com/exercism/x-common/issues/166

rbasso:
Add test to match x-common/clock.json.
Remove legacy tests.

Closes #125

### DIFF
--- a/exercises/clock/clock_test.hs
+++ b/exercises/clock/clock_test.hs
@@ -21,29 +21,86 @@ main = exitProperly $ runTestTT $ TestList
 
 clockTests :: [Test]
 clockTests =
-  [ testCase "test constructor and toString" $ do
-    "08:00" @=? toString (fromHourMin 8 0)
-    "09:00" @=? toString (fromHourMin 9 0)
-    "11:09" @=? toString (fromHourMin 11 9)
-  , testCase "fromInteger should work in minutes" $ do
+  [ testCase "fromInteger should work in minutes" $ do
     "00:03" @=? toString 3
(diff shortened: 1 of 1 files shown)


exercises/clock/clock_test.hs
```diff
@@ -21,29 +21,86 @@ main = exitProperly $ runTestTT $ TestList
 
 clockTests :: [Test]
 clockTests =
-  [ testCase "test constructor and toString" $ do
```
petertseng: I'm assuming these are getting removed because the two components (`toString` and `fromHourMin`) are being tested separately elsewhere, so it would not be necessary to test them together here. I agree.

petertseng:
This all seems good. Note that the commit message still has clok.json in it; only the PR title got updated

rbasso:
Fixed!

petertseng:
:+1: on this

PR #159, opened 2016-06-22
Add haskell exercises' dependency control.

### LINKED ISSUE #151: discussion: Tracking dependencies (update README.md to include package "lens").
I was tracking the [dependencies](https://github.com/exercism/xhaskell/files/319495/dependencies.txt) for each of the exercises and example solutions - to run some local experiments with compartmentalized tests - and I found a minor omission in `README.md`.

The table below summarizes all dependencies, except for the package `base`:

| package | all | exercises | check-exercises | README | bootstrap | GHC |
| :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| array | x | x | | | | x |
| attoparsec | x | x | | x | x | |
| containers | x | x | | | | x |
| directory | x | | x | | | x |
| filepath | x | | x | | | x |
| HUnit | x | x | | x | x | |
| lens | x | x | | | x | |
| parallel | x | x | | x | x | |
| process | x | | x | | | x |
| QuickCheck | x | x | | x | x | |
| random | x | x | | x | x | |
| split | x | x | | x | x | |
| stm | x | x | | x | x | |
| text | x | x | | x | x | |
| time | x | x | | | | x |
| unix | x | | x | | | x |
| vector | x | x | | x | x | |

I opened this issue not because the package `lens` isn't listed in `README.md`, but to ask a question:

**Should we keep a more detailed track of dependencies?**

The advantages would be:

- Automatically generate correct lists of dependencies.
- Allow *sandboxed*, automated tests with Cabal, Stack or Nix.
- Optionally, deliver exercises as automatically generated projects.

I'm not proposing any of those thing! (I'm already happy doing it locally)
I'm also not proposing any specific form of dependency track. This we can discuss later, if enough people think it's a good idea.

What I would like to know is if anyone has anything in favor of against keeping some form of updated track of each exercise's dependencies inside `xhaskell`.

So... anyone?

rbasso:
Install three files:

- `dependencies.txt` - List of dependencies for all Haskell sources.
- `dependencies` - Shell script to query dependencies.
- `stackalize` - Shell script to convert an exercise in a stack project.

Sample usage:

```
$ dependencies go-counting
array
base
containers
HUnit

$ dependencies `ls exercises`
array
attoparsec
...
time
vector

$ dependencies --only-in-examples `ls exercises`
attoparsec
lens
..
split
stm

$ cd exercises/leap
$ stackalize --resolver lts-6.4
...
$ stack test
$ stackalize --undo
```

Closes #151

### DIFF
--- a/_test/stackalize
+++ b/_test/stackalize
@@ -1,0 +1,139 @@
+#!/usr/bin/env bash
+
+shopt -s extglob
+
+exampleFile="example.hs"
+testSuffix="_test.hs"
+srcDir="src"
+testDir="test"
+programName=`basename "$0"`
+
+depsFinder="`dirname ${BASH_SOURCE}`/dependencies"
+
+generateStackYaml () {
+    echo "resolver: ${stackResolver}"
+}
+
+generatePackageYaml () {
+    local exercise=`basename ${PWD}`
+    local commonDeps=`${depsFinder} --examples-and-tests ${exercise}`
+    local exampleDeps=`${depsFinder} --only-in-examples ${exercise}`
+    local testDeps=`${depsFinder} --only-in-tests ${exercise}`
+
+    echo "name: $exercise"
+    echo ""
+    echo "dependencies:"
+
+    for dependency in ${commonDeps}; do
+      echo "  - ${dependency}"
+    done
+
+    echo ""
+    echo "library:"
+    echo "  source-dirs: ${srcDir}"
+    echo "  dependencies:"
+
+    for dependency in ${exampleDeps}; do
+      echo "    - ${dependency}"
+    done
+
+    echo ""
+    echo "tests:"
+    echo "  test:"
+    echo "    main: ${exercise}${testSuffix}"
+    echo "    source-dirs: ${testDir}"
+    echo "    dependencies:"
+    echo "      - ${exercise}"
+
+    for dependency in ${testDeps}; do
+      echo "      - ${dependency}"
+    done
+}
+
+stackalizeProject () {
+    local exercise=`basename ${PWD}`
+    local module=`grep "^module" ${exampleFile} | cut -d " " -f 2`
+    if [ ! -e "${exampleFile}"           ] ||
+       [ ! -e "${exercise}${testSuffix}" ] ; then
+       echo "$0: Cannot stackalize. Exercise files not found."
+       exit 1
+    else
+        mkdir .extraFiles && # This first move all files to .extraFiles
+        mv * .extraFiles  && # to avoid additional files breaking the build.
+        mkdir ${srcDir}   && # After that we move again just the example
+        mkdir ${testDir}  && # and the test to their destination folders.
+        mv .extraFiles/${exampleFile} ${srcDir}/${module}.hs &&
+        mv .extraFiles/${exercise}${testSuffix} ${testDir}   &&
+        (generateStackYaml   > stack.yaml)                   &&
+        (generatePackageYaml > package.yaml)
+    fi
+}
+
+stackalizeUndoProject () {
+    local sourceFiles=`find ${srcDir} -type f`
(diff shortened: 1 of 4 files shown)


rbasso:
This may not be the best dependency control system designed by humankind, but I expect it works well enough, and it's better to have something than nothing. :smile:

`stackalize` is in the pack because it depends on `dependencies`, and it's the simplest way I found to get a running environment to run the test and check if the dependencies are correctly specified. I also makes very easy to test locally against multiple GHC versions before committing.

rbasso:
The dependency-list for each of the exercises was [successfully tested](https://travis-ci.org/rbasso/xhaskell/builds/139647895) in individual stack projects with Travis, using lts-2.22, lts-6.4 and nightly-2016-06-21. The dependencies of examples were specified and tested independently, so they can be regarded as correct.

petertseng:
Perhap I'll have to grab Stack and play around with it this weekend to try out the effects of this for myself. I can reason about bash well enough to understand what's here though.

One interesting question I had is - we would want to note somewhere that adding a new exercise to this track will also want the dependencies added to dependencies.txt, right? Unless I missed a way to automatically populate that file

_test/stackalize
```diff
@@ -0,0 +1,139 @@
...
+stackalizeProject () {
+    local exercise=`basename ${PWD}`
+    local module=`grep "^module" ${exampleFile} | cut -d " " -f 2`
+    if [ ! -e "${exampleFile}"           ] ||
+       [ ! -e "${exercise}${testSuffix}" ] ; then
+       echo "$0: Cannot stackalize. Exercise files not found."
+       exit 1
+    else
+        mkdir .extraFiles                                    &&
+        mv * .extraFiles                                     &&
```
petertseng: hmm, would you explain to me what happens if not moving files to here? What would break? Maybe worth a comment here, unless you think it's obvious to anyone who knows what they doing (I certainly don't with Stack, I know)

rbasso:
You are correct, @petertseng.

It is, in fact, impossible to automatically populate the dependencies, so that would demand an update to the `README.md` explaining the contributors that the file must reflect the dependencies.

It's a way to enforce good discipline and minimize maintenance. It would also allow reproducible build environments for contributors.

If in the future we move to stack based tests, Travis will check for correctly specified dependencies automatically.

_test/stackalize
rbasso: First, you need to know that this is not the best example of *bash scripting* in the world. It's just something that I improvised to solve this problem. :smile: 

What the script does is:

- move everything to `.extraFiles`, so we get a known directory struture.
- move from there the *example* and *test* files to their correct locations.
- generate a `stack.yaml` with the provided resolver.
- generate a `package.yaml` (simpler than a *.cabal* file) with the correct dependencies.

It can centainly be improved, but It's been 17 years since the last time I played with `bash` as a sysadmin. Let's say it works under normal conditions. :grin: 

About your question, if there are additional source files in the directory, as in `forth` and `zipper`, they could be included accidentally in the project and break the build. Also, I wanted to be able to *unstackalize* the project, moving everything back to their places.

rbasso: If you think these scripts deserve a rewrite, fell free to change anything. I just need a way to assemble the project.

rbasso: Added the comments.
Now working on the `README.md`.

rbasso:
Added a new README.md with some instructions for contributors on how to use *stack*.

Also updated the list of packages and included a small section mentioning dependency control.

rbasso:
Considering that this PR only adds functionalities and updates an outdated documentation file, there is no risk in merging it now, even if there are some problem with the scripts we are not catching until now.

This way we can at least keep the dependency list updated and start testing #162.

petertseng:
I'm late to the show but you correctly determined you didn't need to wait for me. And I would have said :+1:  anyway.

PR #160, opened 2016-06-22
New .gitignore based on GitHub's default.

rbasso:
Add new entries to .gitignore to make move convenient for
contributors to use alternative build methods.

 - Default .ignore exclusions for Haskell from GitHub.
 - Exclusions for development with cabal-sandboxes, stack and hpack.
 - Reordering.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,29 @@
-*.swp
-.DS_Store
-tmp
+# exercism specific patterns
 bin/configlet
 bin/configlet.exe
-/.vagrant
+
+# exercism/xhaskell specific patterns
+*.cabal
+package.yaml
+stack.yaml
+
+# sorted default .gitignore from github/gitignore/Haskell.gitignore
+*.aux
+*.chi
+*.chs.h
+*.dyn_hi
+*.dyn_o
+*.eventlog
+*.hi
+*.hp
+*.o
+*.prof
+.cabal-sandbox/
+.hpc
+.hsenv
+.stack-work/
+cabal-dev
+cabal.project.local
+cabal.sandbox.config
+dist
+dist-*


rbasso:
At the moment, it's only practical to build the examples and tests in `xhaskell` using a globally installed GHC or a global Stack profile. With #159 , it will be possible to build and test exercises locally as auto-contained projects.

To avoid accidental inclusion of build directories and local project files in a commit, it's important to update `.gitignore`. Given the opportunity, we also update it with all the most common exclusions for Haskell development.

.gitignore
```diff
@@ -1,6 +1,25 @@
...
 bin/configlet
 bin/configlet.exe
-/.vagrant
+cabal.sandbox.config
+cabal-dev
+dist
+package.yaml
+stack.yaml
+tmp
+.DS_Store
```
kotp: Should these be involved in Exercism?  Or is it better to educate contributors on where these go in `~/.gitignore`  Anything Haskell related I think should be here, but editor and OS are outside the scope of this project.

.gitignore
```diff
@@ -1,6 +1,25 @@
...
+.cabal-sandbox
+.hpc
+.hsenv
+.stack-work/
+.vagrant
```
kotp: Same with `.vagrant` unless this is required for this track.

rbasso:
Your are right, @kotp. Those patterns did not belong here. Thanks!

I rewrote it completely with some comments and I would appreciate if you could take another look at that. :smile:

kotp:
Yep, looks better.  Thanks!

If you sort those in each categories, I think that would be an improvement, and will allow any additional changes to be in some easily scanned and findable order.

rbasso:
I'll leave this PR open for a while to see if anyone else has a rationale to include/exclude anything before merging.

Thank you again!

rbasso:
Done.

Thanks again!

petertseng:
All seems reasonable. If we are really wanting to nit I would note that I more commonly see "specific" than "especific" but of course this is minor.

rbasso:
Fixed and now mostly English compliant, @petertseng. :smile:

petertseng:
only *mostly*?! =D

all seems good to me :+1:

Issue #124, opened 2016-04-07
travis: Document why we test the GHC versions that we do


petertseng:
For example #82 explains that we test 7.6 because some versions of Linux use that. An example I found was that [Ubuntu trusty uses 7.6](http://bazaar.launchpad.net/~ubuntu-branches/ubuntu/trusty/haskell-platform/trusty/view/head:/debian/control). It would be good if we would express the rationale as comments in the .travis.yml file so that we may potentially answer questions such as "when is it safe to remove our support for 7.6?"

To be clear, not that I'm advocating removing support of 7.6 anytime soon. In fact I may even advocate for adding 7.4, since I found that [Ubuntu Precise](http://bazaar.launchpad.net/~ubuntu-branches/ubuntu/precise/haskell-platform/precise/view/head:/debian/control) ships with that.

Edit: Indeed I still would have advocated for 7.4 but experiments in #131 to try to get 7.4 working were unsuccessful. (no `mkdtemp` in `System.Posix.Temp` in the version of `unix` we get with 7.4).

rbasso:
I completely agree with you! Would you mind adding that to `.travis.yml`? :grin:

rbasso:
Do you think something like this would be enough?

```
matrix:                  # Some, still used, major linux distros pack
  include:               # old GHC versions, so we test them too, to
                         # make the Haskell track more accessible.

    - env: GHCVER=7.6.3  # Minor version released on 2013-04-21.
      addons:            # Debian stable (jessie).
        apt:             # Fedora 20 and 21.
          <<: *apt       # Ubuntu 14.04 LTS (trusty).

          packages: [ghc-7.6.3, *cabal]

    - env: GHCVER=7.8.4  # Fedora 22, 23 and 24.
      addons:            # Ubuntu 15.10 LTS (wily).
        apt:             # Stackage's snapshots from lts-0.0
          <<: *apt       # to lts-2.22 use ghc-7.8.x.
          packages: [ghc-7.8.4, *cabal]

    - env: GHCVER=7.10.3 # Debian testing (strech).
      addons:            # Ubuntu 16.04 LTS (xenial).
        apt:             # Stackage's snapshots from lts-3.0
          <<: *apt       # to lts-6.4 (latest) use ghc-7.10.x.
          packages: [ghc-7.10.3, *cabal]

    - env: GHCVER=8.0.1  # Newer versions of the Haskell Plaftorm.
      addons:            # Stackage's nightly snapshots.
[...]

PR #162, opened 2016-06-24
Experimental Stack-based Travis-CI (DO NOT MERGE IT)

### LINKED ISSUE #129: travis: Automatically invalidate caches if new package versions exist
While working on #101 I discovered that because we cache .cabal, we'll get conflicts (and thus build failures) if there's a new version of a cabal package we use.

That's less than ideal (and we had to manually clear the cache to fix the build), but I believe there's a way out. Should examine the work done in https://github.com/hvr/multi-ghc-travis/blob/master/make_travis_yml.hs - it seems to check the install plan and compares it to the last known one. If there's a change, it clears the packages and asks cabal to reinstall. I think this will work, though it may need some testing... and depending on how often our used packages release new versions opportunities to test may be rare.

### LINKED ISSUE #153: travis: Travis-CI Roadmap
Right now, we have 4 *Issues* ( ~~#119~~ , #124 , #129 and ~~#130~~ ) and 2 "Pull Requests" ( ~~#120~~ , ~~#128~~ ) related to *Travis* pending. Except for 129, that seems to require substantial changes, all the others appear to be simple and will probably be merged soon.

However, before solving 129, it would be useful to at least have an idea about what we plan to achieve, so we can focus our work in a more productive way.

I think that, ideally, those things would be desirable in *Travis*:

- [x] Test with multiple versions of GHC.
- [ ] Isolated, reproducible test environments.
- [ ] Automatic handling of dependencies ( Depends on #151 )
- [ ] Minimal maintenance on software updates.

The only idea I had until now was to use a list of dependencies to generate *stack projects* on-the-fly to pass to *Travis* test, using cache to reduce subsequent build times. It works surprisingly well, but only if GHC version >= 7.8.2.

It would still be possible to treat GHC 7.6.x as a special case and use *cabal*. But that's is an ugly solution and adds a lot of complexity.

Alternatively, we can keep using a global GHC installation with *cabal*, which is simple and easy, but is also less reproducible and doesn't allow us to enforce correct declaration of exercises' dependencies.

I would really love some feedback about this from as many people as possible.

So... anyone?

rbasso:
As we add dependency control at the source level with #159, It turns possible to automatically build and test each exercise in a *sandbox* with *Stack* inside *Travis-CI*.

There are some good reasons to experiment with this solution:

- Locally reproducible test with automatic handling of dependencies, multiple version of GHC and packages snapshots. :white_check_mark: 
- It automatically tests for newer versions of compilers and packages, allowing us to be notified of problems and prepare earlier for changes. :white_check_mark: 
- We get the ability to enforce correctly specified dependencies for each of the exercises for free. :white_check_mark: 
- Playing with new tools is good for learning. :smile: 
- There is no loss if it fails completely. :+1: 

The idea is to keep this *pull request* always based on *master* without being merged, so we can compare how it performs with our actual solution and correct any dependency change not correctly specified in `dependencies.txt`.

If the test are successful, merging this would:
- Close #129
- Close #153 

Also, #124 is affected, because stack doesn't support GHC 7.6.3, and we would have to quit testing with it.


petertseng:
One advantage of merging this is that we wouldn't have to do things like #172 because Travis would test that for us. But the disadvantage would be losing 7.6 in CI, is that right? That is a real shame; were it not for that, I'd be in favor of merging. I know in https://github.com/exercism/xhaskell/issues/153 the idea of "as a special case let's test 7.6 with plain cabal and not Stack" was brought up though I haven't explored what that looks like yet. Maybe it'd be worthwhile to take a look at what some other projects do with this and how they manage the resulting complexity.

Now, it's true that the maintenance burden of keeping this PR up to date with master is currently handled by you, so nobody else has to deal with that. It would still be good to resolve the tension, though, and to get the advantage of getting our dependencies checked.

So why don't I do some exploring for the hybrid solution, we see what it looks like, then figure out what to do

petertseng:
stack tests both with/without stack: https://github.com/commercialhaskell/stack/blob/master/.travis.yml

rbasso:
I think there are two obstacles to implementing a hybrid solution:

- Conditionally install dependencies and run test.  *- Easy, but ugly!*
- Change `bootstrap.sh` to use `dependencies.txt` as source. *- Hard and ugly!*

I'm in favor of pushing users to use never GHC versions, mostly because I believe  that people should get used to how Haskell code is actually being written (*ghc-7.10.3* and *ghc-8.0.1*, *stack* and *cabal sandboxes*), but also to make *xhaskell* more maintainable.

That said, I agree that we can have a mixed solution, so we can postpone the decision of dropping *ghc-7.6.3* until we think it is time to let it go.

I propose to merge the mixed solution in two phases:

1. Change `bootstrap.sh` to use `_test/dependencies`. This has it's own merit, centralizing all dependencies in one place.
2. Finally, the hybrid tests.

I see that you already have a PR for this, so let's discuss the details there.

petertseng:
> Change `bootstrap.sh` to use `_test/dependencies`. This has it's own merit, centralizing all dependencies in one place.

That seems good. Will you do this or should I give it a try? Might be worth a separate issue to keep track of this task if you think it's worthy.

rbasso:
> That seems good. Will you do this or should I give it a try? Might be worth a separate issue to keep track of this task if you think it's worthy.

Definitely worthy it's own issue. :smile: 
If you want to write it, please do it. Today I'm kind of busy and it's 33C here, so I'm not thinking clearly. :sweat_smile:

rbasso:
That was fast, @petertseng! Congrats! :tada: 

With #179 and #174 I guess we will be fine with a mixed Travis configuration. :smile: 

Let's just make quick a comparison so that we have an idea about the price we are paying for ghc-7.6.3:

- `bootstrap.sh` (0 -> 24 lines = **+24**)
- `.travis.yml` (53 -> 111 lines = **+58**)
-  #129 is still **unsolved** :worried: 

I'll trust your judgement in this one.
Just let me know when you are comfortable with merging #179 and then #174. :+1:

PR #164, opened 2016-06-28
leap: Tests rewritten to match x-common/leap.json

rbasso:
- Three cases added
- Rewritten to reduce code repetition and make it more readable.


petertseng:
Looks good, exits 0 with the good impl (well, I didn't need to say that, Travis would already test that) and shows reasonable errors and exits 1 with bad impl

```
### Failure in: 6:leap y2k                
leap_test.hs:14
expected: False
 but got: True
Cases: 7  Tried: 7  Errors: 0  Failures: 5
```

so :+1:

PR #165, opened 2016-06-29
Updating the Nucleotide Count exercise to return Either

### LINKED ISSUE #157: nucleotide-count: Change types to return an Either.
We are considering changing the interface of the problem `nucleotide-count`. The idea (#115)  is to allow a more idiomatic solution to this problem, returning a *Maybe*, an *Either* or something else, instead of forcing the user to throw an `error` with the specific message: **invalid nucleotide 'X'**.

It could be something like this:

```haskell
module DNA (count, nucleotideCounts) where

count :: Char -> String -> Maybe Int
nucleotideCounts :: String -> Maybe (Map Char Int)
```

This is just and sketch, and it can/should change. It depends only on the feedback here!

Any suggestion or comment is welcome! :smile:

samjonester:
The resulting exercise is more idiomatic to Haskell. The `Either` will provide a useful error message to end users of the functions.

Closes #157

### DIFF
--- a/exercises/nucleotide-count/example.hs
+++ b/exercises/nucleotide-count/example.hs
@@ -1,19 +1,20 @@
 module DNA (count, nucleotideCounts) where
-import Data.Map.Strict (Map, fromListWith)
 
-count :: Char -> String -> Int
-count needle = length . filter (nucleotide ==) . map verifyDNA
-  where nucleotide = verifyDNA needle
+import qualified Control.Applicative as A
+import Data.Map.Strict (Map, (!), fromDistinctAscList, fromListWith, findWithDefault)
 
-nucleotideCounts :: String -> Map Char Int
-nucleotideCounts strand = fromListWith (+) (defaults ++ map pair strand)
- where defaults = zip dna (repeat 0)
-       pair nucleotide = (verifyDNA nucleotide, 1)
+count :: Char -> String -> Either String Int
+count x xs = (!) A.<$> nucleotideCounts xs A.<*> valid x
 
-dna :: String
-dna = "ACGT"
+nucleotideCounts :: String -> Either String (Map Char Int)
+nucleotideCounts xs = fromDistinctAscList A.<$> mapM count' "ACGT"
+  where
+    count' x = (\c -> (x, c)) A.<$> occur' x
+    occur' x = findWithDefault 0 x . countOccurrences A.<$> mapM valid xs 
+    countOccurrences = fromListWith (+) . flip zip (repeat 1)
 
-verifyDNA :: Char -> Char
-verifyDNA nucleotide | nucleotide `elem` dna = nucleotide
-                     | otherwise = error ("invalid nucleotide " ++
-                                          show nucleotide)
+valid :: Char -> Either String Char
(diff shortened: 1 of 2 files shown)


samjonester:
@rbasso I borrowed your implementation for the example implementation. However, it requires the `multiset` package. I'm not sure whether it should be added to the build, or if I should implement the example using only the standard lib. I'm fine with either approach, but which do you think I should take?

exercises/nucleotide-count/example.hs
```diff
@@ -1,19 +1,18 @@
...
+count :: Char -> String -> Either String Int
+count x xs = (!) <$> nucleotideCounts xs <*> valid x
 
-dna :: String
-dna = "ACGT"
+nucleotideCounts :: String -> Either String (Map Char Int)
+nucleotideCounts xs = fromDistinctAscList <$> mapM count' "ACGT"
+  where
+    count' x = (\c -> (x, c)) <$> occur' x
+    occur' x = occur x . fromList <$> mapM valid xs
```
rbasso: All the uses of `<$>` would fail if `base < 4.8`, because it was only in `Control.Applicative` before that.

It's a little [tricky](https://ghc.haskell.org/trac/ghc/wiki/Migration/7.10#GHC7.10.xMigrationGuide) to use it in a compatible way, because if you `import Control.Applicative ((<$>))` it would raise a warning on newer versions, and we intentionally fail all tests with warnings on *Travis-CI*.

rbasso:
Thanks for making this pull request, @samjonester. :smile: 

Right now we don't have too much flexibility about the packages we can use.
To add a package, we would have to update:

- ` README.md`
- `docs/INSTALLATION.md`
- `_test/bootstrap.sh`
- `_test/dependencies.txt`

We expect to simplify this situation in the future with #162 , but now this is what we have.

Also, considering that it's not easy for beginners using cabal to deal with new dependencies, I would avoid adding new packages for now unless they are really needed. 

Recently, I updated the `README.md` with instructions on how to use *stack* to test an exercise locally. Could you please take a look at that, as I believe you current version would throw an error with GHC 7.8.4.

samjonester:
I've subverted the problem with prelude with `import qualified Control.Applicative as A`, and using `A.<$>` + `A.<*>`.

Also, I've removed the dependency on `Data.Multiset` by rolling my own naive implementation of the two functions were being used.

@rbasso I was unable to get `_test/stackilize` to run.

```
$ _test/stackalize --resolver lts-6.4 exercises/nucleotide-count
_test/stackalize: No resolver specified.
```

rbasso:
Thanks, @samjonester !

About the `stackalize` problem, it is a bash script I improvised to run on Travis-CI. I was tested only on my machine and on Travis, so you are probably the first alpha-tester. :grin: 

If you have the time, could you open an issue with further details? I would like to fix it, as it is the only practical way we have to test problems locally with restricted dependencies.

It would be good to know:

- What is your OS?
- What `bash` version are you running? `bash --version`
- What `getopt` version are you running? `getopt --version`

I guess the problem is with `getopt`, from package `util-linux`

PR #168, opened 2016-06-30
rna-transcription: Changes return type to Maybe

### LINKED ISSUE #166: rna-transcription: Change return type to Maybe
This exercise doesn't define a test case for "non dna" characters, so many people are throwing errors for invalid input unnecessarily. It seems like that solution has gone viral, and since it's early in the haskell track could influence people to throw errors instead of using the safer type of `Maybe`. Thoughts on making this change?

samjonester:
- Error case is now defined, eliminating the desire to throw an `error`
for invalid input.
- The test and example solution are more idiomatic to Haskell, using a
  `Maybe`.

Closes #166

### DIFF
--- a/exercises/rna-transcription/DNA.hs
+++ b/exercises/rna-transcription/DNA.hs
@@ -1,0 +1,6 @@
+module DNA (toRNA) where
+
+-- | if string contains invalid character, return Nothing
+-- | if string contains only valid nucleotides, return Just transcription
+toRNA :: String -> Maybe String
+toRNA = undefined
--- a/exercises/rna-transcription/example.hs
+++ b/exercises/rna-transcription/example.hs
@@ -1,9 +1,12 @@
 module DNA (toRNA) where
 
-toRNA :: String -> String
-toRNA = map $ \c -> case c of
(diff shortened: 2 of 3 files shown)


rbasso:
Great! Thanks! :+1: 

I'm going to merge it!

Just one thing. It's always good to take a look at [exercism/x-common](https://github.com/exercism/x-common) to see if there is a `.json` file [for the exercise](https://github.com/exercism/x-common/blob/master/rna-transcription.json). Usually those files have a good list of test cases we should implement, this way we don't have to duplicate work thinking about that.

samjonester:
Do you know why CI didn't run for this PR? Also, thanks for the link. I can amend the commit with the other test cases before you merge.

rbasso:
Trying to find out what's happening with Travis. I'll hold this PR until the problem is solved

rbasso:
I wrote a new PR #169 , just to test it. Same problem!

samjonester:
Looks like it's working again. I've also updated the test cases.

rbasso:
Perfect! :smile:

PR #170, opened 2016-06-30
README: fixup spelling and grammar

petertseng:
sorry that I had to do it, but I guess that if I didn't, someone else would, and it may be good to pre-empt so that they get fixed fast.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,15 +10,15 @@ Please see the [contributing guide](https://github.com/exercism/x-api/blob/maste
 
 ### Development Dependencies
 
-There are currently two distinct ways to set up an environment to colaborate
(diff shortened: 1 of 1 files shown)


rbasso:
Thank you very much, @petertseng!

> sorry that I had to do it, but I guess that if I didn't, someone else would, and it may be good to pre-empt so that they get fixed fast.

Don't be sorry for doing the right thing. :smile:

Can you solve the conflicts so I can merge this one?

petertseng:
Kind of makes me think I should have put both the #171 and #170 into one PR but I am in this habit. And I am not afraid to deal with conflicts

PR #172, opened 2016-07-01
rna-transcription: Fix missing dependency cause by exercism/xhaskell#168.

rbasso:
Add `containers` to `rna-transcription/example.hs` dependencies.


petertseng:
ah, es

rbasso:
I just forgot to check this when I merged #168., But #162 got that for me. :smile:

PR #173, opened 2016-07-02
all-your-base: Add new exercise

### LINKED ISSUE #167: all-your-base: Add exercise to allow deprecation of binary, trinary, octal and hexadecimal.
It was decided in exercism/x-common#276 that `all-your-base` will eventually replace the four base conversion exercises we currently have: `binary`, `trinary`, `octal`, `hexadecimal`. But, before deprecating them, we need to implement `all-your-base` in this track.

This exercise's interface could be something like this:

```haskell
module Base (rebase) where

rebase :: Integral a => a -> a -> [a] -> Maybe [a]
rebase b1 b2 ...
```

Where `b1` is the the input's base and `b2` the output's base.

There was a long discussion in the PR exercism/x-common#280 that added `all-your-base` about how the corner cases should be handled, and it was decided that each track will have to choose the most idiomatic solution.

IMHO, we should follow most of the semantics used by the package [digits](https://hackage.haskell.org/package/digits-0.2/docs/Data-Digits.html), but also checking for invalid digits and bases. This means that the canonical representation for the digits of zero would be zero `[]`, and that leading zeros would be always superfluous:

```haskell
rebase 2 10    [] = Just  [] -- Zero has no digits
rebase 2 10   [0] = Just  [] -- One zero digit is the same as []
rebase 2 10 [0,0] = Just  [] -- Multiple zero digits are the same as []
rebase 2 10 [0,1] = Just [1] -- Leading zeros are ignored
rebase 2 10 [1,7] = Nothing  -- Invalid digits return Nothing
rebase 2  1 [1,0] = Nothing  -- Invalid bases return Nothing
```

I believe this would allow for some really simple and elegant solutions, but some people prefer to represent the digits of zero as `[0]`.

It's up to the one writing the PR to decide how the cases undefined in `all-your-base.json` should be treated! :+1:

rbasso:
Add the tests, an example solution and a stub.
Add exercise to `config.json`, after `atbash-cipher`.
Add dependencies to `dependencies.txt`.

Closes #167

### DIFF
--- a/_test/dependencies.txt
+++ b/_test/dependencies.txt
@@ -1,5 +1,7 @@
 accumulate/example.hs: base
 accumulate/accumulate_test.hs: base HUnit
+all-your-base/example.hs: base
+all-your-base/all-your-base_test.hs: base HUnit
 allergies/example.hs: base
 allergies/allergies_test.hs: base HUnit
 anagram/example.hs: base
--- a/config.json
+++ b/config.json
@@ -34,6 +34,7 @@
     "raindrops",
     "allergies",
     "atbash-cipher",
+    "all-your-base",
     "bank-account",
     "crypto-square",
     "kindergarten-garden",
--- a/exercises/all-your-base/Base.hs
+++ b/exercises/all-your-base/Base.hs
@@ -1,0 +1,4 @@
+module Base (rebase) where
+
+rebase :: Integral a => a -> a -> [a] -> Maybe [a]
+rebase inputBase outputBase inputDigits = undefined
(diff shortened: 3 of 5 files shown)


petertseng:
Hmm. I honestly don't think I have anything I would change, so I think I can say :+1: . We can see after some time if people really want to be able to output `[0]`; this may be something we can't really predict.

I think the placement is reasonable enough since it comes in between the existing to-be-deprecated base-conversion exercises.

rbasso:
Thanks for the feedback, @petertseng. :smile:

PR #174, opened 2016-07-03
travis: test both Stack and non-stack (Cabal)

### LINKED ISSUE #124: travis: Document why we test the GHC versions that we do

### LINKED ISSUE #153: travis: Travis-CI Roadmap

petertseng:
The work in #162 is advantageous because we can check our dependencies
and have hermetic builds (do not break on upstream changes).
    
One question about whether to merge it was the necessity to drop GHC
7.6.
    
This commit introduces to Travis the ability to test with either Stack
or without (the old way of doing things). We test 7.6, 7.8, 7.10, 8.0
with cabal-install as we did before, and also Stack as done in #162.

Closes #162 (is based on it)
Closes #124 (explanatory comments provided)
Closes #153 (based on #162 which also closed #153)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,6 +71,7 @@ before_install:
         export PATH="${TRAVIS_BUILD_DIR}/_test:$PATH" # For stackalize.
         ;;
       cabal)
+        export PATH=/opt/cabal/$CABALVER/bin:/opt/ghc/$GHCVER/bin:$PATH
         ./_test/bootstrap.sh
         ;;
     esac
@@ -105,7 +106,6 @@ script:
         done
         ;;
       cabal)
-        export PATH=/opt/cabal/$CABALVER/bin:/opt/ghc/$GHCVER/bin:$PATH
         runhaskell -Wall ./_test/check-exercises.hs
         ;;
     esac
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,50 +1,111 @@
 # The Travis docs (https://docs.travis-ci.com/user/languages/haskell)
 # suggest that we use `language: haskell` and `ghc:` to specify GHC versions.
 # However, as of this writing, Travis only supported versions up to 7.8.
+# In addition, `language: haskell` doesn't support using Stack, which we use.
 # If Travis supports all versions we would like to test, then we could switch.
-# Until then, we use the https://github.com/hvr/multi-ghc-travis setup.
+# Until then, we use a hybrid of the https://github.com/hvr/multi-ghc-travis and
+# https://www.fpcomplete.com/blog/2016/02/updated-haskell-travis-config setups.
 language: c
-sudo: false
-addons:
-  apt: &apt
-    sources:
-      - hvr-ghc
-cabal: &cabal cabal-install-1.22
-env:
-  global:
-    - CABALVER=1.22
-matrix:
-  include:
-    - env: GHCVER=7.6.3
-      addons:
-        apt:
-          <<: *apt
-          packages: [ghc-7.6.3, *cabal]
-    - env: GHCVER=7.8.4
-      addons:
-        apt:
-          <<: *apt
-          packages: [ghc-7.8.4, *cabal]
-    - env: GHCVER=7.10.3
-      addons:
-        apt:
-          <<: *apt
-          packages: [ghc-7.10.3, *cabal]
-    - env: GHCVER=8.0.1
-      addons:
-        apt:
-          <<: *apt
-          packages: [ghc-8.0.1, *cabal]
+sudo: false                      # Use the container-based infrastructure.
+
 cache:
   directories:
-    - '$HOME/.ghc'
-    - '$HOME/.cabal'
+    - $HOME/.ghc                 # For BUILD=cabal
+    - $HOME/.cabal               # For BUILD=cabal
+    - $HOME/.stack               # Global stack's cache.
+    - $HOME/.foldercache         # Per exercise `.stack-work` cache.
+
+matrix:
+  include:
+    # These are builds using cabal-install to install packages.
+    # The primary motivation is GHC 7.6 support (Stack does not support 7.6).
+    # We would like to keep support of 7.6 since various Linux distros still use it:
+    # - Debian stable (jessie)
+    # - Ubuntu 14.04 LTS (trusty)
+    # - Fedora 20 and 21.
+    # 7.6 is the minimum version we can test since our test runner uses mkdtemp from System.Posix.Temp.
+    # This was only added in unix 2.6.0.0, which ships with GHC 7.6.x.
+    - env: BUILD=cabal CABALVER=1.16 GHCVER=7.6.3
+      addons: {apt: {packages: [cabal-install-1.16, ghc-7.6.3], sources: [hvr-ghc]}}
+    - env: BUILD=cabal CABALVER=1.18 GHCVER=7.8.4
+      addons: {apt: {packages: [cabal-install-1.18, ghc-7.8.4], sources: [hvr-ghc]}}
+    - env: BUILD=cabal CABALVER=1.22 GHCVER=7.10.3
+      addons: {apt: {packages: [cabal-install-1.22, ghc-7.10.3], sources: [hvr-ghc]}}
+    - env: BUILD=cabal CABALVER=1.24 GHCVER=8.0.1
+      addons: {apt: {packages: [cabal-install-1.24, ghc-8.0.1], sources: [hvr-ghc]}}
+
+
+    # These are builds using Stack to install packages.
+    # The advantages:
+    # - reproducible builds
+    # - verification of per-exercise dependencies.
+    - env: BUILD=stack RESOLVER=lts-2.22 # Last GHC 7.8.x snapshot.
+      addons: &stack-addons
+        apt:
+          packages:
+            - libgmp-dev # Stack's GHC depends on this.
+    - env: BUILD=stack RESOLVER=lts-6.4 # Last checked LTS snapshot (GHC 7.10.3).
+      addons:
+        <<: *stack-addons
+    - env: BUILD=stack RESOLVER=nightly-2016-06-21 # Last checked nightly snapshot (GHC 8.0.1).
+      addons:
+        <<: *stack-addons
+    - env: BUILD=stack RESOLVER=lts # Latest LTS snapshot.
+      addons:
+        <<: *stack-addons
+    - env: BUILD=stack RESOLVER=nightly # Latest nightly snapshot.
+      addons:
+        <<: *stack-addons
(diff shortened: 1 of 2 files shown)


petertseng:
No longer WIP, let's take a look.

So we'll have the hermetic builds and dependency checking provided by Stack, but do not leave 7.6 behind.

One big question is whether the extra complexity is maintainable, which we should discuss.

It is possible that we can have 7.6 be the *only* version that uses cabal-install (Note that the current iteration of this PR tests 7.6, 7.8, 7.10, 8.0 with cabal-install, which may potentially be gratuitous.)

.travis.yml
```diff
@@ -1,50 +1,88 @@
...
 script:
-  - 'export PATH=/opt/cabal/$CABALVER/bin:/opt/ghc/$GHCVER/bin:$PATH'
-  - 'runhaskell -Wall ./_test/check-exercises.hs'
-  - './bin/fetch-configlet'
-  - './bin/configlet .'
+  - |
+    configlet .
+    case "$BUILD" in
+      stack)
+        set -e
```
rbasso: If `configlet` fails, why continue testing?

_test/bootstrap.sh
```diff
@@ -2,12 +2,6 @@
 # This installs a the prerequisites for a given GHC/Cabal config
 set -x
 set -e
-if [ ! -z "$GHCVER" ]; then
```
petertseng: no longer needed as now handled by `before_install`. Can make it a separate commit if desired, though it will be a bit tricky as there will be one commit that has `PATH` being set in `script/cabal`, then one that moves it up to `before_install`

.travis.yml
petertseng: I take it this means we should move `set -e` above `configlet .` I agree.

.travis.yml
```diff
@@ -1,50 +1,88 @@
...
+        ;;
+    esac
+
 install:
-  - true
+  - |
+    travis_retry fetch-configlet
+    if [ "$BUILD" = "stack" ]; then
+      travis_retry curl -L https://www.stackage.org/stack/linux-x86_64 -o pack.tgz
+      tar xzf pack.tgz --wildcards --strip-components=1 -C ${HOME}/bin '*/stack'
```
rbasso: `tar` will even if  `fetch-configlet` and `curl -L` fail?

petertseng:
updated to move `set -e` above `configlet`.

```diff
$ git diff e4b68e6 14f272b
diff --git a/.travis.yml b/.travis.yml
index c9fb04f..759ba58 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,10 +64,10 @@ install:
 
 script:
   - |
+    set -e
     configlet .
     case "$BUILD" in
       stack)
-        set -e
         for exercise in ${TRAVIS_BUILD_DIR}/exercises/* ; do
             pushd ${exercise}
             # `stack --word-dir` fails if not targeting a subdirectory, so we just
```

.travis.yml
petertseng: It might be best to fix that in #162 but I can do it here too. Trying to decide whether I should amend it into 9b73dee05da4bf7132c89dbeeed3ea7245737d3c (yours), have a separate one, or amend it into 14f272b0ead0ee61242b6a7662a8b078147b563f (mine), but perhaps I split hairs here

_test/bootstrap.sh
rbasso: I will try to completely rewrite `bootstrap.sh` to use `dependencies` if I get the time. In #162 I removed it but, if we want to keep tests with cabal, we need to at least to centralize the data in `dependencies.txt`.

.travis.yml
petertseng: ok, just added a commit on

_test/bootstrap.sh
petertseng: Seems like a good idea.

.travis.yml
rbasso: In #162 I used distinct lines of `install` so that Travis automatically fails everything if a line fails. When you moved the commands inside the `if` Travis cannot check intermediary steps.

rbasso: You don't have to base this PR on #162.
The change are so significant that maybe it's better to just base it directly on master.

petertseng: aha! a very subtle distinction that I did not notice! Thanks for the catch, then.

petertseng: in that case I'll remove the intermediate commit

petertseng: Maybe I should use `&&` rather than `set -e` here, hmm

petertseng:
Moved some things to separate lines if possible.

```diff
$ git diff faf9502 247f0cf
diff --git a/.travis.yml b/.travis.yml
index 1b596f4..5b8a467 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,8 @@ matrix:
   fast_finish: true                     # is released, so we list them here.
 
 before_install:
+  - export PATH="${TRAVIS_BUILD_DIR}/bin:$PATH"       # For {,fetch-}configlet.
   - |
-    export PATH="${TRAVIS_BUILD_DIR}/bin:$PATH"       # For {,fetch-}configlet.
     case "$BUILD" in
       stack)
         mkdir -p ${HOME}/bin                          # Create folder for stack.
@@ -55,18 +55,18 @@ before_install:
     esac
 
 install:
+  - travis_retry fetch-configlet
   - |
     set -e
-    travis_retry fetch-configlet
     if [ "$BUILD" = "stack" ]; then
       travis_retry curl -L https://www.stackage.org/stack/linux-x86_64 -o pack.tgz
       tar xzf pack.tgz --wildcards --strip-components=1 -C ${HOME}/bin '*/stack'
     fi
 
 script:
+  - configlet .
   - |
     set -e
-    configlet .
     case "$BUILD" in
       stack)
         for exercise in ${TRAVIS_BUILD_DIR}/exercises/* ; do
```

.travis.yml
```diff
@@ -1,50 +1,89 @@
...
+
 script:
-  - 'export PATH=/opt/cabal/$CABALVER/bin:/opt/ghc/$GHCVER/bin:$PATH'
-  - 'runhaskell -Wall ./_test/check-exercises.hs'
-  - './bin/fetch-configlet'
-  - './bin/configlet .'
+  - configlet .
+  - |
+    set -e
+    case "$BUILD" in
```
rbasso: In `script:` Travis has a special behavior. If a command fails, Travis will fail everything, but it will not shortcut evaluation, so every entry will be executed.

I used `set -e` and only one command in #162 to avoid that.

petertseng: Ah interesting. Would we like this to happen? Consider this case:

If I'm adding a new exercise and have forgotten to add it to config.json AND its tests are failing: Maybe I want to see both of those things. If we stop the build immediately on `configlet` failing, then I have to fix that first, and only after I do that will I find out my tests are failing too. Is that OK?

Then again, the argument may be "well maybe I should have run the tests locally to see whether they were passing" which... is valid. Maybe the tests only fail on an old version of GHC though?

petertseng: editing into previous comment:

I'm OK with either of these behaviors, but we should make the explicit decision.

.travis.yml
petertseng: OK, I'll not base it on #162 so that _bootstrap.sh doesn't get removed and then re-added, which was the main weird thing about basing this on #162.

.travis.yml
rbasso: I prefer to fail ASAP:

- Most of the time we are waiting for Travis, so faster is better.
- Usually the changes are small, so we have just one error.
- Travis is just a final check, it should be tested locally.
- It saves resources for other projects.
- It marginally decreases global warming. :grimacing:

petertseng:
OK, I pretty much squashed #162 and this together, this avoids the delete and then re-add of bootstrap.sh. The only noticeable change is I also tried to add explanatory comments. The functionality as Travis sees it should not change.

.travis.yml
petertseng: OK, in that case I'll move configlet back into the single command, after `set -e`

petertseng:
Almost certainly has merge conflicts with #179, feel free to merge #179 first

petertseng:
Rebased now that #179 is in.

I'll assume we want to keep the `cabal-install` 7.8, 7.10, 8.0 unless I hear otherwise.

It's worth asking because, well, if the only reason we want to keep the `cabal-install` in Travis is to keep 7.6, then we should ask whether we are getting value out of the 7.8, etc.
It's possible there are other reasons we keep them though. Need to think that a bit more.

rbasso:
About keeping the other GHC versions with cabal, I don't see the need, but that doesn't bother me at all.
We can leave the way it is just to see if we get luck and just one of the tests fail. That would be interesting...

PR #176, opened 2016-07-03

### LINKED ISSUE #175: _test/bootstrap.sh should use _test/dependencies.txt

petertseng:

Closes #175

### DIFF
--- a/_test/bootstrap.sh
+++ b/_test/bootstrap.sh
@@ -8,20 +8,20 @@ fi
 if [ ! -z "$CABALVER" ]; then
     export PATH=/opt/cabal/${CABALVER}/bin:$PATH
 fi
+
+# What directory is this script in?
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+deps=$(
+  cut -d: -f2 "$DIR/dependencies.txt" |
+  xargs echo |
+  tr " " "\n" |
+  sort |
+  uniq |
+  # Exclude packages included in GHC
+  grep -v 'base\|array\|containers\|time'
+)
+
 cabal update
-# This is a fairly minimal set
-cabal install \
-      primitive \
-      random \
-      tf-random \
-      HUnit \
-      QuickCheck \
-      split \
-      text \
-      bytestring \
-      attoparsec \
-      vector \
-      parallel \
-      stm \
-      old-locale \
-      lens
+
+echo "$deps" | xargs cabal install


petertseng:
Completely untested so I probably will not get it right on the first try. Oh well, let's see what we got.

_test/bootstrap.sh
```diff
@@ -8,20 +8,20 @@ fi
...
-      attoparsec \
-      vector \
-      parallel \
-      stm \
-      old-locale \
-      lens
+
+for dep in $unique_deps; do
+  # Don't update packages included in GHC
+  if [ "$dep" = "base" -o "$dep" = "array" -o "$dep" = "time" ]; then
```
petertseng: this is a "wat". how can we improve this?

_test/bootstrap.sh
```diff
@@ -1,27 +1,25 @@
...
-      attoparsec \
-      vector \
-      parallel \
-      stm \
-      old-locale \
-      lens
+
+for dep in $deps; do
+  # Don't update packages included in GHC
+  if [ "$dep" = "base" -o "$dep" = "array" -o "$dep" = "containers" -o "$dep" = "time" ]; then
```
petertseng: this is a "wat". how can we improve this? I would not like have to hard-code these.

petertseng:
I'm stunned that it appear that with a single cabal installs it worked but with individual cabal installs we apparently fail on old-locale

_test/bootstrap.sh
```diff
@@ -8,20 +8,20 @@ fi
...
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+deps=$(
+  cut -d: -f2 "$DIR/dependencies.txt" |
+  xargs echo |
+  tr " " "\n" |
+  sort |
+  uniq |
+  # Exclude packages included in GHC
+  grep -v 'base\|array\|containers\|time'
```
petertseng: this is a "wat". how can we improve this? I would not like have to hard-code these.

petertseng:
Just a second there, going to close this one and open a new one, ince I have a suspicion about something I did to Travis.

_test/bootstrap.sh
rbasso: I think this behavior is *by design*. When you ask for a simple package, it will try to install the newest one possible. If you supply multiple packages, it will try to satisfy all restrictions at the same time.

You should probably use a single `cabal install`.

PR #177, opened 2016-07-03

### LINKED ISSUE #175: _test/bootstrap.sh should use _test/dependencies.txt

petertseng:

Closes #175

### DIFF
--- a/_test/bootstrap.sh
+++ b/_test/bootstrap.sh
@@ -8,20 +8,20 @@ fi
 if [ ! -z "$CABALVER" ]; then
     export PATH=/opt/cabal/${CABALVER}/bin:$PATH
 fi
+
+# What directory is this script in?
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+deps=$(
+  cut -d: -f2 "$DIR/dependencies.txt" |
+  xargs echo |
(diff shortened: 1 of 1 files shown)


_test/bootstrap.sh
```diff
@@ -8,20 +8,20 @@ fi
...
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+deps=$(
+  cut -d: -f2 "$DIR/dependencies.txt" |
+  xargs echo |
+  tr " " "\n" |
+  sort |
+  uniq |
+  # Exclude packages included in GHC
+  grep -v 'base\|array\|containers\|time'
```
petertseng: reminder that this sucks and I want a better way

petertseng:
Unbelievable. Well, it appears I need a bit more education before pursuing this one. Closing it for the time being while I pursue that education, anyone reading this code is free to use (or not) what's here and beat me to the punch in achieving this goal.

_test/bootstrap.sh
rbasso: Instead of trying to do it by yourself, you can run ``dependencies `ls exercises` ``.

After that, it would be nice to exclude the packages already installed. Maybe `ghc-pkg list` with some options can solve this.

PR #178, opened 2016-07-03
empty commit. DO NOT MERGE

petertseng:


petertseng:
sanity check - travis can't be broken on master, now can it? So why the difference bewtween thi and #177, hmm?

PR #179, opened 2016-07-03

### LINKED ISSUE #175: _test/bootstrap.sh should use _test/dependencies.txt

petertseng:

Closes #175

### DIFF
--- a/_test/bootstrap.sh
+++ b/_test/bootstrap.sh
@@ -8,20 +8,15 @@ fi
 if [ ! -z "$CABALVER" ]; then
     export PATH=/opt/cabal/${CABALVER}/bin:$PATH
 fi
-cabal update
-# This is a fairly minimal set
-cabal install \
-      primitive \
-      random \
-      tf-random \
-      HUnit \
-      QuickCheck \
-      split \
-      text \
-      bytestring \
-      attoparsec \
-      vector \
-      parallel \
-      stm \
-      old-locale \
-      lens
+
+# What directory is this script in?
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+deps=$($DIR/dependencies $(ls $DIR/../exercises))
+already_installed=$(ghc-pkg list --names-only --simple-output | tr " " "\n" | sort)
+not_installed=$(comm -23 <(echo "$deps") <(echo "$already_installed"))
+
+if [ -n "$not_installed" ]; then
+  cabal update
+  echo "$not_installed" | xargs cabal install
+fi
--- a/_test/bootstrap.sh
+++ b/_test/bootstrap.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 # This installs a the prerequisites for a given GHC/Cabal config
-set -x
 set -e
 if [ ! -z "$GHCVER" ]; then
     export PATH=/opt/ghc/${GHCVER}/bin:$PATH
@@ -18,5 +17,11 @@ not_installed=$(comm -23 <(echo "$deps") <(echo "$already_installed"))
 
 if [ -n "$not_installed" ]; then
   cabal update
+
+  echo "$(echo "$not_installed" | wc -l) packages to be installed:"
+  echo "$not_installed"
+
   echo "$not_installed" | xargs cabal install
+else
+  echo "all packages installed already"
 fi


_test/bootstrap.sh
```diff
@@ -8,20 +8,32 @@ fi
...
+  # Exclude packages included in GHC
+  # TODO: This is obviously not ideal
+  # what if the list of packages included with GHC changes?
+  # How can we avoid having to hard-code this list?
+  grep -v 'base\|array\|containers\|time'
+)
+
+if echo "$deps" | grep -q 'old-locale'; then
+  # Move old-locale to the bottom.
+  # Unfortunate, but without this cabal-install tries to update old-locale.
```
petertseng: If you ask me to explain why, I have no idea.

petertseng: Maybe I should grab a machine with GHC 7.6 or 7.8 and investigate, because an explanation would be great.

_test/bootstrap.sh
```diff
@@ -8,20 +8,23 @@ fi
...
+# What directory is this script in?
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+deps=$($DIR/dependencies $(ls $DIR/../exercises))
+already_installed=$(ghc-pkg list --names-only --simple-output | tr " " "\n" | sort)
+not_installed=$(comm -23 <(echo "$deps") <(echo "$already_installed"))
+
+if echo "$not_installed" | grep -q 'old-locale'; then
+  # Move old-locale to the bottom.
+  # Unfortunate, but without this cabal-install tries to update old-locale.
```
petertseng: will investigate whether this is still true with `not_installed`

_test/bootstrap.sh
rbasso: To compare lists of itens, `comm` from *GNU coreutils* is usually a good choice. Try this:

```bash
comm -23 <(dependencies `ls exercises`) <(ghc-pkg list --simple-output --names-only | xargs -n1 | sort)
```

In my local setup I cannot test it easily, but it will probably work.

_test/bootstrap.sh
petertseng: pretty sure it won't be. Safe to remove.

_test/bootstrap.sh
petertseng: although the `xargs -n1` is one thing I was missing. I settled on `tr " " "\n"` instead, guess they do the same thing

petertseng:
OK, testing a few things on Travis leads me to think this will be OK now.

The duplicate PRs with this one are unfortunate, should test things out on Travis on my fork of the repo rather than here, didn't get that idea until while working on this PR.

petertseng:
On reviewing Travis logs:

`set -x` is pretty noisy. Do we want to keep it/

rbasso:
Don't worry about the extra PRs...I did the same a few days ago.
I don't like `set -x` too much in production, but I don't care so much about it. You can decide depending on how much you trust the code will run correctly.

petertseng:
I guess my main complaint is that we see the contents of `$deps` and `$already_installed` twice each: Once when they get set, and again when they get interpolated in the `echo`.

We probably don't even need to see `deps` since that can be reproduced by locally running `dependencies`. I'm not sure `already_installed` gets much use.

Based on this, perhaps remove `set -x` and just print out `$not_installed` will do.

_test/bootstrap.sh
```diff
@@ -1,27 +1,24 @@
...
-      lens
+
+# What directory is this script in?
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+deps=$($DIR/dependencies $(ls $DIR/../exercises))
+already_installed=$(ghc-pkg list --names-only --simple-output | tr " " "\n" | sort)
+not_installed=$(comm -23 <(echo "$deps") <(echo "$already_installed"))
+
+echo "Not installed packages ($(echo "$not_installed" | wc -l)):"
```
petertseng: sigh, off by one error. `Not installed packages (1):` when `not_installed` == "'`

petertseng: `echo -n` it is.

petertseng:
Heh, `echo -n` undercounts by 1 when there is a nonempty list of packages to update (tested by killing caches on my fork). Going to have to just do different things based on `-z "$not_installed"`.

I am satisfied with this now.

rbasso:
Ok. I'll clear all the caches and rerun the Travis builds just to be sure everything is fine before merging this one.

rbasso:
Would you mind if I squash those two commits in a single one?

petertseng:
OK with me to squash

PR #183, opened 2016-07-04
Simple Linked List: Add a handful of tests that assert isNil returns False

cjlarose:
Otherwise, an implemenation of isNil that returned True all the time
would pass the tests


rbasso:
Nice catch! Thanks! :smile:

PR #184, opened 2016-07-06
Place phone-number exercise before nucleotide-count exercise

cjlarose:
I'm open to discussion on this, but I that since Nucleotide Count was changed to return Either (#165) and Phone Number was changed to return Maybe (#163), it makes sense to swap the order. In introductory Haskell materials, such as LYAH, `Maybe` typically comes before `Either`, and I think it makes sense because conceptually Maybe is simpler.


rbasso:
Thank you, @cjlarose! :smile:

PR #189, opened 2016-07-08
Transform exercises in Stack projects - Phase 4 - Migrating all exercises - Part 1

rbasso:
###### Phase 4 - Migrating all exercises - Part 1

Migrate to stack projects:

- nth-prime
- palindrome-products
- pascals-triangle
- pig-latin
- pythagorean-triplet
- saddle-points
- say
- secret-handshake
- series
- sieve
- simple-cipher
- connect
- parallel-letter-frequency
- sgf-parsing
- go-counting
- zipper
- forth
- lens-person

Fix comment in `exercises/pov/package.yaml`.

Add incomplete stub solutions:

- nth-prime
- pythagorean-triplet
- saddle-points
- say
- secret-handshake
- series
- sieve

Add stub solutions based on test suite analysis:

- connect
- pascals-triangle
- pig-latin
- simple-cipher

Add stub solutions based on test instructions:

- go-counting
- palindrome-products
- parallel-letter-frequency
- sgf-parsing

Update stub solutions to allow compilation:

- lens-person


petertseng:
I feel I should say real quick that here and in `pov` as well, "you want to use in you solution." should probably be "you want to use in your solution."

exercises/pythagorean-triplet/src/Triplet.hs
```diff
@@ -0,0 +1,7 @@
+module Triplet (isPythagorean, mkTriplet, pythagoreanTriplets) where
+
+isPythagorean = undefined
+
+mkTriplet =undefined
```
petertseng: probably good to add extra space after `=`, consistency

rbasso:
Thank you for the review! I will fix it, but merging will probably be delayed until we finish Phase 2 (Online testing `pov`).

exercises/sieve/package.yaml
```diff
@@ -0,0 +1,20 @@
+name: sieve
+
+dependencies:
+  - base
+
+library:
+  exposed-modules: Sieve
+  source-dirs: src
+  dependencies:
+    - vector
```
petertseng: This leads me to see an interesting point! (This may apply to exercises other than sieve as well)

This `vector` is here because the example solution uses it, right? It is not necessarily the case that every submitted solution will use it. 

Are we OK with leaving them in here?

Then again, the alternative is to have some custom thing that splices in the example solution's dependencies when we go to Travis, which is not necessarily something we want to tackle.

So our answer might be "We're OK with it", as long as students don't think their solution must use `vector`.

My personal answer is probably to keep it as is for now, see whether any problems come up

rbasso: This is a problem for which I have no good answer. Maybe this will force us to give more attention to the example solutions, as they will define the packages available by default.

It's a design flaw. My bad! :grin:

PR #196, opened 2016-07-11
all-your-base: Fix test formatting

dbalmain:
Make it possible to easily comment out the test cases so that the problem can be
solved iteratively by moving the commas in the vertical list to the start of the
line.


rbasso:
Thank you, @dbalmain ! :smile:

PR #197, opened 2016-07-12
grains: Change return type to Maybe

rbasso:
Currently, the exercise `grains` has a function `square :: Int -> Integer` that returns the number of grains of wheat on sequentially numbered squares on a chessboard. The problem doesn't specify what should be returned if the number given is not in the chessboard.

This is a great opportunity to make it more idiomatic and get the users acquainted with `Maybe`.
I propose to to the following: 

1. Change the return type of function `square` from an `Integer` to a `Maybe Integer`.
2. Add three test:
 - `square (-1)`
 - `square   0`
 - `square  65`
3. Add a stub solution.

### DIFF
--- a/exercises/grains/Grains.hs
+++ b/exercises/grains/Grains.hs
@@ -1,0 +1,5 @@
+module Grains (square, total) where
+
+square = undefined
+
+total = undefined
--- a/exercises/grains/example.hs
+++ b/exercises/grains/example.hs
@@ -1,7 +1,12 @@
 module Grains (square, total) where
 
-square :: Int -> Integer
-square = (2 ^) . pred
+import Data.Maybe (fromJust)
+
+square :: Integer -> Maybe Integer
+square x
+    | x < 1     = Nothing
+    | x > 64    = Nothing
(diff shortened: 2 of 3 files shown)


petertseng:
Change seems good here, :+1: 

Might be good to see if this should be specified in x-common?

rbasso:
Maybe we don't need to change there:

- There is no mention in `exercism/x-common/grains.md` about values out of the board.
- There is no `exercism/x-common/grains.json`.

Should we write something in `exercism/x-common/grains.md`?

petertseng:
Yeah we can see if other tracks might want to check for out of bounds values like here. Maybe they'll like it, in which case it can be standard for all tracks. Or maybe they won't, an argument against might be https://github.com/exercism/x-common/pull/287#issuecomment-231602657 , but that means it'll just be a specific characteristic of this track.

rbasso:
I think that `grains.md` is good enough the way it is, but I understand that it would be nice to exposed the other tracks to out-of-bound cases. Maybe a `exercism/x-common/grains.json` marking the out-of-bound cases as implementation specific could be the solution.

PR #198, opened 2016-07-12
etl: Change letter representation to Char

rbasso:
The `etl` exercise uses `String` to represent each letter in both, the input and the output of `transform`.
This is unidiomatic because, in Haskell, a single haracter is not the same type as a list of characters.

To avoid exposing users to unidiomatic code and ease understanding of what is being asked in this exercise, I propose the following changes:

1. Change letter representation from String to Char.
2. Add stub solution.

### DIFF
--- a/exercises/etl/ETL.hs
+++ b/exercises/etl/ETL.hs
@@ -1,0 +1,6 @@
+module ETL (transform) where
+
+import Data.Map (Map)
+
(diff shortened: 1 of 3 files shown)


petertseng:
:+1:

rbasso:
Thanks for the review, @petertseng ! :+1: :smile:

PR #199, opened 2016-07-12
hamming: Change return to Maybe and match json

rbasso:
- Change return type of `distance` to a *Maybe*.
- Rewrite tests to match `exercism/x-common/hamming.hs`.

### DIFF
--- a/exercises/hamming/Hamming.hs
+++ b/exercises/hamming/Hamming.hs
@@ -1,0 +1,3 @@
+module Hamming (distance) where
+
+distance = undefined
--- a/exercises/hamming/example.hs
+++ b/exercises/hamming/example.hs
@@ -1,4 +1,6 @@
 module Hamming (distance) where
 
-distance :: String -> String -> Int
-distance a b = sum . map fromEnum $ zipWith (/=) a b
+distance :: Integral a => String -> String -> Maybe a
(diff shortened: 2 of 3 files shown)


petertseng:
:+1:

rbasso:
I think the `example.hs` is kind of ugly. Any idea to improve it?

petertseng:
What would be closer to your ideal?

There is the option of checking the length beforehand instead of during, with something like:

```haskell
distance a b = if length a == length b then Just . sum . map fromEnum $ zipWith (/=) a b else Nothing
```

But this is pretty much just a little extra stuff slapped onto the previous solution. And this only work with `distance :: String -> String -> Maybe Int` rather than `distance :: Integral a => String -> String -> Maybe a` because `fromEnum` is `a -> Int`

(I also sometimes wonder whether it traverses `a` and `b` one more time than is necessary - one to find the length, then once more to compare them, but I don't know if everyone thinks about such things)

I may think of other ideas later, given some criteria

petertseng:
You could move the `if x /= y` into a guard if you so desire.

rbasso:
Thanks, @petertseng!

I also feel it traverses the lists two times, but I don't know enough to be sure.
I'll merge it the way it is. If you find any solution you think is good, just open the PR and I'll merge it. The same, of course, applies to other exercises. :+1:

rbasso:
Just saw your last message.

PR #200, opened 2016-07-13
Transform exercises in Stack projects - Phase 4 - Migrating all exercises - Part 2

rbasso:
###### Phase 4 - Migrating all exercises - Part 2

Migrate to stack projects:

- custom-set
- linked-list
- food-chain
- wordy
- trinary
- octal
- ocr-numbers
- minesweeper
- house
- matrix
- clock
- luhn
- largest-series-product
- hexadecimal
- difference-of-squares
- binary-search-tree
- queen-attack
- robot-simulator
- kindergarten-garden
- crypto-square
- bank-account
- all-your-base
- atbash-cipher
- allergies
- raindrops
- prime-factors
- binary
- roman-numerals
- scrabble-score
- triangle

Add incomplete stub solutions:

- allergies
- bank-account
- binary
- binary-search-tree
- clock
- custom-set
- difference-of-squares
- hexadecimal
- largest-series-product
- linked-list
- luhn
- matrix
- octal
- raindrops
- robot-simulator
- roman-numerals
- scrabble-score
- triangle
- trinary
- wordy

Add complete stub solutions:

- atbash-cipher
- crypto-square
- food-chain
- house
- kindergarten-garden
- minesweeper
- ocr-numbers
- prime-factors
- queen-attack


exercises/queen-attack/src/Queens.hs
```diff
@@ -0,0 +1,7 @@
+module Queens (boardString, canAttack) where
+
+boardString :: Maybe (Int, Int) -> Maybe (Int, Int) -> String
+boardString = undefined
```
petertseng: it may not be immediately obvious that the first is white and the second is black, but maybe that's OK it's easily discoverable from reading the tests. unless the goal is to make it obvious without having to read the tests

rbasso: I didn't spend too much time writing the *stubs*, so most of them can probably be improved.
In this one, I just wrote the signature that was compatible with the tests.

From the signature, it's not possible to know that:

- The first is White and the second is Black, as you pointed out. 
- The indexes are (column, row)
- The board starts at (0,0), which is probably completely unnatural for a chess player.

> ... but maybe that's OK it's easily discoverable from reading the tests. unless the goal is to make it obvious without having to read the tests

I still don't know the answer to that. At first, I was just trying to save the users the repetitive work writing boilerplate code, but maybe that is a good idea...

Do you think we should change it to avoid the need to look at the tests?
If you write a PR, I will merge it.

PR #209, opened 2016-07-19
accumulate: fix inefficient example

petertseng:
The intent, I believe, is to show an example that would work and gives
correct results but fails the non-strictness test. It was added in
d238a331c2bccaa1d86522198e0a85235a6356d6.

In order for it to compile, the `xs` needs to be removed (an eta
reduction has happened).

In order for the results to be correct, final `acc` must be reversed.


rbasso:
Nice catch!

Issue #194, opened 2016-07-10
discussion: Making the Haskell track more idiomatic


rbasso:
###### Making the Haskell track more idiomatic

With #115 we started to review some exercises to make them feel more like Haskell. As a result, we did the following:

- Changed the return type of `rna-transcription` to a `Maybe`
- Changed the return type of `phone-number` to an `Either String`
- Changed the return type of `nucleotide-count` to a `Maybe`

There are many other unidiomatic things that can appear in the track:

- Single letters represented as *Strings*.
- Using *Int* when *Integer* makes more sense.
- Use *Integer* when any *Integral* can be used.
- Partial functions instead of returning *Maybe*.
- Requiring lists when any *Foldable* is possible.
- ...

This doesn't means that we should change everything that is not idiomatic, because sometimes unidiomatic code can be easier for beginners. But we should not allow it when it it not clear that it is the case.

###### List of unidiomatic things in the track we could fix

- [x] ~~`grains` is not a total function. Could return a *Maybe*. (#197)~~
- [x] ~~`hamming` is not defined for unequal lists. Could return a *Maybe*. (#199)~~
- [x] ~~`etl` uses *String* to represent a single letter. Should use *Char*. (#198)~~
- [ ] `roman-numerals` could return a *Maybe*.
- [ ] `crypto-square` returns a matrix as a String, with spaces representing row breaks and missing elements. (#259)

###### List of things we could do to promote idiomatic programming

- [x] ~~`strain`: change tests to enforce lazy processing of the lists. (#194)~~

##### Anyone knows of any specific thing that could/should be changed to make this track more idiomatic?

rbasso:
User's comment about `simple-linked-list`:

> Ok. Maybe in Ruby or other OO brain damaged languages it's fun.
> But why do I need to reinvent Prelude here? Dumbest task ever.

rbasso:
Should `roman-numerals` return a *Maybe* or *Either* if *n > 3000* or *n < 1*?

petertseng:
I would find `Maybe String` good.

Note that if truly desired, 3999 could be allowed.

rbasso:
> Note that if truly desired, 3999 could be allowed.

Makes perfect sense. :smile: 
We should also change `3000` to `3999` in `x-common`.

rbasso:
Considering that is issue is open for a while and that we have opened issues to deal with all the agreed changes, I'm closing this PR now, so that we can focus on the more practical issues.

[...]

PR #210, opened 2016-07-19
strain: Add test for non-strict keep and discard

### LINKED ISSUE #208: Should strain test non-strict evaluation, like accumulate does?
We can make this part of #194.

Note that the [accumulate tests](https://github.com/exercism/xhaskell/blob/master/exercises/accumulate/test/Tests.hs) have a test that make sure that the accumulate function is non-strict. That was added in https://github.com/exercism/xhaskell/commit/d238a331c2bccaa1d86522198e0a85235a6356d6

Is this a goal of the track that we promote non-strict evaluation when appropriate? And if so, do we want to add a similar test to [strain](https://github.com/exercism/xhaskell/blob/master/exercises/strain/test/Tests.hs)?

They might look something like this:

```haskell
  , testCase "keep non-strict" $ do
    let ws = "yes" : error "no"
    ["yes"] @=? take 1 (keep (const True) ws)
  , testCase "discard non-strict" $ do
    let ws = "yes" : error "no"
    ["yes"] @=? take 1 (discard (const False) ws)
```

Would want a more descriptive message than "no", of course.

petertseng:
Just as d238a331c2bccaa1d86522198e0a85235a6356d6 added a test for
accumulate to be non-strict, we add a test for keep and discard that
does the same.

For the purpose of #194
Closes #208

### DIFF
--- a/exercises/strain/test/Tests.hs
+++ b/exercises/strain/test/Tests.hs
@@ -46,4 +46,10 @@ strainTests =
   , testCase "discard strings" $ do
     let ws = ["apple", "zebra", "banana", "zombies", "cherimoya", "zealot"]
     ["apple", "banana", "cherimoya"] @=? discard (isPrefixOf "z") ws
+  , testCase "keep non-strict" $ do
+    let ws = "yes" : error "keep should be lazier - don't look at list elements you don't need!"
+    ["yes"] @=? take 1 (keep (const True) ws)
+  , testCase "discard non-strict" $ do
+    let ws = "yes" : error "discard should be lazier - don't look at list elements you don't need!"
+    ["yes"] @=? take 1 (discard (const False) ws)
   ]


exercises/strain/test/Tests.hs
```diff
@@ -46,4 +46,10 @@ strainTests =
   , testCase "discard strings" $ do
     let ws = ["apple", "zebra", "banana", "zombies", "cherimoya", "zealot"]
     ["apple", "banana", "cherimoya"] @=? discard (isPrefixOf "z") ws
+  , testCase "keep non-strict" $ do
+    let ws = "yes" : error "keep should be lazier - don't look at list elements you don't need!"
+    ["yes"] @=? take 1 (keep (const True) ws)
+  , testCase "discard non-strict" $ do
+    let ws = "yes" : error "discard should be lazier - don't look at list elements you don't need!"
+    ["yes"] @=? take 1 (discard (const False) ws)
```
rbasso: Nice error messages!

Issue #211, opened 2016-07-23
tests: Tracking migration to hspec with fail-fast.


rbasso:
##### Tracking migration to `hspec` with *fail-fast*

This issue is just for tracking. If you are interested in the discussion, please see #190.

##### Preparing

- [x] ~~Create branch `hspec-fail-fast`~~

##### Migrating all exercises

- Completely rewrite the test to use `hspec` with *fail-fast*.
- Update test cases to match `x-common`, if needed.
- Update `package.yaml`.
- Merge PR to ~~`hspec-fail-fast`~~ `master`

###### Exercises

- [x] leap
- [x] accumulate
- [x] sublist
- [x] grains
- [x] gigasecond
- [x] bob
- [x] sum-of-multiples
- [x] strain
- [x] hamming
- [x] rna-transcription
- [x] space-age
- [x] anagram
- [x] phone-number
- [x] nucleotide-count
- [x] grade-school
- [x] simple-linked-list
- [x] list-ops
- [x] word-count
- [x] etl
- [x] robot-name
- [x] meetup
- [x] beer-song
- [x] triangle
- [x] scrabble-score
- [x] roman-numerals
- [x] binary
- [x] prime-factors
- [x] raindrops
- [x] allergies
- [x] atbash-cipher
- [x] all-your-base
- [x] bank-account
- [x] crypto-square
- [x] kindergarten-garden
- [x] robot-simulator
- [x] queen-attack
- [x] binary-search-tree
- [x] difference-of-squares
- [x] hexadecimal
- [x] largest-series-product
- [x] luhn
- [x] clock
- [x] matrix
- [x] house
- [x] minesweeper
- [x] ocr-numbers
- [x] octal
- [x] trinary
- [x] wordy
- [x] food-chain
- [x] linked-list
- [x] custom-set
- [x] nth-prime
- [x] palindrome-products
- [x] pascals-triangle
- [x] pig-latin
- [x] pythagorean-triplet
- [x] saddle-points
- [x] say
- [x] secret-handshake
- [x] series
- [x] sieve
- [x] simple-cipher
- [x] connect
- [x] parallel-letter-frequency
- [x] sgf-parsing
- [x] go-counting
- [x] zipper
- [x] forth
- [x] lens-person
- [x] pov

##### Testing

- [X] ~~Clear *Travis-CI* cache.~~
- [X] ~~Check if build and rebuild times are reasonable.~~

##### Merging and cleaning up

[...]

PR #226, opened 2016-07-26
nucleotide-count: Rewrite to use hspec with fail-fast.

rbasso:
- Rewrite to use hspec with fail-fast.
- Change tests to make them more polymorphic.

Related to #211.

### DIFF
--- a/exercises/nucleotide-count/test/Tests.hs
+++ b/exercises/nucleotide-count/test/Tests.hs
@@ -1,47 +1,61 @@
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
+
+import Data.Either       (isLeft)
+import Data.Map          (fromList)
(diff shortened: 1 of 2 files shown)


exercises/nucleotide-count/test/Tests.hs
```diff
@@ -1,47 +1,61 @@
...
+
+          describe "nucleotideCounts" $ do
+
+            it "empty dna strand has no nucleotides" $
+              nucleotideCounts "" `matchesMap` [ ('A', 0)
+                                               , ('C', 0)
+                                               , ('G', 0)
+                                               , ('T', 0) ]
+
+            it "repetitive-sequence-has-only-guanosine" $
```
petertseng: why the dashes?

rbasso:
I have no idea! This came from the original implementation from 2013-08-19. I just adapted the test cases to use `hspec` here, without really looking at them.

exercises/nucleotide-count/test/Tests.hs
rbasso: I have no idea! This came from the original implementation from 2013-08-19. I just adapted the test cases to use `hspec` here, without really looking at them.

PR #235, opened 2016-07-26
meetup: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.

### DIFF
--- a/exercises/meetup/package.yaml
+++ b/exercises/meetup/package.yaml
@@ -17,4 +17,4 @@ tests:
     source-dirs: test
     dependencies:
       - meetup
-      - HUnit
+      - hspec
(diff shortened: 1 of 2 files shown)


petertseng:
I will say :+1: because the tests pass though I admit I didn't take a very close look due to the sheer number of the tests. I do wonder if quickcheck would be a good idea (especially now that octal and trinary our only exercises with quickcheck are getting deprecated), but that's nothing to do with hunit->hspec, a separate issue. I don't have time to do it myself for a while.

rbasso:
> I do wonder if quickcheck would be a good idea (especially now that octal and trinary our only exercises with quickcheck are getting deprecated).

Now that we are using Stack, we have no reason to reduce the number of dependencies. The only thing I don't like about *QuickCheck* is that random testing doesn't allow tests to be designed so that exercises would be solved iteractively, case after case, with increasing complexity.

PR #236, opened 2016-07-26
beer-song: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.

### DIFF
--- a/exercises/beer-song/test/Tests.hs
+++ b/exercises/beer-song/test/Tests.hs
@@ -1,39 +1,31 @@
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
+
+import Test.Hspec        (Spec, describe, it, shouldBe)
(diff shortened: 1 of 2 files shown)


exercises/beer-song/test/Tests.hs
```diff
@@ -1,39 +1,25 @@
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
-import Beer (sing, verse)
-
-exitProperly :: IO Counts -> IO ()
-exitProperly m = do
-  counts <- m
-  exitWith $ if failures counts /= 0 || errors counts /= 0 then ExitFailure 1 else ExitSuccess
+import  Test.Hspec        (Spec, describe, it, shouldBe)
```
petertseng: I just noticed the extra space, do all of the test files have this?

Okay, looked at triangle and it doesn't maybe accidental then?

rbasso:
Nice catch!

exercises/beer-song/test/Tests.hs
rbasso: Nice catch!

PR #239, opened 2016-07-26
roman-numerals: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.


petertseng:
Everything seems fine, :+1:

PR #240, opened 2016-07-26
binary: Rewrite tests to use hspec with fail-fast.

rbasso:
- Rewrite tests to use hspec.
- Update tests to match the json file.
- Fix test descriptions mentioning "errors".

Related to #211.

### DIFF
--- a/exercises/binary/package.yaml
+++ b/exercises/binary/package.yaml
@@ -16,4 +16,4 @@ tests:
     source-dirs: test
     dependencies:
       - binary
-      - HUnit
+      - hspec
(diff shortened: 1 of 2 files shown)


petertseng:
on the one hand, the behavior hasn't changed, so I shouldn't compain too much, but on the other hand, the tests now have descriptions that specifcally include the words "raises an error" yet an expectation of zero certainly isn't raising an error, so adding the descriptions makes the existing tests make less sense! Is it time to do something about this?

rbasso:
Well....this came from `x-common`, so we could change it there.
But the exercise will be deprecated anyway, so It's not really important to update `x-common`.

I'll change the strings just to avoid the confusing error messages!

rbasso:
Updated the descriptions mentioning *errors*

petertseng:
:+1:

Issue #192, opened 2016-07-10
discussion: Making the Haskell track more interesting for beginners


rbasso:
![exercises](https://cloud.githubusercontent.com/assets/16487165/16712630/5938a644-46c8-11e6-8148-028b0d778ceb.png)

We have added, changed and reordered exercises, but maybe we need a more rational approach to keep the track interesting.

At the moment, we are doing a few things to try to keep more beginners going on:

- [x] *Improving the documentation*.
- [x] *Moving exercises to Stack projects* - This makes easier to run the tests and deal with dependencies.
- [ ] Fail-fast the tests, so that only the first failed test will be displayed.
- [x] *Adding stub solutions* - To help users understand what exactly needs to be implemented.
- [ ] *Move instructions to `HINTS.md`* - It would be great if people could solve the exercises just looking at the instructions and results of the tests. Instructions do not belong to `Tests.hs` files of stub solutions.
- [ ] `beer-song`, `house` and `food-chain`could have working stub solutions.
- [ ] *Make the track more gradual* - Allow beginners to solve each new exercises without having to learn too much, but always demanding a little more knowledge.
- [ ] *Make the track more idiomatic* - Most people come to Haskell because it's a very beautiful programming language. When they are forced to write ugly code to solve an exercise they may leave.
- [x] *Remove pointless, repetitive and boring exercises*.

Some other ideas:

- *Display additional test data* - A failed test showing which function call failed would save the work of reading the test suite to search for the failed test case.

What do you think we could do?

petertseng:
> Remove pointless, repetitive and boring exercises.

Is it about time we deprecate binary, hexadecimal, trinary, octal?

There are at least three exercises which all involve generating a repetitive string: Do we want to keep all? food-chain, house, beer-song

rbasso:
> Is it about time we deprecate binary, hexadecimal, trinary, octal?

I was waiting for exercism/x-common#279 to deprecate them, because I have no idea about how to do it correctly:

kytrinyx commented 19 days ago 

> FYI @rbasso @NobbZ @ErikSchierboom @wobh @Cohen-Carlisle @yurrriq
>
> Before we can submit a blazon ticket we need to finish clarifying the process for deprecating exercises. Up until now it's been messy and prone to error, and I'd rather have that all straightened out before we deprecate four at a time across a bunch of language tracks. exercism/discussions#40

Is moving the exercises to `deprecated` in `config.json` enough? What does `foregone` means there? When should we use it?

[...]

Issue #115, opened 2016-04-07
discussion: Should we return Maybe when there is a possibility of invalid input?


petertseng:
While working on #114 I noticed that the binary tests just return 0 when there is an error.

That's simpler to implement. The disadvantage is that it makes the two cases "valid input whose value is zero" and "invalid input" indistinguishable to someone who can only observe the output values. Would we like to return `None`s for invalid inputs? I did get `Maybe`s added to largest-series-product in #89 .

Personally I would argue for it since I think it's a common thing in Haskell and more exposure to it is good, but there may be a possibility that there's too much. Wonder what others think.

To clarify scope, this is not suggesting that every exercise that possibly can test invalid inputs should (for example, "should the meetup exercise test nonexistent months") - that is a separate question that is worth another issue if someone wish to discuss it. The question is simply: For the exercises that *already* test invalid inputs, should they return `Maybe` as the result?

If this is done, at the very least all the base conversion exercises will want it (binary, trinary, octal, hexadecimal). I can possibly investigate the other exercises in Haskell to see where else it would make sense to add too.

Survey of existing exercises and how they handle invalid inputs:

* base-conversion exercises (binary, trinary, octal, hexadecimal) have already been mentioned - they return 0.
* nucleotide uses `error` https://github.com/exercism/xhaskell/blob/master/exercises/nucleotide-count/example.hs and the tests check for it via `Control.Exception`. Seems a big hammer, whereas I'd rather make the possibility of error apparent in the type system if we're going to check for it (it's one of the reasons why I use Haskell anyway)
* phone number also returns 0000000000, whereas it could return `Maybe`
* `say`, `wordy`, `zipper` already return `Maybe`s.

petertseng:
nucleotide uses `error` https://github.com/exercism/xhaskell/blob/master/exercises/nucleotide-count/example.hs and the tests check for it via `Control.Exception`. Seems a big hammer, whereas I'd rather make the possibility of error apparent in the type system if we're going to check for it (it's one of the reasons why I use Haskell anyway)

petertseng:
phone number also returns 0000000000, whereas it could return `Maybe`

petertseng:
`say` already returns `Maybe`s.

rbasso:
Those exercises fell awkward and force unidiomatic solutions. It's desirable, at least, to avoid promotion of bad coding practices.

[...]

PR #241, opened 2016-07-30
config: deprecate binary, trinary, octal, hexadecimal

petertseng:
Dicussion in #192 indicates it is safe to do so.
Sidesteps having to change these for #115.
As part of https://github.com/exercism/x-common/issues/279

### DIFF
--- a/config.json
+++ b/config.json
@@ -29,7 +29,6 @@
     "triangle",
     "scrabble-score",
     "roman-numerals",
-    "binary",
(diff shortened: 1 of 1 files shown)


petertseng:
We're not the only track, https://github.com/exercism/xfsharp/issues/169 did it.

ErikSchierboom:
@petertseng As far as I can tell, everything still works fine.

PR #242, opened 2016-07-31
anagram: Update test cases to match json file.

rbasso:
Related to exercism/x-common#308 from @petertseng.


petertseng:
yup, :+1:

PR #244, opened 2016-08-01
raindrops: Rewrite tests to use hspec with fail-fast.

rbasso:
- Rewrite test to use hspec
- Remove test not matching x-common.

Related to #211.

### DIFF
--- a/exercises/raindrops/package.yaml
+++ b/exercises/raindrops/package.yaml
@@ -16,4 +16,4 @@ tests:
     source-dirs: test
     dependencies:
       - raindrops
-      - HUnit
+      - hspec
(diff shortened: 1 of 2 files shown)


petertseng:
:+1: as-is, but while we're thinking about it, how about a potential question:

For all cases, it is true that `description` is just `show number`. Given this consideration, is `description` really necessary? Then again, I imagine these were generated straight from the JSON file, so this question may have been better for the JSON file in x-common.

So :+1: here.

rbasso:
I agree! It's ugly to have the descriptions equal to the numbers. Considering that they don't add any useful information, It would be nice to remove them in `x-common`, but I can also remove it now from `xhaskell`. Should I do it?

petertseng:
Ah, right. :+1: then, no question.

Optional: This would be a candidate to use tuples since the two fields should be unambiguous, as in [prime-factors the other PR](https://github.com/exercism/xhaskell/pull/243), but as I've said I do not object to a record even if there is an opportunity to use a tuple.

rbasso:
That was fast, @petertseng!

I'm followed your advice and changed to tuples. :+1:

petertseng:
:+1:

PR #245, opened 2016-08-01
allergies: Rewrite tests to use hspec with fail-fast.

rbasso:
- Rewrite tests to use hspec.
- Fixed to allow allergens to be returned in any order.
- Fixed to supress warning on polymorphic solutions.

Related to #211.


petertseng:
:+1:

PR #246, opened 2016-08-01
atbash-cipher: Rewrite tests to use hspec with fail-fast.

rbasso:
- Add `decode` function to the tests.
- Add `decode` to the stub solution.
- Rewrite the example solution to pass the new tests.

Related to #211.

### DIFF
--- a/exercises/atbash-cipher/test/Tests.hs
+++ b/exercises/atbash-cipher/test/Tests.hs
@@ -1,26 +1,84 @@
-import Test.HUnit ((@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
-import Atbash (encode)
+{-# LANGUAGE RecordWildCards #-}
 
-exitProperly :: IO Counts -> IO ()
-exitProperly m = do
-  counts <- m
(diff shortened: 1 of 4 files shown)


exercises/atbash-cipher/test/Tests.hs
```diff
@@ -1,26 +1,84 @@
...
+           }
+    , Case { description = "encode ignores non ascii"
+           , phrase      = "non ascii éignored"
+           , expected    = "mlmzh xrrrt mlivw"
+           }
+    ]
+
+decodeCases :: [Case]
+decodeCases =
+    [ Case { description = "decode exercim"
```
petertseng: is it misspelled in the json file too?

petertseng: indeed, of course it is. I'll fix it right up.

petertseng:
:+1: , you can probably fix the exercim thing as well. I've got the JSON file covered https://github.com/exercism/x-common/pull/316

rbasso:
Fixed to match #316.

exercises/atbash-cipher/test/Tests.hs
rbasso: Nice catch! :smile:

petertseng:
Right then, :+1:

PR #248, opened 2016-08-02
bank-account: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.

### DIFF
--- a/exercises/bank-account/test/Tests.hs
+++ b/exercises/bank-account/test/Tests.hs
@@ -1,9 +1,15 @@
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
-import BankAccount ( BankAccount, openAccount, closeAccount
-                   , getBalance, incrementBalance )
-import Control.Concurrent
-import Control.Monad (void, replicateM)
+import Control.Concurrent (forkIO, newEmptyMVar, putMVar, takeMVar)
+import Control.Monad      (replicateM)
+import Data.Foldable      (for_)
+import Test.Hspec         (Spec, describe, it, shouldReturn)
+import Test.Hspec.Runner  (configFastFail, defaultConfig, hspecWith)
+
+import BankAccount
+  ( closeAccount
+  , incrementBalance
+  , getBalance
+  , openAccount
+  )
 
 {-
 The BankAccount module should support four calls:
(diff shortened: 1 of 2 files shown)


exercises/bank-account/test/Tests.hs
```diff
@@ -24,50 +29,40 @@ updateBalance account amount
...
-checkReturn :: (Show a, Eq a) => a -> IO a -> IO ()
-checkReturn expect test = do
-  v <- test
-  expect @=? v
+    it "incrementing balance from other processes then checking it from test process" $ do
+        account  <-  openAccount
+        replicateM_ 20 $ do
+            v <- newEmptyMVar
+            _ <- forkIO $ incrementBalance account 1 >> putMVar v ()
+            takeMVar v
```
petertseng: I'm a bit unsure on this, but can it be the case that the previous code would potentially perform `incrementBalance` in parallel, but this code can never do that (since the `takeMVar` is in the same `replicateM_` as the `forkIO`?)

petertseng:
I realized the tests (both new and old) never test the result of `incrementBalance` against an account with a nonzero balance, but I suppose that change doesn't necessarily have to be made here,

exercises/bank-account/test/Tests.hs
rbasso: Thanks, @petertseng! I think I accidentally changed the semantics here. I will rewrite it to keep the old behavior. :smile:

rbasso:
The `replicateM 20` has the effect of incrementing non-zero balance accounts, so I think this is somehow covered. But I'll write a more explicit test. :smile: 

> I realized the tests (both new and old) never test the result of incrementBalance against an account with a nonzero balance, but I suppose that change doesn't necessarily have to be made here,

exercises/bank-account/test/Tests.hs
rbasso: Now I think it's fixed and the old semantics is restored.
What I understand is that this is not really parallel and just test updating from distinct threads. It spans 20 threads and then collects the `MVars`. 19 of them will block immediately and will be sequentially unblocked as `takeMVar`s are evaluated. Right?

rbasso:
I changed the the *"incrementing and checking balance"* test to test this case.

>  realized the tests (both new and old) never test the result of incrementBalance against an account with a nonzero balance.

petertseng:
:+1:

PR #249, opened 2016-08-02
crypto-square: Rewrite tests to use hspec with fail-fast.

rbasso:
- Rewrite tests to use hspec.
- Remove old tests.
- Add tests to match `x-common`.
- Remove function `squareSize` from the test suite.
- Remove function `squareSize` from the stub solution.
- Remove function `squareSize` from example's export list.

Related to #211.

### DIFF
--- a/exercises/crypto-square/test/Tests.hs
+++ b/exercises/crypto-square/test/Tests.hs
@@ -1,60 +1,93 @@
-import Test.HUnit ((@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
-import CryptoSquare (normalizePlaintext,
-                     squareSize,
-                     plaintextSegments,
-                     ciphertext,
-                     normalizeCiphertext)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
 
-exitProperly :: IO Counts -> IO ()
-exitProperly m = do
-  counts <- m
-  exitWith $ if failures counts /= 0 || errors counts /= 0 then ExitFailure 1 else ExitSuccess
+import CryptoSquare
+  ( ciphertext
+  , normalizeCiphertext 
+  , normalizePlaintext
+  , plaintextSegments
+  )
 
 main :: IO ()
-main = exitProperly $ runTestTT $ TestList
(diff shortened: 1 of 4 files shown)


exercises/crypto-square/test/Tests.hs
```diff
@@ -1,60 +1,93 @@
...
+
+      it "empty plaintext results in an empty ciphertext" $
+          normalizeCiphertext ""
+          `shouldBe`          ""
+
+      it "9 character plaintext results in 3 chunks of 3 characters" $
+          normalizeCiphertext "This is fun!"
+          `shouldBe`          "tsf hiu isn"
+
+    {- In this track the encoded text chunks are not padded with spaces.
```
petertseng: do you think they should be?

I realize it would probably make the example implementation a bit more complex

To make it clear, I don't care enough to insist on one way or the other

petertseng: Should we instead keep the old test that expects `imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn sseoau`?

rbasso: I think that this exercise is not very idiomatic and deserves a major rewrite, with better functions name and, maybe, using lists of strings. But for now, I am just trying to document the differences between the track and x-common.

I'm against that last test. The multiple spaces between the words make sense only if we look at the string as a matrix - and this is probably the reason why there is space padding in the exercise - but in Haskell we would never represent a matrix as a single string.

What do you think? Should we enforce non-padding (old behavior), padding (json behavior) or accept both?

petertseng: Given that you have plans pending, I'm fine with leaving this commented as it already is. The student can uncomment it if truly desired, and there is a test for the perfect square case above. As long as there is a note in either #194 or a new issue describing the change planned!

rbasso: Issue opened! Thanks again, @petertseng!

I know it is probably boring to review so many similar Hspec PRs, but we almost there. :smile:  


As long as there is a note in either #194 or a new issue describing the change planned!

petertseng: > I know it is probably boring to review so many similar Hspec PRs, but we almost there.

I think to be commended more is your persistence in doing all of them, actually!

rbasso: I had a lot of fun rewriting the first 10 exercises... :grin:

PR #251, opened 2016-08-02
robot-simulator: Rewrite tests to use hspec with fail-fast.

rbasso:
- Remove old tests.
- Add tests to match `x-common`.
- Remove import of `Robot` in test suite.
- Remove export of `Robot` in example solution.
- Remove export of `Robot` in stub solution.
- Remove instructions to derive `(Eq, Ord)` in stub.

Related to #211.

This exercise is too distant from what is defined in `x-common`, so it will probably be good to rewrite it. I tried to leave comments about the deviations to make the task easier.

I also simplified the interface so that it's not necessary to export the `Robot` data type. This is recommended for better modularization and allows more creative solutions.

### DIFF
--- a/exercises/robot-simulator/test/Tests.hs
+++ b/exercises/robot-simulator/test/Tests.hs
@@ -1,48 +1,105 @@
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
-import Robot (Bearing(..), Robot, mkRobot,
-              coordinates, simulate,
-              bearing, turnRight, turnLeft)
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
 
-exitProperly :: IO Counts -> IO ()
-exitProperly m = do
-  counts <- m
-  exitWith $ if failures counts /= 0 || errors counts /= 0 then ExitFailure 1 else ExitSuccess
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
 
-testCase :: String -> Assertion -> Test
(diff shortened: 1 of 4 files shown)


exercises/robot-simulator/test/Tests.hs
```diff
@@ -1,48 +1,105 @@
...
-    West  @=? turnLeft North
-    South @=? turnLeft West
-    East  @=? turnLeft South
-    North @=? turnLeft East
-    East  @=? turnRight North
-    South @=? turnRight East
-    West  @=? turnRight South
-    North @=? turnRight West
-  , testCase "robbie" $ do
-    let robbie :: Robot
```
petertseng: I wonder why we gave it a name...

petertseng: guess it's always been that way (assume we ported from clojure) https://github.com/exercism/exercism.io/commit/33216cfbcd78086400a7f03167ae9dba497afded

exercises/robot-simulator/test/Tests.hs
```diff
@@ -1,48 +1,105 @@
...
+
+    describe "simulate" $ do
+
+    -- The function described by the reference file as
+    -- `instructions` is called `simulate` in this track.
+
+      let simulation pos dir = simulate (mkRobot dir pos)
+
+      it "instructions to move west and north" $ do
+        let robot = simulation (0,0) North "LAAARALA"
```
petertseng: do you mean to add a space after the comma?

petertseng:
:+1: if you add the missing space

exercises/robot-simulator/test/Tests.hs
rbasso: You don't miss anything :smile:
I forgot that one. Thanks!

PR #253, opened 2016-08-03
triangle: Rewrite to match the new json file.

rbasso:
- Complete rewrite the test suite to match `x-common/triangle.json`.
- Fix the example solution to pass the new test suite.
- Fix the stub solution to match changes in `TriangleType`.

Related to #211.

### DIFF
--- a/exercises/triangle/src/Example.hs
+++ b/exercises/triangle/src/Example.hs
@@ -5,12 +5,12 @@ import Data.List (sort)
 data TriangleType = Equilateral
                   | Isosceles
                   | Scalene
-                  | Illogical
+                  | Illegal
                   deriving (Eq, Read, Show)
 
 triangleType :: (Num a, Ord a) => a -> a -> a -> TriangleType
 triangleType a b c
-  | any (<=0) [a, b, c] || s1 + s2 <= s3 = Illogical
+  | any (<=0) [a, b, c] || s1 + s2 <= s3 = Illegal
   | otherwise = [Equilateral, Isosceles, Scalene] !! pred uniqueSides
   where sortedSides  = sort [a, b, c]
         [s1, s2, s3] = sortedSides
--- a/exercises/triangle/src/Triangle.hs
+++ b/exercises/triangle/src/Triangle.hs
@@ -3,7 +3,7 @@ module Triangle (TriangleType(..), triangleType) where
 data TriangleType = Equilateral
                   | Isosceles
                   | Scalene
-                  | Illogical
+                  | Illegal
                   deriving (Eq, Show)
 
 triangleType = undefined
--- a/exercises/triangle/test/Tests.hs
+++ b/exercises/triangle/test/Tests.hs
@@ -4,7 +4,14 @@ import Data.Foldable     (for_)
 import Test.Hspec        (Spec, describe, it, shouldBe)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
 
-import Triangle (TriangleType(..), triangleType)
+import Triangle
+  ( TriangleType ( Equilateral
+                 , Illegal
+                 , Isosceles
+                 , Scalene
+                 )
+  , triangleType
+  )
 
 main :: IO ()
 main = hspecWith defaultConfig {configFastFail = True} specs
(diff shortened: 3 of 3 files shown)


rbasso:
## Coding-style in the test suites

This exercise is simple enough to make this a good opportunity to discuss coding-style in the test suites.

A lot of questions arised when trying to port `triangle` to match [exercism/x-common/triangle.json](https://github.com/exercism/x-common/blob/master/triangle.json).

### Where should we change the list of sides to a triple?

The reference file uses a list of numbers to represent the sides:

```json
{
    "description": "equilateral triangle has all sides equal",
    "sides"      : [2, 2, 2],
    "expected"   : "equilateral"
}
```
 This is unfortunate in Haskell, because the type system cannot enforce the correct number of elements.

We can rewrite it to use tuples...

```haskell
    , Case { description = "larger equilateral triangle"
           , sides       = (10, 10, 10)
           , expected    = Equilateral
           }
```

or, we can pattern-match it later:

```haskell
  where
    [a, b, c] = sides
```

### Should we use a tuple, a custom data type or insert the test data directly in the specification?

Let's take the following tests as an example:

```json
        {
            "description": "scalene triangle with sides in descending order",
            "sides": [5, 4, 2],
            "expected": "scalene"
        },
        {
            "description": "small scalene triangle with floating point values",
            "sides": [0.4, 0.6, 0.3],
            "expected": "scalene"
        },
```

We can write it as a list of tuples...

```haskell
cases :: [(String, [Rational], TriangleType)]
cases = [ ("scalene triangle with sides in descending order"  , [  5,   4,   2], Scalene)
        , ("small scalene triangle with floating point values", [0.4, 0.6, 0.3], Scalene) ]
```

... which is short, but exceeds the 80 columns easily and also forces us to use a data type generic enough to capture the integers and floating-point numbers from distinct test cases.

We can use a custom data type...

```haskell
data Case = Case { description :: String
                 , sides       :: [Rational]
                 , expected    :: TriangleType
                 }

cases :: [Case]
cases = [ Case { description = "scalene triangle with sides in descending order"
               , sides       = [5, 4, 2]
               , expected    = Scalene
               }
        , Case { description = "small scalene triangle with floating point values"
               , sides       = [0.4, 0.6, 0.3]
               , expected    = Scalene
        ]
```

...that solves the width problem, but is verbose and also forces us to fit every test in a single data type.

Another option is to sacrifice the separation between test data and test logic...

```haskell
describe "triangleType" $ do

  it "scalene triangle with sides in descending order" $
    triangleType (5, 4, 2) `shoudBe` Scalene

  it "small scalene triangle with floating point values" $
    triangleType (0.4, 0.6, 0.3) `shouldBe` Scalene
```

... repeating a lot of code, but giving more flexibility.

### How should we write exports and imports?

```haskell
-- Imports

import Triangle (TriangleType(..) , triangleType)

import Triangle
  ( TriangleType (..)
  , triangleType
  )

import Triangle
  ( TriangleType ( Equilateral
                 , Illegal
                 , Isosceles
                 , Scalene
                 )
  , triangleType
  )

-- Exports

module Triangle (TriangleType(..), triangleType) where

module Triangle
  ( TriangleType (..)
  , triangleType
  ) where

module Triangle
  ( TriangleType ( Equilateral
                 , Illegal
                 , Isosceles
                 , Scalene
                 )
  , TriangleType (..)
  , triangleType
  ) where
```

Explicit importing constructors is verbose, but keep us safe name conflicts and the compiler can also check if we use all the imported constructors, helping us to identify some errors.

I don't have an opinion about explicit exporting constructors in the stub solutions...

### Should we keep compatibility with older compilers and libraries?

In this test suite we need to add the following lines to keep compatibility with compilers and libraries versions that our users should not be using:

```haskell
-- base >= 4.8 re-exports Control.Applicative.(<$>).
import Control.Applicative -- This is only need for <$>,  if GHC <  7.10.
import Prelude             -- This trick avoids a warning if GHC >= 7.10.
```

It gets worse in some other exercises, like `gigasencond`:

```haskell
{-# LANGUAGE CPP #-}

...

#if MIN_VERSION_time(1,5,0)

import Data.Time.Format
  ( ParseTime
  , TimeLocale
  , defaultTimeLocale
  , iso8601DateFormat
  , parseTimeOrError
  )

readTime :: ParseTime t => TimeLocale -> String -> String -> t
readTime = parseTimeOrError True

#else

import Data.Time.Format (readTime)
import System.Locale    (defaultTimeLocale, iso8601DateFormat)

#endif
```

Considering that we demand the usage of Stack in the Haskell track, maybe we are just making the code ugly without benefit.

petertseng:
> Where should we change the list of sides to a triple?

If someone asked me to make the choice, I would have put them in the `cases`, such as

```haskell
    , Case { description = "larger equilateral triangle"
           , sides       = (10, 10, 10)
           , expected    = Equilateral
           }
```

> Should we use a tuple, a custom data type or insert the test data directly in the specification?
> We can write it as a list of tuples which is short, but exceeds the 80 columns easily

I don't suppose there's anything wrong with multi-line tuples?

> and also forces us to use a data type generic enough to capture the integers and floating-point numbers from distinct test cases.

This problem is true whether the tests use tuples or records though, so the only way around it is direct insertion.

> Another option is to sacrifice the separation between test data and test logic

I didn't favor this, because of the repetition. Would only do it if the type requirement were too cumbersome.

Can also skip the (0.4, 0.6, 0.3) test if we don't find it super useful.

> Explicit importing constructors is verbose, but keeps us safe from name conflicts and the compiler can also check if we use all the imported constructors, helping us to identify some errors.

I tend to be as explicit as possible in imports.

> I don't have an opinion about explicit exporting constructors in the stub solutions...

Yeah I can't say I'm all that concerned either.

> Should we keep compatibility with older compilers and libraries?
> Considering that we demand the usage of Stack in the Haskell track, maybe we are just making the code ugly without benefit.

I suppose we only have an obligation to keep compatibility with everything in `.travis.yml`. This may also lead to the question of "can we remove old resolvers now such as lts-2.x?" Possibly a separate issue, and I personally don't care a lot since I'm not on an old version. Would be good if someone on an old version spoke up.

petertseng:
I'm :+1: on what's in this PR as-is though

rbasso:
I rewrote it to use multi-line tuples as you suggested, @petertseng . I think it isn't as readable as the solution using records, but it is certainly less repetitive.

> I don't suppose there's anything wrong with multi-line tuples?

petertseng:
:+1: 

If there's any discussion you wanted to keep open so that it doesn't get forgotten (maybe the one about compatibility?)  probably good to open an issue so it does not get forgotten

rbasso:
Done. #264.

PR #254, opened 2016-08-03
binary-search-tree: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.

### DIFF
--- a/exercises/binary-search-tree/test/Tests.hs
+++ b/exercises/binary-search-tree/test/Tests.hs
@@ -1,65 +1,78 @@
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
-import BST (bstLeft, bstRight, bstValue, empty, singleton, insert, fromList, toList)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
 
-exitProperly :: IO Counts -> IO ()
-exitProperly m = do
-  counts <- m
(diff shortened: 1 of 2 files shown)


exercises/binary-search-tree/test/Tests.hs
```diff
@@ -1,65 +1,78 @@
...
+      let t = insert 2 (singleton int4)
+      bstValue t `shouldBe` Just 4
+      (bstLeft t >>= bstValue) `shouldBe` Just 2
+
+    it "inserting same" $ do
+      let t = insert 4 (singleton int4)
+      bstValue t `shouldBe` Just 4
+      (bstLeft t >>= bstValue) `shouldBe` Just 4
+
+    it "inserting right" $ do
```
petertseng: This is obviously not your fault (you just took the names that were already there), but it makes me wonder why one of the test names refers to a value comparison ("less") and the other refers to direction ("right") rather than them being consistent...

petertseng:
:+1:

exercises/binary-search-tree/test/Tests.hs
rbasso: I agree!

This test suite also have other problems. Looking at the **"insert same"** test, it seems to expect a left biased insertion. Was this supposed to be tested?

I'm trying to fix the easiest problems in the test suites while porting them to *hspec*, but in this case I think it deserves an additional PR after we merge this one.

petertseng: >  it seems to expect a left biased insertion. Was this supposed to be tested?

Well, the README https://github.com/exercism/x-common/blob/master/binary-search-tree.md at least says something:

> ll data in the left subtree is less than or equal to the current node's data

It doesn't necessarily mean every track will have to test it, but it's there

PR #255, opened 2016-08-03
difference-of-squares: Rewrite tests to use hspec with fail-fast.

rbasso:
- Update tests to match `x-common`
- Rewrite tests for an `Integral` implementation.
- Add `Integral` type signatures to stub solution.

Related to #211.


petertseng:
:+1:

PR #256, opened 2016-08-03
hexadecimal: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.


petertseng:
I'll just say :+1:  and be done with it because deprecated

petertseng:
I'll just say :+1:  and be done with it because deprecated

(but I did look to make sure everything was reasonable, don't worry)

rbasso:
Thanks again, @petertseng! :smile: 

> (but I did look to make sure everything was reasonable, don't worry)

Issue #193, opened 2016-07-10
Making the Haskell track more gradual


rbasso:
###### Making the Haskell track more gradual
Anyone would like to take the humongous task of solving at least part of this?

1. Check what knowledge is needed to solve each of the exercises.
2. Reorder the exercises creating the most gradual progression possible.
3. Change and move some exercises to make it even more gradual.

If we did this at least for the first 20 exercises it would be nice. If we did this for the entire track it would be fantastic.

rbasso:
Comment from a user about `robot-name`: 

> This exercise was brutal! :)
>
> I found the jump in difficulty from the last exercise to this one to be quite severe!

petertseng:
I went through and tried to write down a few notes for each exercise. Each line has the exercise name, a space, then a list of notes. I could not be exhaustive, and for some exercise I was too lazy to figure out what to write so I only wrote "?"

Stars mean I think the exercise is quite involved.

I don't know how useful this list will be. It is likely it will be need to be improved and completed. I would say big things to pay attention to are the problem that use the IO monad, and stuff I marked with stars.

```
accumulate reimplement library, recursion, test enforces non-strict
allergies bitwise
all-your-base Maybe, math
anagram filter, String manipulate
atbash-cipher String manipulate
bank-account IO monad
beer-song repetitive String generation
binary we should deprecate soon, math
binary-search-tree make a data structure, recursion
bob String manipulate
clock math
connect searching, **
crypto-square String manipulate
custom-set make a data structure
difference-of-squares math
etl combine library functions
food-chain repetitive String generation
forth parser, **
gigasecond math, time
go-counting partition grid, **
grade-school make a data structure (kinda)
grains math, Maybe
hamming simultaneous iteration, Maybe
hexadecimal we should deprecate soon, math
house repetitive String generation
kindergarten-garden ?
largest-series-product ?
leap math
lens-person lens, if you just use the library maybe it's easy, maybe interesting learning about the types if you implement yourself, *
linked-list make a data structure, IO monad
list-ops make a data structure
luhn ?
matrix ?
meetup date manipulation
minesweeper count grid, *
nth-prime math, filter
nucleotide-count Either
ocr-numbers ?
octal we should deprecate soon, math
palindrome-products ?
parallel-letter-frequency parallel
pascals-triangle ?
phone-number Maybe
pig-latin String manipulate
pov searching, *
prime-factors math
pythagorean-triplet math
queen-attack ?
raindrops math
rna-transcription Maybe
[...]

PR #257, opened 2016-08-03
largest-series-product: Rewrite tests to use hspec with fail-fast.

rbasso:
- Rewrite example solution to pass the tests.

Related to #211.


rbasso:
It's important to note that, because this PR changes the tests to match the newer and more detailed cases from `x-common`, this exercise will be a little harder.

petertseng:
:+1: , though did you think the exercise needed to be moved around because of the increased difficulty? Though I suppose that's a question that can be deferred until #193 is tackled.

rbasso:
It's not so much more difficult, so we can re-position it later. :smile:

PR #258, opened 2016-08-04
luhn: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.


petertseng:
:+1: since it seems to just convert all existing. Maybe one day we'll ask "Are the track-specific tests giving us anything extra that the common tests don't? Should any of them get added to the common tests?" but I don't know enough about this problem to have an opinion on that (I haven't done it in any language)

PR #262, opened 2016-08-07
house: Rewrite tests to use hspec with fail-fast.

rbasso:
- Remove use of `unlines`.

Considering that this exercise is about refactoring, the test suite should contain the most crude representation of the song, so we avoid `unlines` to let the user figure that out.

Related to #211.


petertseng:
:+1: seems easy enough since this is the only test...

PR #266, opened 2016-08-09
ocr-numbers: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.


petertseng:
Seems right :+1: 

I guess if we wanted to check for invalid sizes we'd have to return `Maybe`, so we may leave it for later

PR #270, opened 2016-08-09
food-chain: Rewrite tests to use hspec with fail-fast.

rbasso:
- Remove use of `unlines`.

Related to #211.


petertseng:
👍

PR #277, opened 2016-08-23
queen-attack: switch row/column in comment

### LINKED ISSUE #276: queen-attack row and column mismatch
The comment in Queens.hs states that

-- Positions are specified as two-tuples.
-- The first element is is the column (file in Chess terms).
-- The second element is the row (rank in Chess terms).
-- (0, 0) is the top left of the board, and (7, 7) is the bottom right.

This is the reverse of the ordering of tuples used in the test file, assuming standard depiction of a chess board with white at bottom.  Also, the README.md does not specify the ordering of rank and file.

petertseng:
In #207 the comment was added, but the comment was incorrect compared to
what the test cases were testing. This commit repairs the comment to
match the tests.

All four corners of the board are specified to make this abundantly
clear.

Closes #276

### DIFF
--- a/exercises/queen-attack/src/Queens.hs
+++ b/exercises/queen-attack/src/Queens.hs
@@ -1,9 +1,10 @@
 module Queens (boardString, canAttack) where
 
 -- Positions are specified as two-tuples.
--- The first element is is the column (file in Chess terms).
--- The second element is the row (rank in Chess terms).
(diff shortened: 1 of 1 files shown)


rbasso:
Sorry for the delay, but I'm travelling without a computer and it sucks to to use a cell phone. :grin:

Considering that Queen.hs was not changed in hspec-fail-fast, I think there will be no conflict.

Thanks! :+1:

Issue #263, opened 2016-08-07
[all-your-base, leap, hamming] tests are swapped


NobbZ:
By stepwise doing `all-your-base`, I encountered that a `Just [1]` was expected but a `Nothing` got, while my code was only able to return `Just`s at that stage:

```hs
rebase :: Integral a => a -> a -> [a] -> Maybe [Integer]
rebase inputBase outputBase input =
  let x = fromBase (toInteger inputBase) input
  in Just [x]

fromBase :: Integral a => Integer -> [a] -> Integer
fromBase b = go 0
  where
    go !n []     = n
    go !n (d:ds) = go (n * (toInteger b) + (toInteger d)) ds
```

And I got following test-output (only one out of many):

```txt
### Failure in: 20:both bases are negative
test\Tests.hs:25
expected: Just [1]
 but got: Nothing
```

After changing all occurences of `(~=?)` with `(~?=)` in the `test/Tests.hs` I got a more appropriate output in the testrun:

```txt
### Failure in: 20:both bases are negative
test\Tests.hs:25
expected: Nothing
 but got: Just [1]
```

rbasso:
Thanks for reporting that, @NobbZ.

I guess I confused the operators when writing it. :grin: 

This behavior is already fixed in branch `hspec-fail-fast`...

```haskell
    test Case{..} = it description assertion
      where
        assertion  = expression `shouldBe` outputDigits
        expression = rebase inputBase outputBase inputDigits
```

...but we only had plans to merge it to `master` after migrating the remaining 30 exercises to `Hspec`.

Should we fix that now in `master` or wait for the merge, probably in two weeks?

NobbZ:
I've fixed it locally, so I'm fine with whatever you do :) As far as I did
understand the procedure, @kytrinix needs to sync the tests by hand into
the main repository. So I think we can just let it as is until HSpec is
there.

rbasso <notifications@github.com> schrieb am So., 7. Aug. 2016 um 22:28 Uhr:

kytrinyx:
> As far as I did
> understand the procedure, @kytrinyx needs to sync the tests by hand into
> the main repository.

Yeah basically. We have a separate API that delivers the problem data. It's not a lot of code, but it includes all of the language-specific tracks as submodules, so I have to update the submodules, commit, and deploy, which is still a manual process. I'm sure this could be automated to some degree, but I'm wary of missing something stupid and breaking the site while I'm asleep or traveling.

PR #278, opened 2016-09-02
Fix the order of 'expected' and 'actual' values

hakantunc:


rbasso:
Thank you for the PR and sorry for the delay. I was traveling without a computer, and arrived home a few hours ago.

This is related to #263, that identified a few exercises where I stupidly confused `(~=?)` and `(~?=)`.
The problem was accidentally fixed in branch `hspec-fail-fast`, when the exercises where rewritten to use `hspec`, and everything would be solved with #211, after merging it back to `master`.

Unfortunately, the migration to `hspec` is being much slower than expected, and it will probably not be finished soon. My fault!

We have some options here:

1. Merge the PR, creating a conflict when merging `hspec-fail-fail` to `master`.
2. Bring the new version from `hspec-fail-fast` to `master`, avoiding the conflict.
3. Merge `hspec-fail-fast` to `master` right now and continue the migration to `hspec` directly in `master`, avoiding similar problems in the future.

I don't know much about *git* and *GitHub*, so I'm not very confident about what is the best decision here, but I'm inclined to follow the **third option**.

What do you think?

Anyone else has an opinion on this?

petertseng:
Can you clarify - does applying the second option mean that on master only all-your-base would be using hspec?

What disadvantages do you see with taking the third option? That students will get hspec with some exercises but hunit with others? And you wish to avoid having them see that inconsistency?

If the disadvantages are acceptable, it seems good to let students reap the benefits of the work sooner, and take the third option.

If the disadvantages are not acceptable, I would suggest to merge this PR then rebase `hspec-fail-fast` on master to resolve the conflict sooner rather than later.

rbasso:
> Can you clarify - does applying the second option mean that on master only all-your-base would be using hspec?

Yes. Sounds horrible, right?

> What disadvantages do you see with taking the third option? That students will get hspec with some exercises but hunit with others? And you wish to avoid having them see that inconsistency?
> If the disadvantages are acceptable, it seems good to let students reap the benefits of the work sooner, and take the third option.

I regret creating the `hspec-fail-fast` branch, it was not really needed. My original plan was to finish everything in less than a month, so that would not be a huge problem, and we would be able to migrate all exercises at the same time. It was a bad idea, sorry!

So, maybe it's already time to:

- [ ] Merge `hspec-fail-fast` in `master`.
- [ ] Close this PR and #263.
- [ ] Update #211 to reflect the changes.
- [ ] Delete the branch `hspec-fail-fast`.

Sorry for the trouble, @hakantunc. :bow:

petertseng:
OK, I'm all for merging `hspec-fail-fast` now.

Because no matter when we merge it, there will likely be students partway through the track who were seeing hunit and now suddenly they see hspec. So it's not like waiting avoids that. The only thing that might happen is: a student was partway through, used to see hunit, and then suddenly sees hspec... then finishes enough problems and sees hunit again.

If a student gets bothered by that, maybe student can help out with the conversion? =D

I haven't done any myself since I've had other matters to deal with myself as well.

PR #280, opened 2016-09-03
gigasecond: Deprecate exercise.

### LINKED ISSUE #230: gigasecond: Completely rewrite the exercise or deprecate it.
The way it is now, `gigasecond` has some problems:

- It is too easy and boring.
- All answers are almost the same: `fromDay = addUTCTime 1e9`
- Does not match `exercism/x-common/gigasecond.json`.

Seems that, at least in other tracks, it can be interesting: https://github.com/exercism/xhaskell/issues/192#issuecomment-234806673

> In the Go track gigasecond is a fantastic exercise. People tackle it in so many different ways, and there are myriad opportunities to talk about idioms in the language and certain language features.
>
> I think that if it can be turned into a more interesting exercise, then that's worth doing. If not, just deprecate it.

**Edit:** This exercise was changed in branch `hspec-fail-fast`. Please use that branch for any PR related to it or wait until it's merged in `master`.

rbasso:
Closes #230.

### DIFF
--- a/config.json
+++ b/config.json
@@ -8,7 +8,6 @@
     "accumulate",
     "sublist",
     "grains",
-    "gigasecond",
     "bob",
     "sum-of-multiples",
     "strain",
@@ -74,6 +73,7 @@
   ],
   "deprecated": [
     "binary",
+    "gigasecond",
     "hexadecimal",
     "octal",
     "trinary"


petertseng:
Hi I could rubber-stamp :+1:  this. Note! I'm taking your word for it that everyone's gigasecond submission is the same since I haven't done it and can't see the submissions. Thus, take anything I say with a grain of salt. And in case anyone sees this comment and asks, no, I don't really have time to look at gigasecond submissions so don't magically give me access to that.

Anyway, this is valid JSON so if nobody else has anything to say, what you say goes =D

PR #282, opened 2016-09-03
travis: Drop compatibility with old Stack snapshots.

### LINKED ISSUE #264: discussion: Dropping compatibility with older snapshots
### Should we keep compatibility with older compilers and libraries?

In some exercises, we are forced to write ugly compatibility code, especially to deal with changes in `base`:

```haskell
-- base >= 4.8 re-exports Control.Applicative.(<$>).
import Control.Applicative -- This is only need for <$>,  if GHC <  7.10.
import Prelude             -- This trick avoids a warning if GHC >= 7.10.
```

It gets worse when we have to use `time`:

```haskell
{-# LANGUAGE CPP #-}

...

#if MIN_VERSION_time(1,5,0)

import Data.Time.Format
  ( ParseTime
  , TimeLocale
  , defaultTimeLocale
  , iso8601DateFormat
  , parseTimeOrError
  )

readTime :: ParseTime t => TimeLocale -> String -> String -> t
readTime = parseTimeOrError True

#else

import Data.Time.Format (readTime)
import System.Locale    (defaultTimeLocale, iso8601DateFormat)

#endif
```

Considering that we demand the usage of Stack in the Haskell track, and that this forces users to run exactly the same compiler and library versions, maybe we could drop compatibility and start writing code using all the new features available in the newest software.

Cons:

- If a user is not using Stack and is also running an older GHC version directly, he/she may be forced to install Stack to be able to run some tests.
- Users will not be exposed to compatibility code.

Pros:

- More beautiful code
- Optimized for happiness.
- Less lines of code.
- Exposes users to new features.
- Saves resources for other Travis-CI users.

Should we do it?

rbasso:
This removes integration tests for unused Stack snapshots in Travis CI. Only the current snapshot and `nightly` will be automatically tested.

Compatibility code in the test suites can now be removed safely.

Closes #264.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,6 @@ cache:
 env:
  - RESOLVER="nightly-2016-07-17" CURRENT="YES" # Equal to each stack.yaml.
  - RESOLVER="nightly"                          # Latest nightly snapshot.
- - RESOLVER="lts-2.22"                         # Last GHC 7.8.x snapshot.
- - RESOLVER="lts-6.7"                          # Last GHC 7.10.x snapshot.
 
 matrix:
   allow_failures:             # The snapshot `nightly` is just an aliases to


petertseng:
Parting is such sweet sorrow.

:+1:

PR #283, opened 2016-09-03
custom-set: Rewrite tests to use hspec with fail-fast.

rbasso:
- Rewrite tests to use hspec.
- Remove explicit testing of `Show`.

Related to #211.

### DIFF
--- a/exercises/custom-set/test/Tests.hs
+++ b/exercises/custom-set/test/Tests.hs
@@ -1,92 +1,204 @@
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
-import Data.List (sort)
-import CustomSet (CustomSet)
-import qualified CustomSet as S
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
+
+import Data.List         (sort)
+import Test.Hspec        (Spec, describe, it, shouldBe)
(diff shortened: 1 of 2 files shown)


exercises/custom-set/test/Tests.hs
```diff
@@ -1,92 +1,204 @@
...
+        it "delete existing element" $
+          delete 2 (fromList [1, 2, 3]) `shouldBe` fromList [1, 3]
+
+      describe "size" $ do
+
+        it "size of an empty set is zero" $
+          size (fromList ([] :: [Integer])) `shouldBe` 0
+
+        it "size is the number of elements" $
+          size (fromList [1, 2, 3, 2]) `shouldBe` 3
```
petertseng: the same as the below test!

exercises/custom-set/test/Tests.hs
```diff
@@ -1,92 +1,204 @@
...
+
+        it "size doesn't count repetitions in `fromList`" $
+          size (fromList [1, 2, 3, 2]) `shouldBe` 3
+
+        it "size doesn't count repetitions from `insert`" $
```
petertseng: I don't think this was a repetition though. I would have said repetition if it were ```size (insert 3 (fromList [1, 3])) `shouldBe` 2```

rbasso: Thank you very much for the review! The `size` tests where really wrong! :smile:

I think it is fixed now. :+1: :question:

petertseng:
:+1:

PR #284, opened 2016-09-05
Minesweeper match json

### LINKED ISSUE #273: Update minesweeper to match x-common
exercism/x-common#330 added a json file with the test cases for `minesweeper`. Now we need to update the exercise to match it.

~~Because the exercise `minesweeper` was changed in `hspec-fail-fast`, we need to update it there or wait until it's merged into master.~~

rbasso:
- Rewrite example solution to pass the new tests.

Closes #273.

### DIFF
--- a/exercises/minesweeper/test/Tests.hs
+++ b/exercises/minesweeper/test/Tests.hs
@@ -12,9 +12,6 @@ specs = describe "minesweeper" $
           describe "annotate" $ for_ cases test
   where
 
-    -- As of 2016-08-09, there was no reference file
-    -- for the test cases in `exercism/x-common`.
-
     test (description, board) = it description assertion
       where
         assertion  = annotate (clearBoard board) `shouldBe` board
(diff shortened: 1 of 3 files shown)


exercises/minesweeper/test/Tests.hs
```diff
@@ -12,41 +12,62 @@ specs = describe "minesweeper" $
...
 
-            , ("surrounded", [ "***"
-                             , "*8*"
-                             , "***" ] )
+            , ("space surrounded by mines", [ "***"
+                                            , "*8*"
+                                            , "***" ] )
 
-            , ("horizontal line", [ "1*2*1" ] )
+           , ("horizontal line", [ "1*2*1" ] )
```
petertseng: I think a space disappeared, should it have? (I don't think it should have)

exercises/minesweeper/test/Tests.hs
```diff
@@ -12,41 +12,62 @@ specs = describe "minesweeper" $
...
+
+           , ("horizontal line, mines at edges", [ "*1 1*" ] )
```
petertseng: I think this one should also have an extra space before the comma

petertseng:
I'm surprised that the example solution needed a change to pass the new test! Let me see what it would have failed on...

petertseng:
Oh, I get it now. the example had X and Y flipped. Amazing that it was not found until now, guess it was very important to have a non-symmetric board!!!

Okay, :+1: from me.

exercises/minesweeper/test/Tests.hs
rbasso: Fixed. Thanks!

petertseng:
Eveyrthing appears to be in order. :+1:

rbasso:
I was really surprised too, @petertseng. But instead of trying to find the bug, I just took the opportunity to rewrite it without `array`. I'm not sure if it's better... :grin:

petertseng:
> guess it was very important to have a non-symmetric board!!!

I don't think this explanation ever satisfied me. If the only condition is to have an asymmetric board, then all the 1xN and Nx1 boards should have been sufficient to catch the bug.

Turns out, the condition is that the boards must have both dimensions larger than 1 AND be asymmetric. Code coming in a moment (a PR will link to this PR)

petertseng:
To see why, see that `listArray (1, 1) (X, Y)` assigns indices to
elements in the order `(1, 1), (1, 2) ... (1, Y), (2, 1) ...`
However, we `concat` the `[String]` board, which means that successive
elements should increment the X coordinate, not the Y coordinate.
By switching X and Y in `listArray` (and all other uses), we get a
correct assignment.

Note that it's necessary for both dimensions to be larger than 1. If
either dimension is 1, only one of the dimensions gets increased in
`listArray`, therefore there is no danger of an incorrect assignment.

PR #285, opened 2016-09-06
nth-prime: Rewrite tests to use hspec with fail-fast.

rbasso:
- Rewrite tests to use hspec.
- Update tests to match exercism/x-common#332.
- Change return type to Maybe.

Related to #211.


petertseng:
:+1:

PR #286, opened 2016-09-07
palindrome-products: Rewrite tests to use hspec with fail-fast.

rbasso:
- Add `HINTS.md`.
- Move old hints to `HINTS.md`.
- Update the stub solution.

We took the opportunity to use this PR as a first experiment with `HINTS.md` in the Haskell track.

### DIFF
--- a/exercises/palindrome-products/test/Tests.hs
+++ b/exercises/palindrome-products/test/Tests.hs
@@ -1,73 +1,36 @@
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
+
+import Data.Foldable     (for_)
+import Data.List         (nub, sort)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
 import Palindromes (largestPalindrome, smallestPalindrome)
-import qualified Data.Set as S
-
--- largestPalindrome, smallestPalindrome :: Integral a => a -> a -> (a,[(a, a)])
--- largestPalindrome minFactor maxFactor = (value, [(factor1, factor2)])
-
--- It's ok to return duplicates in the factor list, and the order of the factors
--- is irrelevant.
-
--- You should consider using a slightly different algorithm to find small or
--- large palindromes.
-
-exitProperly :: IO Counts -> IO ()
-exitProperly m = do
-  counts <- m
-  exitWith $ if failures counts /= 0 || errors counts /= 0 then ExitFailure 1 else ExitSuccess
-
-testCase :: String -> Assertion -> Test
-testCase label assertion = TestLabel label (TestCase assertion)
 
 main :: IO ()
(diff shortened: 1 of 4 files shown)


petertseng:
Seems good.

Don't let this hold up merging, but a question!

Regarding the experiment to use HINTS.md: What result do you think you will use to decide whether to apply it to other exercises?

exercises/palindrome-products/test/Tests.hs
```diff
@@ -1,73 +1,36 @@
...
+          it "value"   $ value             `shouldBe` lPal
+          it "factors" $ normalize factors `shouldBe` lPalFactors
+
+    -- As of 2016-09-07, there was no reference file
+    -- for the test cases in `exercism/x-common`.
+
+    cases = [ ("palindromes from single digit factors",     1,     9,         1, [(    1,     1)],          9, [(1, 9), (3, 3)])
+            , ("palindromes from double digit factors",    10,    99,       121, [(   11,    11)],       9009, [(   91,    99)])
+            , ("palindromes from triple digit factors",   100,   999,     10201, [(  101,   101)],     906609, [(  913,   993)])
+            , ("palindrome from four digit factors"   ,  1000,  9999,   1002001, [( 1001,  1001)],   99000099, [( 9901,  9999)])
```
petertseng: simgular versus plural, maybe they should all be the same?

rbasso: Nice catch! Fixed! :smile: 

> Regarding the experiment to use HINTS.md: What result do you think you will use to decide whether to apply it to other exercises?

Right now we have hints as comments in the test suites and in the stub solutions.

The comments in the stub solutions proved to be a tragedy, because most of the users just leave the hints in the submitted code.

The comments in the test suite are not that bad, but I think that the users should get feedback primarily running the test suite, not inspecting it.

If the hints file works as expected, I think it would be probably a good idea to do the same for all exercises. But first I would like to check if it is correctly inserted in the `README.md` and if it doesn't have unexpected effects.

What do you think about it?
Any other track is using it?

rbasso:
> Regarding the experiment to use HINTS.md: What result do you think you will use to decide whether to apply it to other exercises?

Right now we have hints as comments in the test suites and in the stub solutions.

The comments in the stub solutions proved to be a tragedy, because most of the users just leave the hints in the submitted code.

The comments in the test suite are not that bad, but I think that the users should get feedback primarily running the test suite, not inspecting it.

If the hints file works as expected, I think it would be probably a good idea to do the same for all exercises. But first I would like to check if it is correctly inserted in the `README.md` and if it doesn't have unexpected effects.

What do you think about it?
Any other track is using it?

petertseng:
I wish I could just search for https://github.com/search/advanced?l=&q=org%3Aexercism++path%3AHINTS.md&ref=advsearch&type=Code&utf8=%E2%9C%93 but I was not able to get this to work. Maybe someone else will have better luck.

Oh well, at least I know that xrust has https://github.com/exercism/xrust/blob/master/exercises/space-age/HINTS.md

NobbZ:
Also I do know about xelixir, which does something similar: https://github.com/exercism/xelixir/tree/master/exercises/zipper/hints

There it is a subfolder in the exercise, containing a `README.md` as well as three files `hint_i.md`.

PR #287, opened 2016-09-08
pascals-triangle: Rewrite tests to use hspec with fail-fast.

rbasso:
- Rename function `triangle` to `rows`
- Change tests do expect `row` to count lines starting at zero.
- Rewrite the stub solution to be more concise and match the new test suite.

~~Considering that there is a ongoing discussion in exercism/x-common#362, I think we should wait it before merging this PR.~~

Related to #211.

### DIFF
--- a/exercises/pascals-triangle/test/Tests.hs
+++ b/exercises/pascals-triangle/test/Tests.hs
@@ -1,36 +1,28 @@
 {-# OPTIONS_GHC -fno-warn-type-defaults #-}
 
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
-import Triangle (row, triangle)
+import Data.Foldable     (for_)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
 
-exitProperly :: IO Counts -> IO ()
-exitProperly m = do
-  counts <- m
-  exitWith $ if failures counts /= 0 || errors counts /= 0 then ExitFailure 1 else ExitSuccess
-
-testCase :: String -> Assertion -> Test
-testCase label assertion = TestLabel label (TestCase assertion)
+import Triangle (rows)
 
 main :: IO ()
-main = exitProperly $ runTestTT $ TestList
-       [ TestList triangleTests ]
+main = hspecWith defaultConfig {configFastFail = True} specs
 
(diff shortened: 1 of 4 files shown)


petertseng:
remind when that discussion concludes, I think. I may forget otherwise. I have no problems with what I see in here so far

rbasso:
Tests rewritten to match `x-common`.

The only relevant divergence is that, for negative numbers, I decided to return an empty list, mimicking the behavior of `take`. We already have exercises using `Maybe`, so I though it would be a good idea, and it also allows simpler implementations.

exercises/pascals-triangle/test/Tests.hs
```diff
@@ -1,36 +1,28 @@
...
-    , 92378, 75582, 50388, 27132, 11628, 3876, 969, 171, 19, 1] @=? row 20
-  , testCase "201st row maximum" $
-    (product [101..200] `div` product [1..100] :: Integer) @=? row 201 !! 100
-  ]
+    rowsCases = [ ("no rows"      , 0, [                                    ])
+                , ("single row"   , 1, [[1]                                 ])
+                , ("two rows"     , 2, [[1], [1, 1]                         ])
+                , ("three rows"   , 3, [[1], [1, 1], [1, 2, 1]              ])
+                , ("four rows"    , 4, [[1], [1, 1], [1, 2, 1], [1, 3, 3, 1]])
+                , ("negative rows", 0, [                                    ]) ]
```
petertseng: you didn't actually put a negative number here

petertseng:
I'm debating whether to say "maybe the test should explain why negative rows is empty because it's like `take`", but maybe that will hint too strongly to people about using `take`! So maybe no explanation needed? What say you?

exercises/pascals-triangle/test/Tests.hs
rbasso: My stupid mistake. Have no idea why I changed that.

rbasso: Fixed! Thanks for catching that, @petertseng! :smile:

petertseng:
Yup. 👍

rbasso:
> ... maybe the test should explain why negative rows is empty because it's like `take` ...

Both `take` and `drop` silently ignore invalid inputs. This can be seem as a feature or as a bad design, but I don't have a strong opinion on this.

Considering that we have this behavior in `Prelude`, I don't think we have to explain anything, unless there is a consensus that `take` and `drop` have an undesirable specification.

PR #289, opened 2016-09-11
pig-latin: Rewrite tests to use hspec with fail-fast.

rbasso:
- Rewrite test to use `hspec`.
- Update the tests to match `x-common`.
- Patch the example solution to pass the new tests.

### DIFF
--- a/exercises/pig-latin/package.yaml
+++ b/exercises/pig-latin/package.yaml
@@ -16,4 +16,4 @@ tests:
     source-dirs: test
     dependencies:
       - pig-latin
-      - HUnit
+      - hspec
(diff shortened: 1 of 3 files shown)


petertseng:
Plan to leave the ugly hack in indefinitely?

In my point of view it is more valuable to have the tests match than the solution be elegant, so I say :+1: for this since I don't feel like taking the time to come up with a more elegant solution, depends on what you want to do with it long term.

rbasso:
Thanks for the review! :smile: 

> Plan to leave the ugly hack in indefinitely?

Until someone writes a more elegant solution :grin:.

The examples are really low on my priority-list.

> In my point of view it is more valuable to have the tests match than the solution be elegant...

I completely agree!

PR #290, opened 2016-09-12
pythagorean-triplets: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.

### DIFF
--- a/exercises/pythagorean-triplet/package.yaml
+++ b/exercises/pythagorean-triplet/package.yaml
@@ -16,4 +16,4 @@ tests:
     source-dirs: test
     dependencies:
       - pythagorean-triplet
-      - HUnit
+      - hspec
--- a/exercises/pythagorean-triplet/test/Tests.hs
+++ b/exercises/pythagorean-triplet/test/Tests.hs
@@ -1,32 +1,44 @@
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
-import Triplet (mkTriplet, isPythagorean, pythagoreanTriplets)
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
 
-exitProperly :: IO Counts -> IO ()
-exitProperly m = do
-  counts <- m
-  exitWith $ if failures counts /= 0 || errors counts /= 0 then ExitFailure 1 else ExitSuccess
+import Data.Foldable     (for_)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
 
-testCase :: String -> Assertion -> Test
-testCase label assertion = TestLabel label (TestCase assertion)
+import Triplet (isPythagorean, mkTriplet, pythagoreanTriplets)
 
 main :: IO ()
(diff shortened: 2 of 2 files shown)


petertseng:
:+1: 

By the way, something I didn't notice before, have you always used `<$>` instead of `map`?

I know that for the `uncurry3 mkTriplet` it saves a pair of parentheses, guess it look better if it's all consistent and all use `<$>` =D

Don't mind me, just musing

rbasso:
I usually favor `<$>` over `map|fmap` when it has both arguments applied. But I don't have a very strong opinion about it. When it's partially applied, I avoid it, because I think it's a little strange as a *section*:

```haskell
h = fmap f . g

vs

h = (f <$>) . g
```

Now that we have dropped compatibility, we don't have to explicitly import `<$>` anymore, so I try to use it a lot.

I'm open to discuss it and even set a coding style to keep consistency in the track.

What is your preference :question:

petertseng:
I don't feel strongly enough to prescribe a style for this repo (and besides, very few lines in this repo are mine, by this time!), but I can sure talk about what I had used in the past

I think if we look at the Haskell code I submitted to Exercism, I always used `fmap` over `<$>`. It was also rare for me to use `$` and I instead used more parentheses. However, I realize this is in contravention of https://wiki.haskell.org/Programming_guidelines#Application_notation . I probably got it from reading https://github.com/chrisdone/haskell-style-guide/blob/8e3b0221ea5862a977b0954d420980e313fc6df5/README.md but note that it has now been [deprecated](https://github.com/chrisdone/haskell-style-guide) in favor of https://github.com/tibbe/haskell-style-guide/blob/master/haskell-style.md which makes no such prescription! I'm also not going to prescriptive about `fmap` and `<$>`, I think I can still read both.

If we *did* set a coding style, I wonder if we can have some tool to run in CI that can help check it automatically. Maybe https://github.com/chrisdone/hindent - I have not tested it so I do not know if it's something we want.

hlint might also be something that's nice to run in CI but by this time we are off topic since hlint is checking a different thing from style (and also last I checked hlint had quite a few things to say about thi repo...)

petertseng:
On that note, it seems hlint likes `<$>` in certain situations (not all, it seems)

```
./hamming/src/Example.hs:4:26: Suggestion: Use <$>
Found:
  fmap (if x /= y then (+ 1) else id) $ distance xs ys
Why not:
  (if x /= y then (+ 1) else id) Control.Applicative.<$>
    distance xs ys
```

rbasso:
Never heard of `hindent` before. I'll try to take a look at it when a get the time.

> hlint might also be something that's nice to run in CI but by this time we are off topic since hlint is checking a different thing from style

Great idea! I would gladly rewrite any test suite that is not *hlint friendly* as soon as we finish #211, which will overwrite a lot of code. If we decide to make everything pass `hlint`, then we can put it in CI.

I know that I should run `hlint` before every *push*, but I admit that I forgot to do it a lot of times. :grin:

kytrinyx:
> I admit that I forgot to do it a lot of times.

Welcome to my world :)

I finally admitted that my human brain would never be up to the task of remembering these things and added a pre-push commit hook. Then I changed computers and forgot to install the hook on my new laptop... it's a never-ending struggle.

PR #291, opened 2016-09-12
saddle-points: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.


petertseng:
:+1:

rbasso:
Thanks again for reviewing all this, @petertseng! :smile:

petertseng:
Well, it is the least I can do since you are doing all the hard work :x

PR #292, opened 2016-09-12
say: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.

### DIFF
--- a/exercises/say/package.yaml
+++ b/exercises/say/package.yaml
@@ -17,4 +17,4 @@ tests:
     source-dirs: test
     dependencies:
       - say
-      - HUnit
+      - hspec
--- a/exercises/say/test/Tests.hs
+++ b/exercises/say/test/Tests.hs
@@ -1,54 +1,44 @@
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
+
+import Data.Foldable     (for_)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
 import Say (inEnglish)
 
-exitProperly :: IO Counts -> IO ()
-exitProperly m = do
-  counts <- m
-  exitWith $ if failures counts /= 0 || errors counts /= 0 then ExitFailure 1 else ExitSuccess
-
-testCase :: String -> Assertion -> Test
-testCase label assertion = TestLabel label (TestCase assertion)
-
 main :: IO ()
-main = exitProperly $ runTestTT $ TestList
-       [ TestList inEnglishTests ]
+main = hspecWith defaultConfig {configFastFail = True} specs
 
-inEnglishTests :: [Test]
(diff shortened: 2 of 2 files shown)


petertseng:
Did not think we would have bounds that low! But it is not for me to question

:+1:

rbasso:
> Did not think we would have bounds that low! But it is not for me to question

You can question anything! :smile:

Considering that there is no json for this exercise, we can change it as we like.

So, what would you like to change :question:

petertseng:
I suppose what I really meant was: this PR is just the conversion from HUnit to HSpec, talking about changes to the tests can come separately! But here is an OK place I suppose, since the lines are getting changed anyway.

Two separate questions for us to answer:

* The upper limit doesn't have to be billions! Shall we also add trillion, quadrillions?
    * But, does that really add a lot to the exercise? Maybe not. If they can get thousand, million, billion, is adding trillion, etc. going to be that interesting? I suspect not.
    * Also https://github.com/exercism/x-common/blob/master/exercises/say/description.md says "it's fine to stop at trillion"
    * So I definitely say no to this.
* The lower limit doesn't have to be zero! Shall we add negative numbers? -1 could be "negative one", etc.
    * Does this add to the exercise? Maybe so. It is a new consideration that almost certainly is dealt with in a different way.
    * However it's not mentioned in https://github.com/exercism/x-common/blob/master/exercises/say/description.md ! So either we add it track-specific or we'd need to change the description in x-common.
    * I could be convinced to say yes to this, but I'm currently not feeling strongly.

rbasso:
You are right, there is no reason to restrict the values.

I completely agree about dropping the upper limit, but I'm still unsure about the lower limit. Do you think that would add something interesting to the exercise?

If we decide to drop the lower limit too, there is no reason to return a `Maybe` anymore. How do you fell about it? Should we drop it too and remove the `Maybe`?

petertseng:
Take my opinion with a grain of salt since I haven't done this exercise myself in any language. I do think negatives would be somewhat interesting (probably more interesting than going to trillions, quadrillions, etc.), but the question of whether it's interesting *enough* to be worth adding? Well, if unsure I'll suggest to try it! But I really don't feel strongly.

It may be the case that Maybe has to be kept though because no matter how far someone goes with trillions, quadrillions, etc. there still has to be a point at which they had to stop. Unless we expect solutions to be able to arbitrarily combine latin prefixes?!

petertseng:
OK. On your recommendation that it would not be more interesting, then I suggest no more tests and merge as-is. If a preponderance of other track maintaners do decide to expand when someone creates say/canonical-data.json then we can consider.

rbasso:
I removed the upper-limit test, but left the test for `-1` returning `Nothing` for now, until we discuss the negative inputs in `x-common`.

petertseng:
Okay, so the idea is that this makes life better for the student - the student can now go to trillions if desired (the tests don't stop you now!) whereas before they could not.

OK

:+1:

rbasso:
If you prefer I can drop the lower limit test too. I just would like to avoid, for now, to expect valid negative outputs.

petertseng:
It's good as-is. Besides, if removed the lower limit test, then a student might ask "Why are all tests `Just`!"

PR #293, opened 2016-09-12
secret-handshake: Rewrite tests to use hspec with fail-fast.

rbasso:
- Move hints to `HINTS.md`

Related to #211.

### DIFF
--- a/exercises/secret-handshake/test/Tests.hs
+++ b/exercises/secret-handshake/test/Tests.hs
@@ -1,52 +1,51 @@
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
 import SecretHandshake (handshake)
 
-exitProperly :: IO Counts -> IO ()
-exitProperly m = do
-  counts <- m
-  exitWith $ if failures counts /= 0 || errors counts /= 0 then ExitFailure 1 else ExitSuccess
-
-testCase :: String -> Assertion -> Test
-testCase label assertion = TestLabel label (TestCase assertion)
-
 main :: IO ()
-main = exitProperly $ runTestTT $ TestList
-       [ TestList handshakeTests ]
+main = hspecWith defaultConfig {configFastFail = True} specs
 
--- The trick to this one is defining the appropriate typeclass or typeclasses
--- to make handshake work with both Int and String. Granted, this is probably
--- not good API design for this exercise, but the technique is useful.
--- You may want to see http://www.haskell.org/haskellwiki/List_instance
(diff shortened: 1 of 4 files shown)


petertseng:
:+1: 

you could see if there's anything in https://github.com/exercism/x-common/pull/159 we want to use, but maybe not

rbasso:
I missed that PR. Thanks! :smile:

I'll take a look at that and maybe wait for it before merging.

exercises/secret-handshake/test/Tests.hs
```diff
@@ -1,52 +1,51 @@
...
+    it "11111 to jump, close your eyes, double blink, and wink" $ do
+      handshake (31 :: Int) `shouldBe` ["jump", "close your eyes", "double blink", "wink"]
+      handshake "11111"     `shouldBe` ["jump", "close your eyes", "double blink", "wink"]
+
+    it "zero" $ do
+      handshake (0 :: Int) `shouldBe` []
+      handshake "0"        `shouldBe` []
+
+    it "gibberish" $
+      handshake "piggies" `shouldBe` []
```
petertseng: Do you think these should be `Nothing` (and the above valid cases `Just`)?

rbasso: Makes perfect sense to me to return `Nothing` on invalid data. :smile:

I'll rewrite this PR. Thanks!

rbasso: I was rewriting this PR to deal with `Maybe` values - which makes sense - but I noticed while rewriting the example solution that everything felt so artificial with this exercise. Even the test suite makes it clear that the API design is far from good:

> The trick to this one is defining the appropriate typeclass or typeclasses
> to make handshake work with both Int and String. Granted, this is probably
> not good API design for this exercise, but the technique is useful.

I'm convinced that this exercise deserves a complete rewrite, so I propose:

1. Merge this PR the way it is, just to keep moving the migration to `hspec`.
2. Wait for exercism/x-common#159 to merge, luckily with a more sane set of tests.
3. Completely rewrite this exercise from scratch.

petertseng: Seems good to merge now.

It may be that that PR will take a while, so maybe we can pre-emptively act on step 3 and not wait for the PR. Depends on how quick we think the PR will get merged, versus how much confusion we think it will cause students. (Of course, someone can adopt the PR and make the requested changes if desired)

rbasso: > It may be that that PR will take a while, so maybe we can pre-emptively act on step 3

I completely agree with  you!

rbasso: > Of course, someone can adopt the PR and make the requested changes if desired

How would that happen? Creating a new PR?

petertseng: >> Of course, someone can adopt the PR and make the requested changes if desired
> 
> How would that happen? Creating a new PR?


Yup.

PR #294, opened 2016-09-12
series: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.

### DIFF
--- a/exercises/series/package.yaml
+++ b/exercises/series/package.yaml
@@ -16,4 +16,4 @@ tests:
     source-dirs: test
     dependencies:
       - series
-      - HUnit
+      - hspec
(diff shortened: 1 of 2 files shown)


petertseng:
:+1:

petertseng:
Do you think it would be instructive and/or useful to test that ```series 0 "" `shouldBe` [[]]``` and ```series 0 "hi" `shouldBe` [[]]```?  I felt strongly about testing size 0 in largest-series-product, but do not feel so strongly here.

rbasso:
Your proposed tests seem correct to me.

Some submissions based on `Data.List.Split.divvy` would certainly break. Because it is implemented in what I humbly call *"the wrong way"*, but there is always a price for changes, and that's ok.

It's you call! If you write the PR, I'll certainly merge it. :smile:

petertseng:
It's now ocurred to me to ask - what would the `divvy` solutions say that `series 0 ""` is?

rbasso:
```
Prelude Data.List.Split> divvy 0 1 ""
[]
```

PR #295, opened 2016-09-12
sieve: Rewrite tests to use hspec with fail-fast.

rbasso:
Related to #211.

### DIFF
--- a/exercises/sieve/test/Tests.hs
+++ b/exercises/sieve/test/Tests.hs
@@ -1,42 +1,70 @@
 {-# OPTIONS_GHC -fno-warn-type-defaults #-}
+{-# LANGUAGE RecordWildCards #-}
 
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
+import Data.Foldable     (for_)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
 import Sieve (primesUpTo)
 
-exitProperly :: IO Counts -> IO ()
-exitProperly m = do
-  counts <- m
-  exitWith $ if failures counts /= 0 || errors counts /= 0 then ExitFailure 1 else ExitSuccess
-
-testCase :: String -> Assertion -> Test
-testCase label assertion = TestLabel label (TestCase assertion)
-
 main :: IO ()
-main = exitProperly $ runTestTT $ TestList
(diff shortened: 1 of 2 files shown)


petertseng:
:+1: 

It makes me realize an interesting thing about the x-common data - "limit is prime" is already covered by 2 case!

exercises/sieve/test/Tests.hs
```diff
@@ -1,42 +1,70 @@
...
-    , 331, 337, 347, 349, 353, 359, 367, 373, 379, 383, 389, 397, 401, 409, 419
-    , 421, 431, 433, 439, 443, 449, 457, 461, 463, 467, 479, 487, 491, 499, 503
-    , 509, 521, 523, 541, 547, 557, 563, 569, 571, 577, 587, 593, 599, 601, 607
-    , 613, 617, 619, 631, 641, 643, 647, 653, 659, 661, 673, 677, 683, 691, 701
-    , 709, 719, 727, 733, 739, 743, 751, 757, 761, 769, 773, 787, 797, 809, 811
-    , 821, 823, 827, 829, 839, 853, 857, 859, 863, 877, 881, 883, 887, 907, 911
-    , 919, 929, 937, 941, 947, 953, 967, 971, 977, 983, 991, 997
-    ] @=? primesUpTo 1000
-  , testCase "first thousand primes" $
-    1000 @=? length (primesUpTo 7919)
```
petertseng: did you want to keep this as a track-specific test?

I'm happy regardless of what the answer is

rbasso: Nice catch!

I just copied the test from `x-common` and forgot about this one. If you think this adds anything to the exercise, I can put it back without problems. :smile:

petertseng: I think most other tracks do not have it, and I don't think it adds a whole lot for us (if they can go up to 1000, they surely can go up to 7919 without problems, right?), so I wouldn't have been sad to lose it. Unless there is a Haskell-specific consideration that I am not aware of :o

petertseng: and in case you wanted to look through history, it's been here since the start at https://github.com/exercism/xhaskell/commit/2fa545b4535d0f30bb75e26c10af871ce4a320e5

rbasso: This is not the first exercise in the Haskell track that tests big numbers without an obvious reason.

If it is there to test performance, it should go in a benchmark suite (this is something that's easy to do now), if it isn't, the test should have a explicit goal.

Let's just drop it! :smile:

PR #296, opened 2016-09-13
series: Test series of zero

petertseng:
This seems a useful case to test. There is one series of size zero in
all series.

The solution does need to be changed here, otherwise it would have said
that `series 0 "012"` was `[[],[],[]]` instead of `[[]]`.

### DIFF
--- a/exercises/series/src/Example.hs
+++ b/exercises/series/src/Example.hs
@@ -4,5 +4,6 @@ import Data.Char (digitToInt)
 import Data.List (tails)
 
 slices :: Int -> String -> [[Int]]
+slices 0 _ = [[]]
 slices n s = map (take n) . take (length s - n + 1) . tails $ numberSeries
     where numberSeries = map digitToInt s
--- a/exercises/series/test/Tests.hs
+++ b/exercises/series/test/Tests.hs
@@ -27,3 +27,7 @@ specs = describe "series" $ do
       slices 3 "ab"   `shouldBe` []
       slices 3 "012"  `shouldBe` [[0,1,2]]
       slices 3 "0123" `shouldBe` [[0,1,2], [1,2,3]]
+
+    it "slices of zero" $ do
+      slices 0 ""    `shouldBe` [[]]
+      slices 0 "012" `shouldBe` [[]]


exercises/series/src/Example.hs
```diff
@@ -4,5 +4,6 @@ import Data.Char (digitToInt)
 import Data.List (tails)
 
 slices :: Int -> String -> [[Int]]
+slices 0 _ = [[]]
```
petertseng: The fact that I had to add the special case makes me wonder about it. As I noted earlier this solution would have said `slices 0 "012" = [[], [], []]` without it. Well, who am I to say that there should only be one subseries of size zero! Maybe it *should* be `[[], [], []]` instead. Maybe this is why it didn't get tested. Let's talk about it.

petertseng: Or how about I make a crazy suggestion! We write the tests to accept either `[[]]` or `[[], [], []]` for `slices 0 "012"` !

rbasso: It we decide to test it, I would say to test it against `[[]]`. Otherwise, if the solution is not unique, I would prefer not to test it at all, because I cannot confidently say that those two are the only acceptable solutions.

petertseng: OK, in that case my recommendation is to merge this as-is. Though really, I would not object to not having it!

Maybe when the day comes when someone creates `series/canonical-data.json` they will decide explicitly to exclude 0... well we can revisit then.

exercises/series/test/Tests.hs
```diff
@@ -27,3 +27,7 @@ specs = describe "series" $ do
       slices 3 "ab"   `shouldBe` []
       slices 3 "012"  `shouldBe` [[0,1,2]]
       slices 3 "0123" `shouldBe` [[0,1,2], [1,2,3]]
+
+    it "slices of zero" $ do
+      slices 0 ""    `shouldBe` [[]]
+      slices 0 "012" `shouldBe` [[]]
```
rbasso: I like the fact that you put the *zero tests* at the end. :+1:

PR #298, opened 2016-09-16
connect: Rewrite test to use hspec with fail-fast.

rbasso:
- Rename `Color(Black,White)` to `Mark(Cross,Nought)`
- Patch the example and the stub solutions to use the new names.

Related to #211.

Now all the tests match `x-common` almost exactly. The only difference is that I kept the expected value as an *ADT*, not a *String* containing `X` or `O`.

Also, I had to change the ADT to have more meaningful names - how would anyone know that `X` was `Black` - so I borrowed them from [Tic-tac-toe's Wikipedia page](https://en.wikipedia.org/wiki/Tic-tac-toe). The alternative would be to use `X` and `O`, but I'm not sure if it is a good idea to have single letter data constructors.

I can change the ADT if someone comes with better names.
Also, considering that we are breaking compatibility with old submissions, we could rename `resultFor`.

##### Updated (3 commits)

- Rewrite tests to use `hspec` with *fail-fast*.

- Rename `Color(Black,White)` to `Mark(Cross,Nought)`.

- Rename `responseFor` to `winner`.

### DIFF
--- a/exercises/connect/package.yaml
+++ b/exercises/connect/package.yaml
@@ -17,4 +17,4 @@ tests:
     source-dirs: test
     dependencies:
       - connect
-      - HUnit
+      - hspec
--- a/exercises/connect/src/Connect.hs
+++ b/exercises/connect/src/Connect.hs
@@ -1,6 +1,6 @@
-module Connect (Mark(..), resultFor) where
+module Connect (Mark(..), winner) where
 
 data Mark = Cross | Nought deriving (Eq, Show)
 
-resultFor :: [String] -> Maybe Mark
-resultFor = undefined
+winner :: [String] -> Maybe Mark
+winner = undefined
--- a/exercises/connect/src/Connect.hs
+++ b/exercises/connect/src/Connect.hs
@@ -1,8 +1,6 @@
-module Connect (resultFor, Color(..)) where
+module Connect (Mark(..), resultFor) where
 
-data Color = Black
-           | White
-           deriving (Show,Eq)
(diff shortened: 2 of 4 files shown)


rbasso:
> I do like the Black/White -> Cross/Nought change since I kept having to look at the README to remember which was which when I was doing this problem.

The fact that the board uses `X` and `O` - both in `README.md` and `canonical-data.json` - but the text in `README.md` explicitly call them `Black` and `White` is confusing.

I'm not sure if I should have changed that here, because I removed all the references to colors in the exercise. Maybe it's a decision to be taken to `x-common`...

Should we open an issue about it there?

> I usually would have liked to put the Black/White -> Cross/Nought in its own commit

You are right! I'll rewrite it, as a little punishment for my lack of discipline! :smile: 

> What would you like for `resultFor`? `winner`?

I just had the feeling that `resultFor` was not good, but had no suggestion.
`winner` seems perfectly clear to me, and makes it obvious why it is a `Maybe. Nice choice!

I'm gonna change it in a separated commit.

petertseng:
> I'm not sure if I should have changed that here, because I removed all the references to colors in the exercise. Maybe it's a decision to be taken to x-common...

> Should we open an issue about it there?

Seems good. You probably have to do it as I will be occupied for a few hours.

rbasso:
Opened exercism/x-common#380.

rbasso:
Thanks for the review!

Rewrote the PR in 3 separated commits, @petertseng.

PR #299, opened 2016-09-16
parallel-letter-frequency: Rewrite tests to use hspec with fail fast.

rbasso:
- Move hints to `HINTS.md`.

Related to #211.

### DIFF
--- a/exercises/parallel-letter-frequency/HINTS.md
+++ b/exercises/parallel-letter-frequency/HINTS.md
@@ -1,0 +1,5 @@
+## Hints
+
+Your code should contain a frequency :: Int -> [Text] -> Map Char Int
+function which accepts a number of workers to use in parallel and a list
+of texts and returns the total frequency of each letter in the text.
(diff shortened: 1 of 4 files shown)


petertseng:
Hmm. guess you can never edit a review summary comment. That's interesting. I was going to edit to link to https://github.com/exercism/xhaskell/commit/1eaff798fa19d4602be0a9b469494df2add99a95 .

rbasso:
> I wonder whether some people will now wonder what the Int arg is if they don't read the README.

We don't have to save the lazy ones, but I think it is nice to save the users the work of looking at the `README.md`.

> If there were a concrete solution to be made, it would be maybe name the args: frequency nWorkers texts = undefined

Granted! :grin: 

> But maybe that will be too prescriptive, if someone wants to write it eta-reduced (I don't know if anyone will)

I don't see that as a problem, because the stub solution is just a suggestion, and I expect the serious student to rewrite it multiple times. What usually bothers me is that I'm never sure about how to name the variables - `workers`, `n`, `nWorkers`, ` texts`, `ts`, `xs` - so I often delegate that choice to the users.

PR #300, opened 2016-09-18
sgf-parsing: Rewrite tests to use hspec with fail-fast.

rbasso:
- Move hints to `HINTS.md`.
- Rewrite tests to use `hspec` with *fail-fast*.

Related to #211.

### DIFF
--- a/exercises/sgf-parsing/test/Tests.hs
+++ b/exercises/sgf-parsing/test/Tests.hs
@@ -8,23 +8,7 @@ import Data.Tree
 import Data.Text (Text)
 import Sgf (parseSgf)
 
--- The Sgf module should export a parseSgf module with the following
--- signature:
---
--- parseSgf :: Text -> Maybe (Tree (Map Text [Text]))
---
--- You may find it useful to copy the type definitions for SgfTree and
--- SgfNode from this file.
---
--- The parsec library is part of the Haskell Platform. Please use it to
--- your advantage.
-
--- | A tree of nodes.
 type SgfTree = Tree SgfNode
-
--- | A node is a property list, each key can only occur once.
---
--- Keys may have multiple values associated with them.
 type SgfNode = Map Text [Text]
 
 exitProperly :: IO Counts -> IO ()
--- a/exercises/sgf-parsing/test/Tests.hs
+++ b/exercises/sgf-parsing/test/Tests.hs
@@ -1,73 +1,39 @@
 {-# LANGUAGE OverloadedStrings #-}
 
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
-import Data.Map (Map)
-import qualified Data.Map as Map
-import Data.Tree
-import Data.Text (Text)
+import Data.Foldable     (for_)
+import Data.Map          (fromList)
+import Data.Tree         (Tree(Node))
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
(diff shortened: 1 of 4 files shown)


exercises/sgf-parsing/test/Tests.hs
```diff
@@ -1,89 +1,39 @@
...
-    let label = "parse " ++ show input
-    in testCase label (expected @=? parseSgf input)
+import Sgf (parseSgf)
 
 main :: IO ()
-main = exitProperly $ runTestTT $ TestList
-       [ TestList sgfParsingTests ]
-
-t :: [(Text, [Text])] -> [SgfTree] -> SgfTree
-t n = Node (Map.fromList n)
```
petertseng: Ah good, I think it was good to remove these, if I had to keep looking back on them to figure out how the test worked it would hvae made things more confusing.

exercises/sgf-parsing/test/Tests.hs
```diff
@@ -1,89 +1,39 @@
...
+
+specs :: Spec
+specs = describe "sgf-parsing" $
+          describe "parseSgf" $ for_ cases test
+  where
+
+    test (input, expected) = it description assertion
+      where
+        description = unwords ["parse", show input]
+        assertion   = parseSgf input `shouldBe` fmap fromList <$> expected
```
petertseng: I had to make sure `fmap` worked as I expected it to on a `Node` (yup, of course it does, no other way the types can work), that's one change that happened now that `t`, `ts` and `tn` are gone, since it made the `fromList` explicit at all levels instead of `fmap`-ping it through. But I think this is a good change since the code becomes more idiomatic and maybe it will encourage exploration.

exercises/sgf-parsing/test/Tests.hs
```diff
@@ -1,89 +1,39 @@
...
+            , (";"                                  , Nothing                                                                         )
+            , ("(;)"                                , Just  $ Node [] []                                                              )
+            , ("(;A[B])"                            , Just  $ Node [("A", ["B"])] []                                                  )
+            , ("(;a)"                               , Nothing                                                                         )
+            , ("(;a[b])"                            , Nothing                                                                         )
+            , ("(;Aa[b])"                           , Nothing                                                                         )
+            , ("(;A[B];B[C])"                       , Just  $ Node [("A", ["B"])] [ Node [("B", ["C"])] []                          ] )
+            , ("(;A[B](;B[C])(;C[D]))"              , Just  $ Node [("A", ["B"])] [ Node [("B", ["C"])] [] , Node [("C", ["D"])] [] ] )
```
petertseng: Will anyone find it harder to read if one node and its children are on the same line? As opposed to making this line

```haskell
, ("(;A[B](;B[C])(;C[D]))"              , Just  $ Node [("A", ["B"])] [
                                                                        Node [("B", ["C"])] []
                                                                      , Node [("C", ["D"])] [] ] )
```

Same with the above line

I'm not sure of the answer, so I wonder what y'all think.

rbasso: Changed!

I agree with you about breaking the children in multiple lines, but I did that just when there where multiple children, to keep the siblings vertically aligned (line 36):

```haskell
... , Just  $ Node [("A", ["B"])] [ Node [("B", ["C"])] [] ] )
... , Just  $ Node [("A", ["B"])] [ Node [("B", ["C"])] []
                                  , Node [("C", ["D"])] [] ] )
```

Is that OK for your?

*ps: It's hard to format trees...*

exercises/sgf-parsing/test/Tests.hs
rbasso: :smile:

PR #306, opened 2016-09-19
new exercise: change

abo64:
implementation of [change](https://github.com/exercism/x-common/tree/master/exercises/change) exercise for Haskell.

### DIFF
--- a/exercises/change/package.yaml
+++ b/exercises/change/package.yaml
@@ -7,7 +7,6 @@ library:
   exposed-modules: Change
   source-dirs: src
   dependencies:
-    - containers
   # - foo       # List here the packages you
   # - bar       # want to use in your solution.
 
--- a/exercises/change/package.yaml
+++ b/exercises/change/package.yaml
@@ -1,0 +1,20 @@
+name: change
+
+dependencies:
+  - base
+
+library:
+  exposed-modules: Change
+  source-dirs: src
+  dependencies:
+    - containers
+  # - foo       # List here the packages you
+  # - bar       # want to use in your solution.
+
+tests:
+  test:
+    main: Tests.hs
+    source-dirs: test
+    dependencies:
+      - change
+      - hspec
--- a/exercises/change/src/Example.hs
+++ b/exercises/change/src/Example.hs
@@ -11,19 +11,20 @@ findFewestCoins target coins = minChange target sortedCoins [] Nothing
   where
     sortedCoins = sortBy (flip compare) coins
 
-    minChange target coins candidate bestResult
-      | target < 0 || worseResult = bestResult
-      | target == 0 = Just candidate
-      | otherwise = dropCoin target coins candidate newBestResult
+    minChange target' coins' candidate bestResult
+      | target' < 0 || worseResult = bestResult
+      | target' == 0 = Just candidate
+      | otherwise = dropCoin target' coins' candidate newBestResult
       where
         worseResult = maybe False (\x -> length x <= length candidate) bestResult
-        newBestResult = addCoin target coins candidate bestResult
+        newBestResult = addCoin target' coins' candidate bestResult
 
-    addCoin target coins@(coin:_) candidate bestResult
-      | newTarget >= 0 = minChange newTarget coins (coin:candidate) bestResult
-      where newTarget = target - coin
+    addCoin target' coins'@(coin:_) candidate bestResult
+      | newTarget >= 0 = minChange newTarget coins' (coin:candidate) bestResult
+      where newTarget = target' - coin
     addCoin _ _ _ bestResult = bestResult
 
-    dropCoin target (_:restCoins) candidate bestResult =
-      minChange target restCoins candidate bestResult
+    dropCoin target' (_:restCoins) candidate bestResult =
+      minChange target' restCoins candidate bestResult
     dropCoin _ _ _ bestResult = bestResult
+
--- a/exercises/change/src/Example.hs
+++ b/exercises/change/src/Example.hs
@@ -19,12 +19,11 @@ findFewestCoins target coins = minChange target sortedCoins [] Nothing
         worseResult = maybe False (\x -> length x <= length candidate) bestResult
         newBestResult = addCoin target' coins' candidate bestResult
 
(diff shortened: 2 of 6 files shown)


rbasso:
I guess you already discovered that, @abo64 , but Travis CI runs `configlet`, so it will refuse a new exercise if it is not added to `config.yaml`.

rbasso:
In what position do you think this exercise should be added?

abo64:
not sure. the algorithm is perhaps not so trivial, so maybe somewhere in the middle?

rbasso:
I would say it is harder than `simple-cipher` for sure, but maybe `connect` and `go-counting` are harder.

I would put it before `connect` for now. After we classify all the exercises, we will reorder everything, so it is just temporary.

exercises/change/test/Tests.hs
```diff
@@ -0,0 +1,76 @@
...
+specs = describe "Change" $
+          describe "change" $ for_ cases test
+  where
+
+    test Case{..} = it description assertion
+      where
+        assertion  = expression `shouldBe` expected
+        expression = findFewestCoins target coins
+
+-- Test cases adapted from `exercism/x-common/change/canonical-data.json` on 2016-09-17.
```
rbasso: Just a note!

The address is now `exercism/x-common/exercises/canonical-data.json`, which is ridiculously long, so I usually just write:
```
-- Test cases adapted from `exercism/x-common` on 2016-09-17.
```

exercises/change/test/Tests.hs
```diff
@@ -0,0 +1,76 @@
...
+        , Case { description = "error testing for change smaller than the smallest of coins",
+                coins = [5, 10],
+                target = 3,
+                expected = Nothing
+               }
+        , Case { description = "cannot find negative change values",
+                coins = [1, 2, 5],
+                target = -5,
+                expected = Nothing
+               }
```
rbasso: The formatting of `Case`s is inconsistent in the list.

There are many ways to format it, but we expected all the cases to have the same exact format.

I used when porting the exercises to `hspec` the following formatting:

```haskell
cases = [ Case { description = "single coin change"
               , coins       = [1, 5, 10, 25, 100]
               , target      = 25
               , expected    = Just [25]
               }
        , Case { ...
               }
        ]
```

But that is not the only acceptable option.

rbasso:
I'm still not used to this new review feature in GitHub, so this time I'll just write normally.

The PR seems great! Thank you @abo64.

You still need to fix the errors that prevents Travis CI to accept your solution.
Also, run `hlint` on your solution. It will give you a suggestion.

exercises/change/package.yaml
```diff
@@ -0,0 +1,20 @@
...
+dependencies:
+  - base
+
+library:
+  exposed-modules: Change
+  source-dirs: src
+  dependencies:
+    - containers
+  # - foo       # List here the packages you
+  # - bar       # want to use in your solution.
```
rbasso: This is cosmetic, and you may or may not change, but in all the other exercises, the used packages are after the comment (those files where generated automatically by a custom script).

The only situation where this may be important, is if we write some script to automatically change those files. Other than that, it is irrelevant.

exercises/change/src/Example.hs
```diff
@@ -0,0 +1,29 @@
...
+type Coin = Integer
+type Coins = [Coin]
+type Amount = Integer
+
+findFewestCoins :: Amount -> Coins -> Maybe Coins
+findFewestCoins target coins = minChange target sortedCoins [] Nothing
+  where
+    sortedCoins = sortBy (flip compare) coins
+
+    minChange target coins candidate bestResult
```
rbasso: This line shadow `target` and `coins` from line 10.

rbasso:
Now the only issue seems to be the warnings about shadowing variable names, that cause error in Travis CI because your new exercises is tests with `stack test --pedantic`.

exercises/change/src/Example.hs
```diff
@@ -0,0 +1,29 @@
...
+      | target < 0 || worseResult = bestResult
+      | target == 0 = Just candidate
+      | otherwise = dropCoin target coins candidate newBestResult
+      where
+        worseResult = maybe False (\x -> length x <= length candidate) bestResult
+        newBestResult = addCoin target coins candidate bestResult
+
+    addCoin target coins@(coin:_) candidate bestResult
```
rbasso: Same shadowing here.

exercises/change/src/Example.hs
```diff
@@ -0,0 +1,29 @@
...
+      | newTarget >= 0 = minChange newTarget coins (coin:candidate) bestResult
+      where newTarget = target - coin
+    addCoin _ _ _ bestResult = bestResult
+
+    dropCoin target (_:restCoins) candidate bestResult =
```
rbasso: More shadowing here. Just `target` this time.

rbasso: Also, there is opportunity for a double eta-reduction.

rbasso: In case you are having problems reproducing the error in Travis, try to test your solution with:

```
stack clean; stack test --pedantic
```

If it passes this test, it will probably work in Travis.

abo64:
about the eta-reductions of `addCoin` and `dropCoin`: compiler complains then about `arity mismatch`?
or am I doing something wrong here?

rbasso:
The compiler seems to enforce that the two definitions of `dropCoin` have the same number of arguments.

You can leave it the way it is, or change the last definition to allow the reduction:
```haskell
    dropCoin _ _ = flip const
```

rbasso:
Anyway, the exercise seems good to merged as it is. :smile:

The one last thing you can do - if you want to practice your git skill - it to squash all of those commits in one or two. One is easier, with two you can separate the change in `config.yaml` if you want.

Just remember that the commit names and messages will be eternalized in this repository's history, and thy name shall never be forgotten, Oh, brave contributor! :smile:

If you don't want to do it, I can squash everything while merging. Your choice!

abo64:
thanks a lot @rbasso 
as always when having a discussion with you I have learnt a lot again.

Issue #301, opened 2016-09-18
Add working stub solutions to refactoring exercises:


rbasso:
It was briefly discussed in https://github.com/exercism/xhaskell/issues/192,

> I think we could keep those three exercises (`beer-song`, `house`, `food-chain`), but maybe we could take the boring part out of some of them.
>
>They are about code refactoring, right?
>Why not providing a working solution?!
>We could also explain what the exercise is about in a `HINTS.md` file...

...that adding working stub solutions to the exercises that are essentially  about refactoring could be a good idea.

> I love the idea of providing the working solution and making it about refactoring.

> :+1: to a refactoring exercise (or a few).

This issue tracks the execution of this idea:

- [x] `beer-song` #390
- [x] `house` #348
- [x] `food-chain` #346

PR #346, opened 2016-09-22
food-chain: Rewrite tests and add working stub solution.

rbasso:
- Use lyrics from `x-common/exercises/food-chain/canonical-data.json`.
- Tests all lines, one by one, and then the entire song.
- Add stub solution passing the tests.
- Patch example solution to work with the new tests.

This is a first try at solving #301. When we get this right, we may extend the same logic to `beer-song` and `house`.

There is some magic in the test suite to check the lyrics line by line, but I couldn't find a better way to write it.

The stub solution has the complete lyrics in the most raw form possible, so that the student will be forced to refactor everything from zero.

The previous tests suite had an empty line at the end, but that makes no sense and doesn't match the reference file.

### DIFF
--- a/exercises/food-chain/HINTS.md
+++ b/exercises/food-chain/HINTS.md
@@ -1,0 +1,19 @@
+## Hints
+
+This exercise is about [code refactoring](https://en.wikipedia.org/wiki/Refactoring),
+so we are providing you with a solution that already passes the tests.
+
+The challenge is to rewrite it until you are proud of it, and learn
+something in the process.
+
+If you don't know where to start, here are some ideas:
+
+- Try to reduce repetition to a minimum.
+- Try to make the code readable.
+- Generalize your solution, allowing the lady to swallow other things.
+
+Take your time.
+
+Change one thing at a time and check if your solution still passes the tests.
+
+Have fun!
--- a/exercises/food-chain/src/Example.hs
+++ b/exercises/food-chain/src/Example.hs
@@ -15,7 +15,7 @@ animalNames = listArray (Fly, Horse) $ map showLower [Fly ..]
                       in (toLower h : t)
 
 song :: String
-song = unlines $ map verse [Fly ..]
+song = init . unlines $ map verse [Fly ..]
 
 verse :: Animal -> String
 verse animal = swallow animal ++ unlines (map catches prey)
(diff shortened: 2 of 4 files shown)


exercises/food-chain/HINTS.md
```diff
@@ -0,0 +1,19 @@
...
+The challenge is to rewrite it until you are proud of it, and learn
+something in the process.
+
+If you don't know where to start, here are some ideas:
+
+- Try to reduce repetition to a minimum.
+- Try to make the code readable.
+- Generalize your solution, allowing the lady to swallow other things.
+
+Take you time.
```
petertseng: "Take your time"

rbasso: Fixed. :smile: 

Anything else?

rbasso:
Thank you for the review, @petertseng! :smile: 

> Were you expecting more opposition?

A little, considering that this is the first exercise where we test two multi-line string, line by line.
Also, I have to admit that my custom *zipWith* function (`go`) is at least unusual. :smile: 

> ...keep in mind I haven't done this exercise in any language.

In that case I recommend you to try this one when you get the time. 

It may seem pointless at first, but I think I learn a few things refactoring it.

petertseng:
I spent a little more time on ```go shouldBe `on` zip [1 :: Int ..] . lines``` than `go` since `go` was easy enough to understand. I'm not sure I could have done better than `go` since the goal is to stop at the longer of the lists rather than the shorter.

> In that case I recommend you to try this one when you get the time. 

I'll start now.

petertseng:
Hmm, here's one thing that caught my eye. The failure message -

```
Failures:

  test/Tests.hs:27: 
  1) food-chain.song matches lines
       expected: Just (2,"I don't know why she swallowed the fly. Perhaps she'll die.")
        but got: Just (2,"Hmm, I messed up")
```

It may not be clear to the reader that this means "there was a mismatch at line 2", maybe?

rbasso:
Yeah! I know... it is far from good.

The problem is that, to compare with a possibly missing line, we need to compare *Maybes*. I don't know how to customize an error message... I need to learn more about `hspec`.

About the line number, we can have it in the data compared, like we did, or in the test description. In the later case, it would create one test case for each line, and that creates a lot of output.

I agree that it deserves to be improved, but I still don't know exactly how. :confused:

PR #348, opened 2016-09-23
house: Fix tests to match x-common and add working stub solution.

rbasso:
- Add `HINTS.md` explaining that this exercise is about refactoring.
- Add a stub solution that passes the tests.
- Remove extra newline in the tests to match `description.md` in `x-common`.
- Patch example solution to pass the new tests.

Related to #301.

### DIFF
--- a/exercises/house/HINTS.md
+++ b/exercises/house/HINTS.md
@@ -1,0 +1,18 @@
+## Hints
+
+This exercise is about [code refactoring](https://en.wikipedia.org/wiki/Refactoring),
+so we are providing you with a solution that already passes the tests.
+
+The challenge is to rewrite it until you are proud of it, and learn
+something in the process.
+
+If you don't know where to start, here are some ideas:
+
+- Try to reduce repetition to a minimum.
+- Try to make the code readable.
+
+Take your time.
+
+Change one thing at a time and check if your solution still passes the tests.
+
+Have fun!
--- a/exercises/house/src/Example.hs
+++ b/exercises/house/src/Example.hs
@@ -1,7 +1,9 @@
 module House (rhyme) where
 
+import Data.List (intercalate)
+
 rhyme :: String
-rhyme = unlines $ map ("This is " ++) $ scanl1 f pieces
+rhyme = intercalate "\n" $ map ("This is " ++) $ scanl1 f pieces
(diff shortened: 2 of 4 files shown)


petertseng:
Oh, no line by line comparison this time? Is this OK? You know what would be really awesome? If hspec has a specialised `shouldBe` on strings so that it shows you a line-by-line diff if there is a mismatch... This is a lazy question, I haven't tested whether there is such a thing yet but I didn't find one after a few quick searches a few hours ago when discussing food-chain. I'll take another look in a few hours

exercises/house/HINTS.md
```diff
@@ -0,0 +1,18 @@
...
+
+The challenge is to rewrite it until you are proud of it, and learn
+something in the process.
+
+If you don't know where to start, here are some ideas:
+
+- Try to reduce repetition to a minimum.
+- Try to make the code readable.
+
+Take your time.
```
petertseng: https://www.youtube.com/watch?v=RuqaVryDRd0&t=10

Ignore this comment, it is not serious.

rbasso:
> Oh, no line by line comparison this time? Is this OK?

I was not sure if I should use the same structure as `food-chain`, because of the uncommon error messages.

Well, I guess that is better than nothing. :smile: 

Fixing it now!

> You know what would be really awesome? If hspec has a specialised shouldBe on strings so that it shows you a line-by-line diff if there is a mismatch...

I completely agree with you! That would be perfect!

I'll try to remember to take a look at that later, to see if we can write something like this.
But first I need to finish the infinite migration to `hspec`. Five to go! :tired_face:

rbasso:
Tests updated to use @petertseng's code to compare the lyrics, taken from `food-chain`.
Thanks again for that. :smile:

rbasso:
I haven't tested it yet, but I guess there is a `shouldBe` diff function, @petertseng !

> You know what would be really awesome? If hspec has a specialised shouldBe on strings so that it shows you a line-by-line diff if there is a mismatch...

https://github.com/myfreeweb/hspec-expectations-pretty-diff#readme

It is already available in the current snapshot: https://www.stackage.org/nightly-2016-07-17/package/hspec-expectations-pretty-diff-0.7.2.4

PR #350, opened 2016-09-24
food-chain: Show clearer line-by-line match messages

petertseng:
Previously, if there was a mismatch, the output would look like:

```
  1) food-chain.song matches lines
       expected: Just (4,"I know an old lady who swallowed a spider.")
        but got: Just (4,"I know an old lady who swallowed a thing.")
```

It's not necessarily clear that the 4 in the output means "the mismatch
was at line 4".

Now the error reads:

```
  1) food-chain.song matches lines
       mismatch at line 4
       expected: Just "I know an old lady who swallowed a spider."
        but got: Just "I know an old lady who swallowed a thing."
```

### DIFF
--- a/exercises/food-chain/test/Tests.hs
+++ b/exercises/food-chain/test/Tests.hs
@@ -1,5 +1,5 @@
-import Data.Function     (on)
-import Test.Hspec        (Spec, describe, it, shouldBe)
+import Control.Monad     (unless)
+import Test.Hspec        (Spec, describe, expectationFailure, it, shouldBe)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
 
 import FoodChain (song)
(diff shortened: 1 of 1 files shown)


exercises/food-chain/test/Tests.hs
```diff
@@ -1,6 +1,8 @@
-import Data.Function     (on)
-import Test.Hspec        (Spec, describe, it, shouldBe)
-import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+import Control.Monad           (unless)
+import Data.Function           (on)
```
petertseng: don't need this anymore, removing

exercises/food-chain/test/Tests.hs
```diff
@@ -24,12 +25,16 @@ specs = describe "food-chain" $
             it "matches full song" $ song `shouldBe` lyrics
   where
 
-    lineAssertions = (go shouldBe `on` zip [1 :: Int ..] . lines) song lyrics
+    lineAssertions = zipWith checkLine [1 :: Int ..] $ zipMaybe (lines song) (lines lyrics)
```
petertseng: I was not very clever here. I suppose I can keep the `on` and extract the line number in `checkLine` (so `checkLine (lineno, got), (_, want)`)

petertseng:
This is still not as good as a diff-with-context, but it's something.

exercises/food-chain/test/Tests.hs
rbasso: I won't judge you. :smile:

exercises/food-chain/test/Tests.hs
```diff
@@ -24,12 +25,16 @@ specs = describe "food-chain" $
...
-    go f    []  (y:ys) = f Nothing  (Just y) : go f [] ys
-    go f (x:xs) (y:ys) = f (Just x) (Just y) : go f xs ys
+    checkLine lineno (got, want) =
+      unless (got == want) $
+        expectationFailure $ "mismatch at line " ++ show lineno ++ "\nexpected: " ++ show want ++ "\n but got: " ++ show got
+
+    zipMaybe    []     []  = []
+    zipMaybe (x:xs)    []  = (Just x , Nothing) : zipMaybe xs []
+    zipMaybe    []  (y:ys) = (Nothing, Just y ) : zipMaybe [] ys
+    zipMaybe (x:xs) (y:ys) = (Just x , Just y ) : zipMaybe xs ys
```
rbasso: Certainly a better name than `go`. :smile:

rbasso:
> This is still not as good as a diff-with-context, but it's something.

It's great the way it is.

A diff would be great, but I'm not sure if it is a simple task (any package for that?).
Maybe later we can come up with something simpler, like a `shouldMatchOrderedList`.

By the way, I regret not knowing `shouldMatchList` before, I could have used it in the test suites that order results before comparing.

exercises/food-chain/package.yaml
```diff
@@ -18,3 +18,4 @@ tests:
     dependencies:
       - food-chain
       - hspec
+      - hspec-expectations
```
petertseng: You know something strange... for some reason I tried importing it from `Test.Hspec.Runner`, didn't find it there, and forgot to look in `Test.Hspec`!

petertseng:
OK, taking `expectationFailure` from `Test.Hspec`, I am ready now :+1:

PR #352, opened 2016-09-24
go-counting: Rewrite tests to use hspec with fail-fast.

rbasso:
- Completely rewrite tests to use `hspec`.
- Add test suite's dependency on package `multiset`.

Related to #211.

This was one of the most challenging re-writings since we started migrating the exercises to `hspec`, so I think I owe a few explanations:

- `shouldScore` is pointfree madness, but I couldn't make it more readable any other way.
- `toOccurList . fromOccurList` is the best way I could find to sum all the scores.
- `first` and `swap` are not commonly used in this repository, but I think they really make sense.

### DIFF
--- a/exercises/go-counting/package.yaml
+++ b/exercises/go-counting/package.yaml
@@ -18,4 +18,5 @@ tests:
     source-dirs: test
     dependencies:
       - go-counting
-      - HUnit
+      - hspec
+      - multiset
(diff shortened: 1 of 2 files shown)


petertseng:
> `shouldScore` is pointfree madness, but I couldn't make it more readable any other way.

It didn't seem so hard to understand to me. It gets harder for me if I see compositions of compositions, like

```
 > pl \x y z -> f (g x y z)
 ((f .) .) . g
```

> `toOccurList . fromOccurList` is the best way I could find to sum all the scores.

That seems OK. Shorter than what I might have come up with (which would be folding with [insertWithKey](https://hackage.haskell.org/package/containers-0.5.8.1/docs/Data-Map-Strict.html#v:insertWith) then `toList`

petertseng:
(OK I realize `shouldHaveTerritories` is pretty much the same as `terrs` helper of the previous, but `shouldScore` I think is better by making clear which score belongs to whom)

rbasso:
Thank you very much, @petertseng! :smile:

PR #353, opened 2016-09-24
documentation: Update README.md with more detailed test instrucions.

rbasso:
Running `stack test --pedantic` will not recompile unchanged
source code already compiled with warnings.

To be really sure that everything compiles correctly without
warnings, it is necessary to first run `stack clean`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,13 +29,24 @@ format that has all dependencies and build instructions for an exercise.
 
 ### Running Tests
 
-Rename the file `src/Example.hs` to match the module name and run:
+In order to be accepted by Travis-CI, every exercise must compile without
+warnings and the example solution must pass the tests without failures.
+
+To test an exercise, first you need to rename the file `src/Example.hs`
+to match the module name, but - because we already have a *stub solution*
+in place - we have to move it first, otherwise it will be overwritten.
+
+Running `stack test --pedantic` compiles and run the tests with
+`-Wall -Werror`, but unfortunately it doesn't recompile unchanged
+source code already compiled with warnings.
+
+To be really sure that everything compiles correctly without
+warnings, your must first run `stack clean`.
 
 ```bash
+stack clean
 stack test --pedantic
 ```
-
-If the stub solution is still in the `/src` folder, build will probably fail.
 
 ### Running HLint
 


petertseng:
It wouldn't be an issue if everyone always used `--pedantic`, eh? But I sometimes forget, so this seems useful

rbasso:
Considering that only the potential collaborators will read the `README.md` file... 

> It wouldn't be an issue if everyone always used `--pedantic`, eh?

Why would that be a problem :question: 

I think it would be great if everybody run it before opening a PR. This would anticipate failures in Travis CI, giving a faster feedback.

I use `--pedantic` most of the time.

petertseng:
Sorry, I was unclear.

I meant to say "If everyone used `--pedantic` all the time, we don't need to remind them to use `stack clean`", but I think you thought I was asking "It would be a problem if everyone used `--pedantic` all the time"

rbasso:
> If everyone used `--pedantic` all the time, we don't need to remind them to use `stack clean`

You are completely right, @petertseng! :+1: 

I updated this PR to include more detailed information on how to run the tests.
It is far from perfect, but it is a start, and it is more pedagogical than the previous version.

What do you think of it?

README.md
```diff
@@ -29,13 +33,37 @@ format that has all dependencies and build instructions for an exercise.
...
+to match the module name, but - because we already have a *stub solution*
+in place - we have to move it first, otherwise it will be overwritten.
+
+```bash
+mv src/ModuleName.hs stub
+mv src/Example.hs src/ModuleName.hs
+```
+
+Running `stack test --pedantic` compiles and run the tests with
+`-wAll -wError`, but unfortunately it doesn't recompile unchanged
```
petertseng: I think this is `-Wall` and `-Werror`. Unless this flag is case insensitive, in which case I learn something!

petertseng:
regarding the instructions to move the stub file aside and then move it back after done: I imagine that this is the most basic workflow, and I hope others will know to adapt it to fit own needs. For example I just mv the example file without backing up the stub file, then `git add -p`, `git commit`, then `git reset --hard`. I wouldn't put all that into the README though so what is currently here is good.

In another track, at https://github.com/exercism/xgo/pull/249 I was requested to have a script to automate the process of moving the stub file aside, running the test, and moving the stub file back - it avoided the problem of someone forgetting to move the stub back and accidentally committing it. If people complain that it's so much work to have to do four `mv` every time wanting to change a file, we could consider a similar thing here.

README.md
rbasso: You are absolutely right!
Nice catch. :smile:

rbasso: Fixed!

rbasso:
> For example I just mv the example file without backing up the stub file...

Same here.

> ... then `git add -p`, `git commit`, then `git reset --hard`.

Never heard of `git add -p` before. Thanks. :smile: 

> I wouldn't put all that into the README though so what is currently here is good.

What if I simply remove the following lines:

```
mv src/ModuleName.hs stub
mv src/Example.hs src/ModuleName.hs

...

Before commiting your changes, you need to move the *stub* and sample
solution back to their places:

mv src/ModuleName.hs src/Example.hs
mv stub src/ModuleName.hs

```

Do you think that it would be better?

rbasso:
> In another track, at exercism/xgo#249 I was requested to have a script to automate the process of moving the stub file aside, running the test, and moving the stub file back...

What I don't like about it is adding code that is dependent the local setup of contributors.
Even in the documentation. I think I should not include the lines:

```
All examples in this document assume that you are running a Unix-like
environment, e.g. GNU/Linux, Darwin. While this is not mandatory, you will
probably have to adapt some commands if you are running a different system.
```

petertseng:
> What if I simply remove the following lines:

That seems OK. It is not too bad of a loss. Those lines can't simply be copy/pasted anyway since `ModuleName` differs per module, so it is of limited value ot have it.

rbasso:
Removed those lines and also the following ones, that are not needed anymore:

```
All examples in this document assume that you are running a Unix-like
environment, e.g. GNU/Linux, Darwin. While this is not mandatory, you will
probably have to adapt some commands if you are running a different system.
```

Issue #355, opened 2016-09-24
hlint: Improve existing code based on suggestions


rbasso:
To promote good coding practices, we are changing `README.md` to recommend contributors to run `HLint` before every submission. #354

That may help in the future, but running `hlint` on `master` indicates that there is also a lot of work to be done right now. The following is a list of files to inspect and decide if we should accept *HLint*'s suggestions:

- [X] `allergies/src/Example.hs` #356 
- [X] `all-your-base/src/Example.hs` #357 
- [X] `all-your-base/test/Tests.hs` #357 
- [x] `bank-account/test/Tests.hs` #379
- [x] `binary-search-tree/test/Tests.hs` #375 
- [X] `bob/src/Example.hs` #358 
- [X] `clock/test/Tests.hs` #359
- [X] `custom-set/src/Example.hs` #360 
- [X] `custom-set/test/Tests.hs` #360 
- [X] `forth/src/Example.hs` #361 
- [X] `forth/test/Tests.hs` #361 
- [X] `go-counting/src/Counting.hs` #362 
- [X] `go-counting/src/Example.hs` #362 
- [X] `hamming/src/Example.hs` #363 
- [x] `hexadecimal/src/Example.hs` #376 (one hint ignored)
- [x] `linked-list/test/Tests.hs` #377 (three hints ignored)
- [X] `list-ops/src/Example.hs` #370, #373 (one hint ignored)
- [X] `matrix/test/Tests.hs` #364 
- [X] `minesweeper/src/Example.hs` #365 
- [x] `ocr-numbers/src/Example.hs` #378 
- [X] `pascals-triangle/test/Tests.hs` #366 
- [X] `pig-latin/src/Example.hs` #367 
- [X] `pig-latin/test/Tests.hs` #367 
- [x] `pov/src/Example.hs` #371
- [x] `pov/test/Tests.hs` #371 
- [X] `raindrops/src/Example.hs` #369 
- [x] `robot-name/src/Example.hs` #380 
- [x] `robot-name/test/Tests.hs` #380 
- [X] `simple-cipher/src/Example.hs` #372 
- [X] `simple-linked-list/src/Example.hs` #368

petertseng:
When this is all said and done perhaps we should consider adding it to travis. On a quick look... it seems a bit strange that there is a *parse error* on `robot-name/src/Example.hs` ...?!

petertseng:
[...]

PR #361, opened 2016-09-25
forth: Improved code based on HLint's suggestions.

rbasso:
Related to #355.


petertseng:
Could have waited until hunit->hspec for the test file (I imagine the `do` would have gone away with the rewrite) but no need to wait now, easy win

PR #368, opened 2016-09-25
simple-linked-list: Improve code base on HLint's suggestions.

rbasso:
Related to #355.


petertseng:
Interesting that hlint can detect even this

PR #371, opened 2016-09-25
pov: Improve code based on HLint's suggestions.

rbasso:
Related to #355.

### DIFF
--- a/exercises/pov/src/Example.hs
+++ b/exercises/pov/src/Example.hs
@@ -1,13 +1,13 @@
 module POV (Graph(Graph), fromPOV, tracePathBetween) where
 
-import Data.Maybe(listToMaybe, catMaybes)
+import Data.Maybe(listToMaybe, mapMaybe)
 
 data Graph a = Graph { value :: a, children :: [Graph a] } deriving (Show, Eq)
 data Crumb a = Crumb a [Graph a] [Graph a] deriving (Show, Eq)
-data Zipper a = Zipper { node :: (Graph a), path ::  [Crumb a] }
+data Zipper a = Zipper { node :: Graph a, path :: [Crumb a] }
 
 fromPOV :: Eq a => a -> Graph a -> Maybe (Graph a)
-fromPOV x = (fmap reparent) . (findLoc x) . rootZipper
+fromPOV x = fmap reparent . findLoc x . rootZipper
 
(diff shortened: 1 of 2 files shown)


exercises/pov/src/Example.hs
```diff
@@ -1,40 +1,40 @@
...
 
 reparent :: Eq a => Zipper a -> Graph a
 reparent (Zipper g []) = g
-reparent (Zipper g (c:cs)) = Graph (value g) $ (children g) ++ [reparented]
+reparent (Zipper g (c:cs)) = Graph (value g) $ children g ++ [reparented]
     where reparented = reparent (Zipper (crumbToGraph c) cs)
 
 down :: Zipper a -> Maybe (Zipper a)
-down (Zipper (Graph v (k:kids)) crumbs) = Just (Zipper k ((Crumb v [] kids):crumbs))
+down (Zipper (Graph v (k:kids)) crumbs) = Just (Zipper k (Crumb v [] kids:crumbs))
```
petertseng: Do you think the spacing now makes it look deceptive? It makes it look as if the precedence is `Crumb v [] (kids:crumbs)`, perhaps a space surrounding the `:` would be fitting? What say you?

exercises/pov/src/Example.hs
```diff
@@ -1,40 +1,40 @@
...
 down _ = Nothing
 
 right :: Zipper a -> Maybe (Zipper a)
-right (Zipper here ((Crumb v lefts (r:rights)):cs)) = Just (Zipper r (shifted:cs))
+right (Zipper here (Crumb v lefts (r:rights):cs)) = Just (Zipper r (shifted:cs))
```
petertseng: same question with the space here, I think

exercises/pov/src/Example.hs
rbasso: I agree!
Gonna fix as soon as possible.

ps: Sorry for taking so long to answer. I'm temporarily with a very slow Internet connection (It takes at least a minute to load a GitHub page)

exercises/pov/src/Example.hs
rbasso: I rewrote both lines, adding spaces and substituting parenthesis by `$`.
I still don't like how it looks, but I expect it to be a little more readable.

PR #377, opened 2016-09-26
linked-list: Supress HLint's suggestions to reduce duplication.

rbasso:
Related to #355.

The following suggestions would probably make the tests shorter, but harder to read:

```
./test/Tests.hs:16:7: Suggestion: Reduce duplication
Found:
  deque <- mkDeque
  push deque 'a'
  push deque 'b'
Why not:
  Combine with ./test/Tests.hs:23:7

./test/Tests.hs:16:7: Suggestion: Reduce duplication
Found:
  deque <- mkDeque
  push deque 'a'
  push deque 'b'
  pop deque `shouldReturn` Just 'b'
Why not:
  Combine with ./test/Tests.hs:44:7

./test/Tests.hs:30:7: Suggestion: Reduce duplication
Found:
  deque <- mkDeque
  unshift deque 'a'
  unshift deque 'b'
Why not:
  Combine with ./test/Tests.hs:37:7

3 hints
```

It is not clear why, but `hlint` doesn't supresses the suggestions if we use `{-# ANN specs ...`, only with `{-# ANN module ...`.

### DIFF
--- a/exercises/linked-list/test/Tests.hs
+++ b/exercises/linked-list/test/Tests.hs
@@ -6,6 +6,7 @@ import Deque (mkDeque, pop, push, shift, unshift)
 main :: IO ()
 main = hspecWith defaultConfig {configFastFail = True} specs
 
+{-# ANN module "HLint: ignore Reduce duplication" #-}
 specs :: Spec
 specs = describe "linked-list" $ do
 


petertseng:
By the way!

> It is not clear why, but hlint doesn't supresses the suggestions if we use `{-# ANN specs ...`, only with `{-# ANN module ...`.

If you want to know my hypothesis, the `{-# ANN specs` would affect the *definition* of `specs`. We are not *defining* `specs`, but *using* it.

petertseng:
Actually, never mind, we definitely *define* `specs`. ?__?

petertseng:
(I originally posed a hypothesis that `{-# ANN specs ...` was not effective since we were not defining `specs`, but we in fact are, so the hypothesis is wrong. Oh well. Leaving it here in case it gives someone any better ideas, and so there is a record of past mistakes I made)

petertseng:
My next best guess is simply that HLint only respects annotations on functions or else the entire module, and `specs` isn't a function. Well, at least, functions are the only examples given in http://community.haskell.org/~ndm/darcs/hlint/hlint.htm so it's not completely out of the question, but who knows...

rbasso:
> My next best guess is simply that HLint only respects annotations on functions or else the entire module, and specs isn't a function.

That is a good theory. :+1: , but why would it be treated differently?! They are just names and types... whatever... :smile:

PR #379, opened 2016-09-26
bank-account: Supress HLint suggestion

rbasso:
Related to #355.

The following suggestion would probably make the tests
shorter, but harder to read:

```
./test/Tests.hs:29:9: Suggestion: Reduce duplication
Found:
  account <- openAccount
  getBalance account `shouldReturn` Just 0
  incrementBalance account 10 `shouldReturn` Just 10
Why not:
  Combine with ./test/Tests.hs:47:9

1 hint
```

It is not clear why, but `hlint` doesn't supresses the suggestions
if we use `{-# ANN specs ...`, only with `{-# ANN module ...`.

### DIFF
--- a/exercises/bank-account/test/Tests.hs
+++ b/exercises/bank-account/test/Tests.hs
@@ -46,7 +46,7 @@ specs = describe "bank-account" $ do
     it "closed banks hold no balance" $ do
         account    <-    openAccount
         getBalance       account    `shouldReturn` Just  0
-        incrementBalance account 10 `shouldReturn` Just 10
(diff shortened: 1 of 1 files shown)


petertseng:
Another possibility is... to no longer duplicate! For example if we change done of those `10` to a `15` or something.

It is not really important to me which of these two choices we take (ignore duplication hint, or change a value)

rbasso:
> Another possibility is... to no longer duplicate! For example if we change done of those 10 to a 15 or something.

That is a great idea! Let's try it! :smile:

PR #380, opened 2016-09-26
robot-name: Fix HLint's suggestions and errors

rbasso:
Related to #355.

- Remove variable named `pattern` in `src/Example` to workaround HLint's bug that causes a parse error.

```
./src/Example.hs:19:31: Error: Parse error: pattern
Found:
    generateName :: IO String
  > generateName = mapM randomRIO pattern
      where pattern = [ letter, letter, digit, digit, digit ]
            letter = ('A', 'Z')
```

- Simplify code with `fmap`, as suggested.

```
./src/Example.hs:10:11: Suggestion: Use fmap
Found:
  generateName >>= newMVar >>= return . Robot
Why not:
  fmap Robot (generateName >>= newMVar)
```

- Supress warning to reduce code duplication that would harm readability.

```
./test/Tests.hs:34:13: Warning: Reduce duplication
Found:
  n1 <- robotName r
  n2 <- robotName r
  n3 <- robotName r
  n1 `shouldBe` n2
  n1 `shouldBe` n3
Why not:
  Combine with ./test/Tests.hs:53:13
```

### DIFF
--- a/exercises/robot-name/src/Example.hs
+++ b/exercises/robot-name/src/Example.hs
@@ -7,7 +7,7 @@ import System.Random (randomRIO)
 data Robot = Robot { robotNameVar :: MVar String }
 
 mkRobot :: IO Robot
-mkRobot = generateName >>= newMVar >>= return . Robot
+mkRobot = fmap Robot $ generateName >>= newMVar
 
 resetName :: Robot -> IO ()
 resetName r = void (generateName >>= swapMVar (robotNameVar r))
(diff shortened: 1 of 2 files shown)


petertseng:
I feel like it may not be so bad here to have a ```r `shouldSatisfy` nameIsPersistent```, but I'm not fussed. The difference too small (benefit not as big as in `go-counting`)

rbasso:
> I feel like it may not be so bad here to have a ``r `shouldSatisfy` nameIsPersistent`` ...

Fixed, but I couldn't use `shouldSatisfy` because `nameIsPersistent` cannot be `Robot -> Bool`. It is in `IO`.

Issue #302, opened 2016-09-18
Move hints from test suites and stub solutions to HINTS.md


rbasso:
Currently we have some exercises with hints as comments in the test suite and/or in the stub solution.

Hints in the stub solutions are convenient - users are almost forced to read them - but proved to be a mess, because most users submit the code with the comments.

Hints in the test suite are not that bad, but they force the users to inspect it. I think that, ideally, users should get feedback running the tests, not studying them.

Anyway, considering that the contents of `HINTS.md` is automatically included in the `README.md`, it is the perfect place for hints. We just need to make it happen!

All the exercises that already have a `HINTS.md` file share the same formatting,

```markdown
## Hints

```

so we are using this as a standard for now.

~~This issue will track the migration's progress as soon as we have a complete list of exercises to be changed.~~

#### Tracking changes

- [x] allergies #324
- [x] bank-account #325
- [x] binary-search-tree #329 
- [x] clock #330
- [x] custom-set #331
- [x] forth #343
- [x] go-counting #344
- [x] grade-school #332
- [x] lens-person #381 (comments removed when ported to `hspec`)
- [x] linked-list #333
- [x] matrix #334
- [x] meetup #335 
- [x] octal #383
- [X] ~~pov (nothing to fix)~~
- [x] queen-attack #336 
- [x] rna-transcription #382 (comments removed)
- [x] robot-name #337
- [x] robot-simulator #338 
- [x] sgf-parsing #300 (fixed when ported to `hspec`)
- [x] simple-linked-list #339
- [x] space-age #340 
- [x] sublist #341 
- [x] trinary #384
- [x] wordy #342 
- [x] zipper #385 (comments removed when ported to `hspec`)

rbasso:
Hints' migration to `HINTS.md` files finished. :smile:

PR #381, opened 2016-09-27
hspec-lens-person: Rewrite test to use hspec with fail-fast.

rbasso:
- Remove unneeded data type `Gregorian` and rewrite stub.

Related to #211 and #302.

### DIFF
--- a/exercises/lens-person/src/Person.hs
+++ b/exercises/lens-person/src/Person.hs
@@ -1,39 +1,34 @@
-module Person where
+module Person
+  ( Address (..)
+  , Born    (..)
+  , Name    (..)
+  , Person  (..)
+  , bornStreet
+  , renameStreets
+  , setBirthMonth
+  , setCurrentStreet
+  ) where
 
-import           Data.Time.Calendar
+import Data.Time.Calendar (Day)
 
-data Person = Person {
-      _name    :: Name,
-      _born    :: Born,
-      _address :: Address
-    }
+data Person = Person { _name    :: Name
+                     , _born    :: Born
+                     , _address :: Address
+                     }
 
-data Name = Name {
-      _foreNames :: String, -- Space separated
-      _surName   :: String
-    }
+data Name = Name { _foreNames :: String
+                 , _surName   :: String
+                 }
 
-data Born = Born {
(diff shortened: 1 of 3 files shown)


exercises/lens-person/src/Person.hs
```diff
@@ -1,49 +1,43 @@
...
+                       , _place       :: String
+                       , _country     :: String
+                       }
 
 bornStreet :: Born -> String
 bornStreet = undefined
 
-setCurrentStreet :: String -> Person -> Person
-setCurrentStreet = undefined
+renameStreets :: (String -> String) -> Person -> Person
```
petertseng: Oh, I see. You made the stub file follow alphabetical order.

The old argument for having them in this order is that a plain set is probably easier than a transform (sorry that I forget the lens specific terminology), and if students implement from top to bottom, they should probably implement the easier ones first.

However, at this point the tests should be guiding the implementation order since the tests will be run in order and the first one will fail.

I'm okay with this, let me know if I missed anything in my reasoning and/or there are other concerns

exercises/lens-person/src/Person.hs
```diff
@@ -1,49 +1,43 @@
...
-data Address = Address {
-      _street      :: String,
-      _houseNumber :: Int,
-      _place       :: String, -- Village / city
-      _country     :: String
-    }
-
--- Valid values of Gregorian are those for which 'Data.Time.Calendar.fromGregorianValid'
--- returns Just.
-data Gregorian = Gregorian {
```
petertseng: yeah I never understood this; I just deleted it in my submission

exercises/lens-person/src/Person.hs
rbasso: I'm not sure if top level functions should be defined in a solution specific order or if they should be alphabetical.

Alphabetical order usually makes finding things easier, and I usually do the same in the exports list, to match. But I don't have a very strong opinion on this.

I left the data declarations in topological order. I'm inconsistent. :grin: 

Should I revert the functions to their old ordering?

rbasso: Fixed. Just reverted to the old order. :+1:

PR #382, opened 2016-09-27
rna-transcription: Remove unneeded stub comments.

rbasso:
In #302, we are moving all the hints in the test suites and stub solutions to `HINTS.md` files. But the comments in `src/DNA.hs` are too obvious and may take all the fun of solving the exercise, so we are just removing them.


petertseng:
Yup, seems OK to remove it. the students probably won't lose too much from it, since this is the natural use of `Maybe`. If they do not know it, finding it out may help to remember it more ffectively than just reading about it.

PR #387, opened 2016-09-30
pov: Rewrite tests to use hspec with fail-fast

rbasso:
- Rewrite tests to use hspec with fail-fast.
- Use Data.Tree from `containers` package instead of a custom ADT.

Related to #211.

I changed the formatting a lot trying to make the test suite more readable. It is longer, but maybe a little better.

The most important change was removing the ADT called `Graph`, that was in fact a `Tree`, so that users will learn a little more from package `containers`.


exercises/pov/test/Tests.hs
```diff
@@ -1,28 +1,81 @@
...
+specs = describe "pov" $ do
 
-x :: String
-x = "x"
+    -- As of 2016-09-28, there was no reference file
+    -- for the test cases in `exercism/x-common`.
 
-root :: Graph a -> a
-root (Graph r _) = r
+    describe "tracePathBetween" $ do
```
petertseng: interesting to see that even in the original, before this PR, `tracePathBetween` tests came before `fromPOV`.

I would like to ask whether you think it is appropriate to change so that `fromPOV` comes first. The reason is that the README says that we can use the re-rooted tree to find the path. So I imagine that rerooting the tree will be implemented before finding the path. You can see this too in the example solution - the example solution uses `fromPOV` in its implementation of `tracePathBetween`. Since we are stopping at the first failed test, now it seems important to put `fromPOV` first.

exercises/pov/test/Tests.hs
```diff
@@ -1,28 +1,81 @@
...
+
+          in for_ cases test
+
+      describe "Should not be able to find a missing node" $
+
+        let cases = [ ("singleton", singleton)
+                    , ("flat"     , flat     )
+                    , ("kids"     , kids     )
+                    , ("nested"   , nested   )
+                    , ("cousings" , cousins  ) ]
```
petertseng: probably `cousins`

exercises/pov/test/Tests.hs
rbasso: Makes complete sense for me. Changed! :smile:

exercises/pov/test/Tests.hs
rbasso: Fixed. Thanks!

PR #388, opened 2016-10-02
forth: Rewrite tests to use hspec with fail-fast.

rbasso:
- ~~~Move hint from test suite to `HINTS.md`.~~~

Related to #211.

### DIFF
--- a/exercises/forth/test/Tests.hs
+++ b/exercises/forth/test/Tests.hs
@@ -1,74 +1,78 @@
 {-# LANGUAGE OverloadedStrings #-}
-import Test.HUnit (Assertion, (@=?), runTestTT, Test(..), Counts(..))
-import System.Exit (ExitCode(..), exitWith)
-import Forth (ForthError(..), evalText, empty, formatStack)
-import Control.Monad (foldM)
-import Data.Text (Text)
 
-exitProperly :: IO Counts -> IO ()
-exitProperly m = do
-  counts <- m
-  exitWith $ if failures counts /= 0 || errors counts /= 0 then ExitFailure 1 else ExitSuccess
+import Control.Monad     (foldM)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
 
-testCase :: String -> Assertion -> Test
-testCase label assertion = TestLabel label (TestCase assertion)
+import Forth (ForthError(..), empty, evalText, formatStack)
 
 main :: IO ()
-main = exitProperly $ runTestTT $ TestList
-       [ TestList forthTests ]
+main = hspecWith defaultConfig {configFastFail = True} specs
 
-runTexts :: [Text] -> Either ForthError Text
-runTexts xs = formatStack <$> foldM (flip evalText) empty xs
+specs :: Spec
+specs = describe "forth" $ do
 
-forthTests :: [Test]
-forthTests =
-  [ testCase "no input, no stack" $
-    "" @=? formatStack empty
-  , testCase "numbers just get pushed onto the stack" $
-    Right "1 2 3 4 5" @=? runTexts ["1 2 3 4 5"]
-  , testCase "non-word characters are separators" $
-    -- Note the Ogham Space Mark ( ), this is a spacing character.
-    Right "1 2 3 4 5 6 7" @=? runTexts ["1\NUL2\SOH3\n4\r5 6\t7"]
-  , testCase "basic arithmetic" $ do
-    Right "-1" @=? runTexts ["1 2 + 4 -"]
-    Right "2" @=? runTexts ["2 4 * 3 /"]
-  , testCase "division by zero" $
-    Left DivisionByZero @=? runTexts ["4 2 2 - /"]
-  , testCase "dup" $ do
-    Right "1 1" @=? runTexts ["1 DUP"]
-    Right "1 2 2" @=? runTexts ["1 2 Dup"]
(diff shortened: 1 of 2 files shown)


exercises/forth/test/Tests.hs
```diff
@@ -1,74 +1,78 @@
...
+    let runTexts = fmap formatStack . foldM (flip evalText) empty
+
+    it "no input, no stack" $
+      formatStack empty `shouldBe` ""
+
+    it "numbers just get pushed onto the stack" $
+      runTexts ["1 2 3 4 5"] `shouldBe` Right "1 2 3 4 5"
+
+    it "non-word characters are separators" $
+      runTexts ["1\NUL2\SOH3\n4\r5 6\t7"] `shouldBe` Right "1 2 3 4 5 6 7"
```
petertseng: unrelated to this PR, but now I would like to figure out the origin of this test...

exercises/forth/test/Tests.hs
```diff
@@ -1,74 +1,78 @@
...
-runTexts xs = formatStack <$> foldM (flip evalText) empty xs
-
-forthTests :: [Test]
-forthTests =
-  [ testCase "no input, no stack" $
-    "" @=? formatStack empty
-  , testCase "numbers just get pushed onto the stack" $
-    Right "1 2 3 4 5" @=? runTexts ["1 2 3 4 5"]
-  , testCase "non-word characters are separators" $
-    -- Note the Ogham Space Mark ( ), this is a spacing character.
```
petertseng: If there is someone who does not read the README but only look at the tests, the lack of this comment might make things confusing. Do you think this is something to worry about? (Are there people who don't read the README?)

exercises/forth/test/Tests.hs
petertseng: found it https://github.com/exercism/exercism.io/pull/1188

exercises/forth/test/Tests.hs
rbasso: I think that, ideally, people should get all the information needed to solve an exercise in the following order:

- Running the tests - As a general rule this should be enough to know what went wrong.
- Reading the hints - This could help when people get stuck.
- Inspecting the test suite - Forcing people to read code should be a last resort solution, but I understand that some things are too complex to explain in English.

> If there is someone who does not read the README but only look at the tests, the lack of this comment might make things confusing.

It could happen but, if `HINTS.md` is the right place to have the hints, we should enforce that there are no hints in the test suites or in the stub solutions.

IMHO, the only comments that should be in the test suites are the ones related to the test's implementation, to assist us in maintaining the tests suites.

An alternative would be to improve the tests and descriptions, if you think they are not good enough.

But this is just my opinion. If you think it is important to keep the comment there, I can update the PR immediately.

petertseng: > I think that, ideally, people should get all the information needed to solve an exercise in the following order:

This seems to make sense; I think it represents how we would interact with a library we are considering to use - we either run the tests or read its documentation (some people switch the order of these two don't they) and only as the resort will they read the code.

When I am solving an exercise I tend to look quite often at the tests (so I know what's coming up next and I know what's expected of me), but I can't speak for everyone.

I think my main question is - now that this line appears in the README, when someone looks at that, what do you think their reaction will be? Will it be "I'll immediately search for all instances of this   character to see what this is all about", or will it be "Why are you telling me this? I will forget about it". Well, it's not my place to predict how people will act.

> An alternative would be to improve the tests and descriptions, if you think they are not good enough.

I'm not sure if "non-word characters are separators, including the Ogham Space Mark ( )" would be too much detail...

I question the case because visually the Ogham space mark is very easy to mistake for a minus sign, so someone looking at this might think this is five minus six, rather than five space six.

Perhaps I will take solace in the fact that most people probably don't fail this test because `isSpace` will do the right thing.

rbasso: >  think my main question is - now that this line appears in the README, when someone looks at that, what do you think their reaction will be? Will it be "I'll immediately search for all instances of this   character to see what this is all about", or will it be "Why are you telling me this? I will forget about it". Well, it's not my place to predict how people will act.

Maybe it doesn't belong in README...

After some thought, I think that we should simply remove the comment. It is too specific!

If it is part of testing non-word character, it doesn't deserve a comment more than NULL, SOH or tabs. If it is really a special case, it should have its own test.

What do you think?

petertseng: I imagine the only reason it was there in the first place is so that people don't confuse it with the minus sign.

If I ever write the canonical data for this problem, I plan not to include the Ogham space. But that is a question for another day.

As far as I can tell, it is not a special case at all. Just another whitespace character. It's not really adding anything to the test.

So what does this mean! I noticed recently that I often imagine problems but don't have evidence that others will experience the same problem (in this case, the problem is "people might mistake the Ogham space for a minus sign"). So if you think the comment should simply be removed, let's do it. Otherwise we won't nkow if others have the same problem.

rbasso: > If I ever write the canonical data for this problem, I plan not to include the Ogham space. But that is a question for another day.

I agree! It adds nothing!

> So if you think the comment should simply be removed, let's do it. Otherwise we won't know if others have the same problem.

Done! :smile:

petertseng:
Suggest that "- Move hint from test suite to `HINTS.md`." get removed from commit message

rbasso:
Sorry for that, I completely forgot to review the commit message. Fixed! :grimacing:

petertseng:
It may be time for me to write that canonical-data.json soon if I want to get rid of that Ogham space mark...

PR #393, opened 2016-10-10
anagram: Make the test suite as flexible as possible

### LINKED ISSUE #392: anagram: Change tests to be order-insensitive
A user [suggested](http://exercism.io/submissions/4e468e7ead3942d6af77c64149e18b61) that we could change `anagram`'s test suite to check the output list ignoring the word's order.

The way it is now, `anagramsFor word` must behave as `filter p` to pass the tests.

The patch would be really simple,

```bash
sed -i 's/shouldBe/shouldMatchList/g' test/Tests.hs
```

... but first we have to decide if it is important to check the output order.

Does anyone have an opinion on this?

rbasso:
For a while, we are doings things that seem to increase the number of solutions accepted by the test suites:

- In #144 we fixed some exercises to allow polymorphic solutions to compile without warnings.
- In #211, we rewrote most of the test suites. In the process, some of the tests were changed to allow more general numeric types.
- In #391, we changed `crypto-square` to stop testing for intermediary functions, and test only for the essential properties of the solution.
- In #392, we recently agreed that we should ignore the order of the words returned by `anagramsFor`, because the order *is not an essential part of the problem*.

All of those things where based on the same idea: **To allow more diverse solutions**.

This seems to be aligned to what was discussed in exercism/discussions#41 and exercism/discussions#44.

Taking this process to the next level, if the essence of the `anagram` problem is to find anagrams, there is no reason to restrict the solutions to work only on *Strings* and *Lists*. The following signatures seem equally reasonable:

```haskell
anagramsFor :: String ->          [String] ->          [String]
anagramsFor :: String -> Sequence  String  -> Sequence  String
anagramsFor :: String -> Set       String  -> Set       String
anagramsFor :: Text   ->          [Text]   ->          [Text]
anagramsFor :: Text   -> Sequence  Text    -> Sequence  Text
anagramsFor :: Text   -> Set       Text    -> Set       Text
```

This PR is an experiment on trying to write the most general possible test suite, allowing users to create beautiful solutions using important libraries that are usually not allowed in most of the exercises.

Closes #392.

### DIFF
--- a/exercises/anagram/test/Tests.hs
+++ b/exercises/anagram/test/Tests.hs
@@ -1,7 +1,8 @@
 {-# LANGUAGE RecordWildCards #-}
 
 import Data.Foldable     (for_)
-import Test.Hspec        (Spec, describe, it, shouldBe)
+import GHC.Exts          (fromList, toList)
+import Test.Hspec        (Spec, describe, it, shouldMatchList)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
 
 import Anagram (anagramsFor)
@@ -14,9 +15,14 @@ specs = describe "anagram" $
           describe "anagramsFor" $ for_ cases test
   where
 
-    test Case{..} = it description $ expression `shouldBe` expected
+    test Case{..} = it description $ expression `shouldMatchList` expected
       where
-        expression = anagramsFor subject candidates
+        expression = map toList
+                   . toList
+                   . anagramsFor (fromList subject)
+                   . fromList
+                   . map fromList
+                   $ candidates
 
 -- Test cases adapted from `exercism/x-common/anagrams.json` on 2016-07-25.
 
--- a/exercises/anagram/HINTS.md
+++ b/exercises/anagram/HINTS.md
@@ -1,0 +1,22 @@
+## Hints
+
+To complete this exercise you need to implement the function `anagramsFor`,
+that takes a *word* and a group of *words*, returning the ones that are
+anagrams of the given *word*.
+
+If it is your first time solving this exercise, it is recommended that you
+stick to the provided signature:
+
+```haskell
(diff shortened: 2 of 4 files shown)


exercises/anagram/test/Tests.hs
```diff
@@ -14,9 +15,14 @@ specs = describe "anagram" $
...
 
-    test Case{..} = it description $ expression `shouldBe` expected
+    test Case{..} = it description $ expression `shouldMatchList` expected
       where
-        expression = anagramsFor subject candidates
+        expression = map toList
+                   . toList
+                   . anagramsFor (fromList subject)
+                   . fromList
+                   . map fromList
```
petertseng: note for my study - this allows either `Text` or `String` to be used. Interesting.

petertseng:
Seems good! I hope the  many `fromList` and `toList` do not make it too hard for students to understand what is going on. That will depend on having their feedback so it seems like it is time to merge this.

exercises/anagram/test/Tests.hs
rbasso: [IsList](http://hackage.haskell.org/package/base-4.9.0.0/docs/GHC-Exts.html#t:IsList) is fantastic. Way more flexible than I imagined. :smile:

petertseng:
I *manually* verified that this works on also `String -> Set String -> Set String` and `String -> [String] -> [String]`. I kind of wish that validation could be performed *automatically*. For example, if we would have multiple example files, and our `travis.yml` would test each one of them in turn.

rbasso:
Hummm...this is a great idea, @petertseng.

I was thinking in something similar to solve the problem of having a separated `package.yaml` for the example. I think that maybe having multiple examples would automatically solve that.

I'll try to think in a way to test multiple solutions on Travis.

PR #396, opened 2016-10-12
travis: Support multiple example directories; anagram: Add multiple examples

### LINKED ISSUE #395: What if we could have multiple example files?
When reviewing #393 I realized I wish that we could test that multiple signatures would be accepted by the test suite. But of course we have to pick a certain signature in our example file... *or do we*?

What about a crazy idea - what if we supported having multiple example files?

This is already supported on x-api - any file that has *example* in its name will be excluded, so we could have `Example2.hs`, `Example3.hs`, etc.

Then we would have to make our `.travis.yml` support it in a reasonable way.

The idea is that Travis CI would test each example file individually (separately from the other examples) and makes sure that the test suite passes for each.

Potential use cases I can think of:

* Verify that the `anagram` test suite can accept multiple signatures.
* (If we also have a way to specify that a certain example *should fail the test suite*) We can test the quality of our test suite, making sure that it catches various cases we want it to catch. In `accumulate`, we have an inefficient example noted at https://github.com/exercism/xhaskell/blob/master/exercises/accumulate/src/Example.hs#L15 and we say that it should fail the test suite. But this was never automatically verified and indeed for the longest time that example did not even compile until #209 

**However**, I hope that this doesn't make our tests take too long to run.

Is this idea worth it? Should I try it out to see what happens?

petertseng:
Closes #395 (though I will plan to open a new issue to be able to mark an example as should-fail. Perhaps with a SHOULDFAIL in its directory name?)

Individual commits should explain well enough.


.travis.yml
```diff
@@ -50,13 +50,36 @@ script:
...
+        else
+            found=
+            for example in examples/*; do
+                echo "testing ${example}"
+                rm -f src/*.hs
+                mv ${example}/*.hs src
+                mv ${example}/package.yaml .
+                found=1
+
+                # ARGH, DUPLICATE CODE
```
petertseng: please help if you can help avoid this duplicate code. use a function?

`test_exercise () {
    stack test ${SET_RESOLVER} ...
}

and then just call `test_exercise` ?

.travis.yml
```diff
@@ -50,13 +50,36 @@ script:
...
+                stack test ${SET_RESOLVER} `# Select the correct resolver. `\
+                           --install-ghc   `# Download GHC if not in cache.`\
+                           --no-terminal   `# Terminal detection is broken.`\
+                           --pedantic      `# Enable -Wall and -Werror.    `
+            done
+
+            if [ -z "$found" ]; then
```
petertseng: kind of a clumsy way to do it, but I think it'll work. Any better?

exercises/anagram/examples/list-string/package.yaml
```diff
@@ -0,0 +1,17 @@
+name: anagram
```
petertseng: this file is pretty much just "the default package.yaml" - perhaps we should support the absence of this file, in which case the one from `exercises/anagram/package.yaml` is used? the only tricky bit is that it may be overwritten by an earlier example, so we need to back it up beforehand.

.travis.yml
rbasso: Don't worry with that for now. Let's just add the features and later we can simplify everything. :+1:

.travis.yml
rbasso: Not yet...but it is good enough the way it is. :smile:

.travis.yml
petertseng: The function worked fine so I am going to use it.

exercises/anagram/examples/list-string/package.yaml
petertseng: The advantage would be... fewer package.yaml to have to keep up to date if we have to change them all?

the disdavntage would be that the test code would get slightly more complex.

.travis.yml
```diff
@@ -50,13 +57,29 @@ script:
...
+                echo "testing ${example}"
+                rm -f src/*.hs
+                mv ${example}/*.hs src
+                mv ${example}/package.yaml .
+                found=1
+
+                test_exercise
+            done
+
+            if [ -z "$found" ]; then
```
petertseng: I think this does not work as intended. On my computer, at least, if `examples` dir is empty or does not exist, this is wha thappens:

```
testing examples/*/
mv: cannot stat 'examples/*//*.hs': No such file or directory
mv: cannot stat 'examples/*//package.yaml': No such file or directory
```

As you can see, since the glob found nothing, it is interpreted literally. And then `mv` fails, which I assume will abort the script as it should. But that means this `found` is unnecessary.

So, is ther ea point in keeping it?

rbasso: Found one link about it, but I couldn't test test it yet.
http://stackoverflow.com/questions/2937407/test-whether-a-glob-has-any-matches-in-bash

rbasso: Do you want to change anything else or can I merge now?

petertseng: I am satisfied!

PR #399, opened 2016-10-15
config.json: remove problems key

petertseng:
I don't like to have to maintain the order of the problems in two
places. So, `problems` is removed in favor of the new `exercises`.

The `exercises` key was added in #288, as part of the work in #274.

We have confirmation from https://github.com/exercism/x-api/pull/137
that it is safe to **delete** the `problems` key.


petertseng:
We are not the first track to do it: https://github.com/exercism/xcsharp/pull/130 and https://github.com/exercism/xfsharp/pull/224

PR #400, opened 2016-10-17
new exercise zebra-puzzle

abo64:
The `Tests.hs` are pretty degenerated as they consist of just one test. So they could be simplified, but I did not want to interfere with the usual format.
And of course the position of the exercise in `config.json` can be changed. This is just some more or less arbitrary guess of mine.
There is this [Logic programming example](https://wiki.haskell.org/Logic_programming_example). I tried to be somewhat close to this one with my `Example.hs`. So we could consider to add this link as a hint.

### DIFF
--- a/config.json
+++ b/config.json
@@ -251,6 +251,12 @@
       ]
     },
     {
+      "slug": "zebra-puzzle",
+      "difficulty": 1,
+      "topics": [
+      ]
+    },
+    {
       "slug": "minesweeper",
       "difficulty": 1,
       "topics": [
--- a/exercises/zebra-puzzle/package.yaml
+++ b/exercises/zebra-puzzle/package.yaml
@@ -1,0 +1,19 @@
+name: zebra-puzzle
+
+dependencies:
+  - base
+
+library:
+  exposed-modules: ZebraPuzzle
+  source-dirs: src
+  dependencies:
+  # - foo       # List here the packages you
+  # - bar       # want to use in your solution.
+
+tests:
+  test:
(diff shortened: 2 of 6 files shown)


config.json
```diff
@@ -251,6 +251,12 @@
       ]
     },
     {
+      "slug": "zebra-puzzle",
+      "difficulty": 1,
+      "topics": [
+      ]
+    },
+    {
```
rbasso: We probably need to do a major reordering of exercises, so that it is easier to figure out where to put a new one...

Considering that this exercise has an example solution with 113 lines, maybe it is a little hard to be be in this position. But I'm not sure because I didn't solve it.

abo64:
Changed `Tests.hs` accordingly.  
Please feel free to change the position of the exercise to any other more adequate place.

config.json
petertseng: Hmm, see, if we just look at `minesweeper`, it seems they are kind of similar difficulty, but I think `minesweeper` could stand to get moved to around where `change` is. This current placement is about 2/3 of the way through the track so I don't think we'll surprise anyone too badly, and we can look at the reordering later.

petertseng:
I haven't solved this in any language either, so we will have to save the discussion of elegant solutions for another day, it seems.

> will find a much more elegant and efficient solution

I imagine if a student wanted to be... "clever", the student could submit a solution that hard-codes the answer. But in that case "you're only cheating yourself!" It *is* efficient, I'm sure. Probably not many points on the elegance scale.

PR #402, opened 2016-10-19
config.json: Assign and order problems by difficulty

petertseng:
As part of the work in #274.

---

To discuss: Are all these difficulties right for us? I got lazy sometimes and looked at what difficulties F# track assigned them, since they are also functional and have all exercises except lens-person. there are some notable exceptions

* bank-account is significantly higher for us since we use IO monad for it
* our `robot-name` is significantly higher - we use IO monad
* our `nth-prime` may be lower since we can be lazy
* our accumulate/strain may be higher since we require laziness
* our `clock` is higher - it introduces a typeclass

Should we strictly order the exercises by difficulty, or is there sometimes a situation where we want to send a harder problem before an eaiser?

### DIFF
--- a/config.json
+++ b/config.json
@@ -11,19 +11,7 @@
       ]
     },
     {
-      "slug": "accumulate",
-      "difficulty": 1,
-      "topics": [
-      ]
-    },
-    {
-      "slug": "sublist",
-      "difficulty": 1,
-      "topics": [
-      ]
-    },
-    {
-      "slug": "grains",
+      "slug": "space-age",
       "difficulty": 1,
       "topics": [
       ]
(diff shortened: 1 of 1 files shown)


petertseng:
Increase difficulty of our robot-name. We use IO monad, and https://github.com/exercism/xhaskell/issues/193#issuecomment-232248821 indicates there is difficulty jump.

config.json
```diff
@@ -11,404 +11,404 @@
...
     },
     {
-      "slug": "accumulate",
+      "slug": "sum-of-multiples",
       "difficulty": 1,
       "topics": [
       ]
     },
     {
-      "slug": "sublist",
```
rbasso: I think it was too hard to be at the start, because people don't have yet a good understanding of recursing.

petertseng:
Maybe I should insert the topics in this PR too so we can easily justify the placements.

rbasso:
Feel free to leave the classification for later if you like. It's really a lot of work.

The dificulties and reordering are already a huge contributions. Thanks! :smile:

petertseng:
Another approach I can take, if this PR is too confusing, is to first make one PR to assign difficulties to the exercises (which we can discuss). And then after we have decided the difficulties, we can figure out individual reorderings we want to do. That is probably most useful if we don't think we will strictly order by difficulty.

rbasso:
Even if we decide not to order by difficulty, we can do it later.
There is really not reason to preserve the current order, because it is not good at all.

The only reason I can think to avoid reordering is if we intent to reorder everything again soon, so we two changes.

config.json
```diff
@@ -11,404 +11,404 @@
...
-      "difficulty": 1,
+      "slug": "phone-number",
+      "difficulty": 3,
       "topics": [
       ]
     },
     {
-      "slug": "list-ops",
-      "difficulty": 1,
+      "slug": "beer-song",
```
petertseng: really? Is beer song really easier than house? Of the three, I thought the difficulty is house < beer song < food chain.

config.json
```diff
@@ -11,404 +11,404 @@
...
-      "difficulty": 1,
+      "slug": "custom-set",
+      "difficulty": 5,
       "topics": [
       ]
     },
     {
-      "slug": "custom-set",
-      "difficulty": 1,
+      "slug": "simple-cipher",
```
petertseng: this does have an IO monad, because of the random aspect. Thus, should we make it even higher than a 5? Compare against bank-account and linked-list both are 6.

config.json
```diff
@@ -11,404 +11,404 @@
...
-      "difficulty": 1,
+      "slug": "robot-name",
+      "difficulty": 6,
       "topics": [
       ]
     },
     {
-      "slug": "pythagorean-triplet",
-      "difficulty": 1,
+      "slug": "bank-account",
```
petertseng: bank--account and robot-name (both IO monad) next to each other - is that OK, or should they be separate? is it good to have two consecutive to reinforce the concept, or apart so that people get a break?

petertseng:
> The only reason I can think to avoid reordering is if we intent to reorder everything again soon, so we two changes.

All right, I'll try my best to get something we like now, then, to reduce the chance of having to change it again later.

config.json
```diff
@@ -11,404 +11,404 @@
...
       "slug": "change",
-      "difficulty": 1,
+      "difficulty": 7,
       "topics": [
       ]
     },
     {
-      "slug": "connect",
-      "difficulty": 1,
+      "slug": "say",
```
petertseng: I think say can be decreased. maybe even be a 6. It's not that bad...

config.json
```diff
@@ -11,404 +11,404 @@
...
       "slug": "go-counting",
-      "difficulty": 1,
+      "difficulty": 9,
       "topics": [
       ]
     },
     {
-      "slug": "zipper",
-      "difficulty": 1,
+      "slug": "lens-person",
```
petertseng: lens-person 9 or 10? If you really want to understand the types, they can take a while to wrap your head around. But once you understand them the rest of the problem is not so hard.

config.json
rbasso: If I remember correctly, in `beer-song` the paragraph is just a function of a number. But the paragraphs have a lot of small variations.
In `house` we have a cumulative song, which is not that hard in Haskell, but the repeating pattern is a little more complex to get.

That said, I don't really care about the order, but it would be nice to put all thee far from each order, so that people don't get tired of the same type of exercise.

I think that `food-chain` is the hardest by far.

config.json
rbasso: Random is hard...specially if people don't know IO yet.

rbasso: Maybe we are missing an exercise to introduce IO without other complications...like random or mutable state.

config.json
rbasso: Maybe having at least one exercise between exercises of the same type is a good idea.

config.json
```diff
@@ -11,404 +11,404 @@
...
-      "difficulty": 1,
+      "slug": "bank-account",
+      "difficulty": 6,
       "topics": [
       ]
     },
     {
-      "slug": "saddle-points",
-      "difficulty": 1,
+      "slug": "queen-attack",
```
rbasso: Follow your instincts! :smile:  :+1:

config.json
rbasso: humm...I would go for a 9.

config.json
petertseng: I'll move it to 6.

> Maybe we are missing an exercise to introduce IO without other complications...like random or mutable state.

Hmm, any ideas on what we would do? Console interactoins seem simple but probably a bit hard to test. File operations a bit more complex.

config.json
petertseng: OK, moved them apart.

config.json
petertseng: I decided to move it to 5, then.

config.json
petertseng: Yeah, down to 6 with you, `say`.

petertseng:
Okay, I'm done with this round of changes.

config.json
rbasso: Maybe package [silently](https://www.stackage.org/nightly-2016-07-17/package/silently-1.2.5) can do it, but I didn't tested it.

If it works we can choose an interesting exercise to change and print the results. It would be a good introduction to the *IO monad*.

PR #403, opened 2016-10-19
word-count: Generalize test suite.

rbasso:
- Generalize the test suite to accept multiple signatures.
- Change the test suite to be case insensitive.
- Add `HINTS.md` explaining that the test suite is flexible.
- Add two example solutions: one very simple; the other with a custom type.
- Add the simplest type signature to the stub solution.
- Remove unneeded dependencies from `package.yaml`.
- Remove old example solution.

These changes tries to make this exercise, at the same time, more accessible and deeper. Beginners will probably be able to solve it easily, while more advanced students will be able the craft more elaborated solutions.

Also, we used `shouldMatchList` to improve error reporting and we normalize case before comparing, because the output's case *is not an essential part of the problem*.

This PR is very similar to #393, that changed `anagram` to be more flexible, and it is also compatible with previously working solutions.

### DIFF
--- a/exercises/word-count/HINTS.md
+++ b/exercises/word-count/HINTS.md
@@ -1,0 +1,21 @@
+## Hints
+
+To complete this exercise you need to implement the function `wordCount`,
+that takes a *text* and returns how many times each *word* appears.
+
+If it is your first time solving this exercise, it is recommended that you
+stick to the provided signature:
+
+```haskell
+wordCount :: String -> [(String, Int)]
+```
+
+Later, it may be a good idea to revisit this problem and play with other data
+types and libraries:
+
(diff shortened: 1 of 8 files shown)


petertseng:
Everything seems good - the only reason I'm not merging myself immediately is to ask this question - do you think we need to address the caching with multiple examples before merging this? This only affects the build time in a limited way - the exercises with multiple examples are always going to be recompiled, right? Or is it more?

rbasso:
I think it is safe to merge and enjoy the new functionality, and then we'll see how critical is to fix the caching. :smile:

petertseng:
OK We'll just take both of the multi-example changes and see what happen.

PR #404, opened 2016-10-20
leap: Add variable to stub solution's signature.

rbasso:
Considering that `leap` is the first exercise of the track, this small change may help a beginner, because (he|she) may think that the part before `=` should not be changed.

More experienced students will certainly already know that they can eta-reduce the functions, so I think there is no harm.

To make the exercises more accessible to newcomers, I believe it is in general a good think to have a stub solution that:

- Includes the simplest type signature acceptable by the test suite.
- Includes a [eta-abstracted](https://wiki.haskell.org/Eta_conversion) fake implementation with reasonable variable names. I have been using `undefined` as the function body.
- Compiles and fails the tests, if possible without ruining the challenge.

What do you think about it, @petertseng?

### DIFF
--- a/exercises/leap/src/LeapYear.hs
+++ b/exercises/leap/src/LeapYear.hs
@@ -1,4 +1,4 @@
 module LeapYear (isLeapYear) where
 
 isLeapYear :: Integer -> Bool
-isLeapYear = undefined
+isLeapYear year = undefined


petertseng:
I think these are good ideas. I had hoped the meaning of the argument would be self-evident, but the note about beginners not thinking to change the part before `=` seems important. Did you observe this difficulty? Perhaps did someone submit a solution of the form `leapYear = \y -> ...` ?

rbasso:
> Did you observe this difficulty?

Not really!

It's just that , now that you reordered the exercises, I'm  trying to lower the bar a little and make everything clear and easy in the first ones, to avoid anyone giving up too soon. Also, sometimes the variable name helps documenting the functions to be implemented.

PR #405, opened 2016-10-21
space-age: Generalize tests and add instructions.

rbasso:
- Generalize the test suite to accept more diverse type signatures.
- Add `HINTS.md` explaining how to solve the exercise.
- Add a simple type signature to the stub solutions.
- Add variable names to the stub's incomplete implementation.
- Move example solution to `examples/success-double`.
- Add new example solution using `Rational`.

The idea here is mostly the same from #404, so I won't elaborate further.


petertseng:
Let me see if I can do the cache thing now

Issue #397, opened 2016-10-12
Allow marking an example as should-fail


petertseng:
As an extension to multiple examples in #395, we'd like to be able to add examples that we can mark as expected to fail the test.

We can use this to make sure that our test cases properly reject some mistaken solutions that we expect them to reject. The example I gave earlier: In `accumulate`, we have an inefficient example noted at https://github.com/exercism/xhaskell/blob/master/exercises/accumulate/src/Example.hs#L15 and we say that it should fail the test suite. But this was never automatically verified and indeed for the longest time that example did not even compile until #209 

The things to figure out here are:
- How will it be specified? e.g. the name of the directory contains `SHOULDFAIL`? Or something else?
- After it is specified, ideas on implementation? Something to do with checking the exit status of `stack test`, of course. We'll have to keep in mind that we usually have `set -e` on, but this time we have to invert the exit code of `stack test`. Maybe something can do that for us. Maybe simply the `!` operator will work.

rbasso:
First, I think we should clarify what it is to fail. I think that we have 3 distinct cases:

- Everything compiles and the example solution passes the tests. (success)
- Everything complies and the example solution fails the tests. (fail)
- The pair (example solution, test suite) fails compilation. (error)

If I was to choose how to mark the expected results, I would go for folders named like this:

- `examples/fail-something/`
- `examples/error-something/`
- `examples/success-something/`

Any other prefix or the lack of a sub-folder should make Travis fail the tests.

To implement this, it would be necessary to separate *building* from *testing*.

Another think that may be a problem is that, because we are substituting files in the same project, the cache for the examples seems to be always overwritten. Maybe we'll have to cache separately each of the examples to keep recompilation times reasonable.

All this will probably make the `.travis.yml` considerably more complex.

petertseng:
The idea seems good to me.

> Another think that may be a problem is that, because we are substituting files in the same project, the cache for the examples seems to be always overwritten.

Ah, I see. Yes, that might be needed. It appears I missed that a cached folder is being used as `stack-work` for each exercise. Now it looks like it indeed should be for each example.

> All this will probably make the `.travis.yml` considerably more complex.

Sometimes I am told it is not wise to write too-complex logic in shell scripts and it makes me wonder if we should go back to having something like https://github.com/exercism/xhaskell/commits/master/_test/check-exercises.hs . Sure, the complexity would still be somewhere (we only change where and what language it is), but maybe it is more desirable to keep the complexity there.

[...]

PR #406, opened 2016-10-21
travis: Cache examples separately

petertseng:
PR #396 gives us the power to have multiple examples. But if the
examples have different dependencies, their caches clobber each other.
Thus, it seems necessary to have a separate cache per example.

There is some discussion on this and some proposed solutions in #397.

We may replace this shell script soon, but it will at least let us know
the strategy we want to use and validate the directory layout.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,12 +52,12 @@ script:
     for exercise in ${TRAVIS_BUILD_DIR}/exercises/* ; do
         pushd ${exercise}
 
-        # `stack --work-dir` fails if not targeting a subdirectory, so we just
-        # symbolic-link `.stack-work` to a subfolder in the cache directory.
-        mkdir -p "${HOME}/.foldercache/${exercise}/.stack-work"
-        ln -f -s "${HOME}/.foldercache/${exercise}/.stack-work"
-
         if [ -f src/Example.hs ]; then
+            # `stack --work-dir` fails if not targeting a subdirectory, so we just
+            # symbolic-link `.stack-work` to a subfolder in the cache directory.
+            mkdir -p "${HOME}/.foldercache/${exercise}/.stack-work"
+            ln -f -s "${HOME}/.foldercache/${exercise}/.stack-work"
+
             # Here we prepare the exercise to be tested by Stack.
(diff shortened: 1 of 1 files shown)


petertseng:
> * Move the example to `./src` and `package.yaml` to `./`.
> * Create a new folder (possibly in the user's home folder) for each pair (exercise/example) and copy everything needed there before testing.

> I think that the second solution is better because it would avoid any present or future problem with the global stack cache.
> I made a few tests and seems that the first solution works too, but sharing exactly the same path for distinct build feels wrong...

Ah. To be clear, I have nothing against the second solution, it was just that the first was the path of least resistance currently

.travis.yml
```diff
@@ -68,6 +68,10 @@ script:
             exit 1
         else
             for example in examples/*/ ; do
+                examplename=$(basename example)
```
petertseng: nope, needs to be `$example`.

petertseng:
Okay, travis seems to have fine paths here.

/home/travis/.foldercache/home/travis/build/exercism/xhaskell/exercises/anagram/list-string/.stack-work/dist/x86_64-linux/Cabal-1.24.0.0/build/Anagram.dyn_hi

/home/travis/.foldercache/home/travis/build/exercism/xhaskell/exercises/anagram/list-string/.stack-work/dist/x86_64-linux/Cabal-1.24.0.0/build/Anagram.dyn_o

Unrelated, but if we would use `basename $exercise` we could change `/home/travis/.foldercache/home/travis/build/exercism/xhaskell/exercises/anagram` to `/home/travis/.foldercache/anagram` which would be a shorter path... but perhaps it does not matter

rbasso:
`basename` seems a nice idea.

The only thing missing now is to move the `package.yaml`, right?

petertseng:
Hold on, I thought the `package.yaml` gets moved already at https://github.com/exercism/xhaskell/blob/master/.travis.yml#L74

Or are you talking about something different?

If the package.yaml didn't get moved in the above fashion, I would be surprised that some of the examples even build...

petertseng:
OK, I'm going to create another issue for the basename thing since I might not get to it just yet.

PR #410, opened 2016-10-31
new exercise alphametics

abo64:
of course feel free to change the place of the exercise in `config.json`. First I thought it makes sense to have it after `sgf-parsing` but then I saw its difficulty of 9 and reconsidered.
I kept the tests for multiplication and exponentiation from a former version of `canconical-data.json` as they make the exercise more interesting.
what annoys me is the more than 30 times worse performance than my same (I think) Scala implementation. would be very much interested to know the reason why.

### DIFF
--- a/exercises/alphametics/HINTS.md
+++ b/exercises/alphametics/HINTS.md
@@ -1,0 +1,5 @@
+## Hints
+
+You can solve this exercise with a brute force algorithm, but this will possibly have a poor runtime performance.
+Try to find a more sophisticated solution. Hint: You could try the column-wise addition algorithm that is usually taught in high school.
+
--- a/exercises/alphametics/package.yaml
+++ b/exercises/alphametics/package.yaml
@@ -1,0 +1,20 @@
+name: alphametics
+
+dependencies:
+  - base
+
+library:
+  exposed-modules: Alphametics
+  source-dirs: src
+  dependencies:
+  # - foo       # List here the packages you
+  # - bar       # want to use in your solution.
+    - parsec
+
+tests:
+  test:
+    main: Tests.hs
+    source-dirs: test
+    dependencies:
+      - alphametics
+      - hspec
--- a/exercises/alphametics/src/Alphametics.hs
+++ b/exercises/alphametics/src/Alphametics.hs
@@ -1,0 +1,4 @@
+module Alphametics (solve) where
+
+solve :: String -> Maybe [(Char, Int)]
+solve puzzle = undefined
--- a/exercises/alphametics/test/Tests.hs
+++ b/exercises/alphametics/test/Tests.hs
@@ -1,0 +1,78 @@
+{-# LANGUAGE RecordWildCards #-}
+
+import Data.Foldable     (for_)
+import Data.Function     (on)
+import Data.List         (sort)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
+import Alphametics (solve)
+
+main :: IO ()
+main = hspecWith defaultConfig {configFastFail = True} specs
+
+specs :: Spec
+specs = describe "alphametics" $
+          describe "solve" $ for_ cases test
+  where
+
+    test Case{..} = it description assertion
+      where
+        shouldMatchSolution = shouldBe `on` fmap sort
+        assertion = solve puzzle `shouldMatchSolution` expected
+
+-- Test cases adapted from `exercism/x-common/exercises/alphametics/canonical-data.json` on 2016-10-26.
+
+data Case = Case { description :: String
+                 , puzzle      :: String
+                 , expected    :: Maybe[(Char, Int)]
+                 }
+
+cases :: [Case]
+cases = [ Case { description = "puzzle with three letters"
+               , puzzle      = "I + BB == ILL"
+               , expected    = Just [('I', 1), ('B', 9), ('L', 0)]
+               }
+        , Case { description = "solution must have unique value for each letter"
+               , puzzle      = "A == B"
+               , expected    = Nothing
+               }
+        , Case { description = "leading zero solution is invalid"
+               , puzzle      = "ACA + DD == BD"
+               , expected    = Nothing
+               }
+        , Case { description = "puzzle with four letters"
+               , puzzle      = "AS + A == MOM"
+               , expected    = Just [('A', 9), ('S', 2), ('M', 1), ('O', 0)]
+               }
+        , Case { description = "puzzle with six letters"
+               , puzzle      = "NO + NO + TOO == LATE"
+               , expected    = Just [('N', 7), ('O', 4), ('T', 9), ('L', 1),
+                                      ('A', 0), ('E', 2)]
+               }
+        , Case { description = "puzzle with seven letters"
+               , puzzle      = "HE + SEES + THE == LIGHT"
+               , expected    = Just [('E', 4), ('G', 2), ('H', 5), ('I', 0),
+                                      ('L', 1), ('S', 9), ('T', 7)]
+               }
+        , Case { description = "puzzle with eight letters"
+               , puzzle      = "SEND + MORE == MONEY"
+               , expected    = Just [('S', 9), ('E', 5), ('N', 6), ('D', 7), ('M', 1),
+                                      ('O', 0), ('R', 8), ('Y', 2)]
+               }
+        -- some more (and more demanding) tests
+--        , Case { description  = "puzzle with ten letters"
+--               , puzzle       = "AND + A + STRONG + OFFENSE + AS + A + GOOD == DEFENSE"
+--               , expected     = Just [('A', 5), ('D', 3), ('E', 4), ('F', 7), ('G', 8),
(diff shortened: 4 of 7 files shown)


exercises/alphametics/test/Tests.hs
```diff
@@ -0,0 +1,76 @@
...
+      where
+        assertion  = solution `shouldBe` sort <$> expected
+        solution = sort <$> solve puzzle
+
+-- Test cases adapted from `exercism/x-common/exercises/alphametics/acanonical-data.json` on 2016-10-26.
+
+data Case = Case { description  ::        String
+                 , puzzle       ::        String
+                 , expected     ::        Maybe[(Char, Int)]
+                 }
```
rbasso: ##### Cosmetic suggestion

A lot of unneeded spaces. One missing space after `Maybe`.

I think that this would be more readable:
```haskell
data Case = Case { description :: String
                 , puzzle      :: String
                 , expected    :: Maybe [(Char, Int)]
                 }
```

exercises/alphametics/test/Tests.hs
```diff
@@ -0,0 +1,76 @@
...
+
+cases :: [Case]
+cases = [ Case { description  = "puzzle with three letters"
+               , puzzle       = "I + BB == ILL"
+               , expected     = Just [('I', 1), ('B', 9), ('L', 0)]
+               }
```
rbasso: ##### Cosmetic suggestion

I would remove one space before each `=` here, aligning everything with `description =`

exercises/alphametics/test/Tests.hs
```diff
@@ -0,0 +1,76 @@
...
+                                      ('O', 0), ('R', 8), ('Y', 2)]
+               }
+        , Case { description  = "puzzle with ten letters"
+               , puzzle       = "AND + A + STRONG + OFFENSE + AS + A + GOOD == DEFENSE"
+               , expected     = Just [('A', 5), ('D', 3), ('E', 4), ('F', 7), ('G', 8),
+                                      ('N', 0), ('O', 2), ('R', 1), ('S', 6), ('T', 9)]
+               }
+        , Case { description  = "puzzle with multiplication"
+               , puzzle       = "IF * DR == DORI"
+               , expected     = Just [('I', 8), ('F', 2), ('D', 3), ('R', 9), ('O', 1)]
```
rbasso: These two last exercises with `*` and `^` seem to make the exercise significantly harder, forcing people to learn a lot about parsing before even starting to solve the actual puzzle.

Most of the users [give up on the track](http://www.exercism.io/stats/haskell) before reaching the final exercises, so I think we should focus on making the exercises more pedagogical to the less advanced students

I found that this was already discussed in https://github.com/exercism/x-common/issues/395, so I think that in this case we should follow `x-common` and remove them.

What do you think about it, @petertseng ?

abo64: thx @rbasso  I will try to implement all your helpful suggestions.
I agree that the last two tests make the whole exercise harder, and we could leave them out. Or perhaps make them somehow optional by commenting them out/marking them as pending so that each student can decide for himself?
Main problem is of course the poor performance. And I still wonder why. As mentioned above it works fine in Scala. Only explanation so far I could come up with is that something causes most or all `permutations` to become evaluated. In Scala `permutations` returns an Iterator, so this might make sure that the whole computation remains lazy? I will try again to find the reason - this is an interesting problem in itself. And of course - any ideas/help appreciated.

exercises/alphametics/src/Alphametics.hs
```diff
@@ -0,0 +1,4 @@
+module Alphametics (solve) where
+
+solve :: String -> Maybe [(Char, Int)]
+solve puzzle = undefined
```
abo64: I thought so, too. One can implement a solution using a `Map`, say, and then easily transform to this basic type.
About including `puzzle` I considered
```haskell
type Puzzle = String

solve :: Puzzle -> Maybe [(Char, Int)]
solve = undefined
```
but decided against it as this coding style with `type`s might not be to everybody's liking.

exercises/alphametics/test/Tests.hs
abo64: Fine then, I will remove them.
That means I will go back to the drawing board for the example solution as the current one indeed is brute force using `permutations` (and has this lousy performance in comparison to Scala; although I still intend to find the reason why, but later then).

petertseng: It could still be the case that many students will still submit the simple brute force solution anyway. It's the first thing that comes to mind. Different people want to see different things from Exercism. So limiting to addition problems opens the door to doing something more considered, for those who want to do that. The decreased difficult does help a bit too, I think.

I realize that another way to expanding this problem is indeed to keep the operations other than addition. There are various directions to take this problem. I chose one way (and it happens to be the way that x-common chose too), but maybe others like a different way.

As for what we can do about this long runtime, maybe we can get some ideas from other tracks. https://github.com/exercism/xruby/pull/455/files#diff-8ec059eaf4c70572282d3173a6325c3aR53 the Ruby track chose to comment out the cases that had many letters since they do take a long time to solve that way. Recently as of https://github.com/exercism/xruby/pull/469 the Ruby example is fast enough to solve the equations with many letters, but they still have a comment telling students it's OK not to do them: https://github.com/exercism/xruby/pull/469/files#diff-8ec059eaf4c70572282d3173a6325c3aR53 . If we decide here that the same ideas are appropriate, we could apply them.

exercises/alphametics/package.yaml
```diff
@@ -0,0 +1,20 @@
...
+dependencies:
+  - base
+
+library:
+  exposed-modules: Alphametics
+  source-dirs: src
+  dependencies:
+  # - foo       # List here the packages you
+  # - bar       # want to use in your solution.
+    - parsec
```
abo64: `word-count` looks good. at least some `HINTS.md` might be a good compromise.

exercises/alphametics/test/Tests.hs
petertseng: > I'll admit I'm not 100% sure either, since there's a lot going on.

But I can surely explain what I would do, if you are interested

abo64: yeah, starting from the rightmost digit/column, doing the sum and then accumulating restrictions going left.
I thought that this might even work for multiplication and exponentiation.

petertseng: I agree it can work with multiplication - we would have to consider more columns at a time, since in:

```
 AB
xCD
---
EFG

both A*D and B*C contribute to F.

But once that detail is ironed out, it would still be doable. I have not tested it out myself to make sure, but seems that way, anyway.

Exponentiation would thus also be doable if we just treat it as repeated multiplication, assuming there are no variable exponents.

**Comment by abo64** [comment]
I have implemented this column-wise partial solution algorithm. So far it only works for addition and only up to eight letter words. Thus I moved the ten letter test as well as the tests for multiplication and exponentiation to `HINTS.md`.

File: exercises/alphametics/HINTS.md [hunk 6]
```diff
@@ -0,0 +1,25 @@
...
+```haskell
+        , Case { description  = "puzzle with multiplication"
+               , puzzle       = "IF * DR == DORI"
+               , expected     = Just [('I', 8), ('F', 2), ('D', 3), ('R', 9), ('O', 1)]
+               }
+        , Case { description  = "puzzle with exponents"
+               , puzzle       = "PI * R ^ 2 == AREA"
+               , expected     = Just [('P', 9), ('I', 6), ('R', 7), ('A', 4), ('E', 0)]
+               }
+```
```
rbasso: I feel that these cases pollute the `HINTS.md` file. Also, I don't believe anyone will copy/paste this in `test/Test.hs`.

I prefer them commented out in the test suite, but I will respect your decision if you prefer them here, @abo64.

abo64:
I moved the tests back to `Tests.hs`. Shall we then keep `HINTS.md`?
And I am afraid I have done some mistake while squashing? :-(

rbasso:
You can use `git rebase -i master` and delete the line with the unwanted commit to fix that.

abo64:
If it is too messed up: I could create a new PR with just the final changes?

rbasso:
What a mess, man! :smile: 

Don't worry!

I'm not good at *git*, but I think that a I have a solution for this problem:

- Type `git rebase --abort`, just to be sure that there is nothing ongoin.
- Check that you are in the branch named `exercise-alphametics`.
- Type `git rebase -i 82ae06`
- Remove lines until it matches **exactly** what is below:

```git
pick 7c98beb new exercise alphametics                                                                           
squash 5d75fb0 fix hlint complaint                                              
squash 39cc180 example with column-wise addition algorithm, removed problematic tests
squash dcb567e moved additional tests back to Tests.hs                          
``` 
- Edit the commit message to something like:

```
alphametics: add new exercise
```

- Push the changes to your fork: `git push -f`

abo64:
thanks, man! :-)

rbasso:
Now you probably will want to fix your `master` branch. I think that something like this will work:

```
$ git checkout master
$ git reset --hard upstream/master
$ git pull upstream master
$ git push -f origin master

abo64:
done.

rbasso:
Thanks, @abo64! I know that this one gave you a lot of work! :+1:

PR #412, opened 2016-11-06
forth: Expect a list rather than a string or Text

petertseng:
Like in https://github.com/exercism/x-common/pull/394, I request that we
question our assumptions. Why have we requested a string? Well, it was
like that in https://github.com/exercism/exercism.io/pull/1188, the
first appearance of this exercise. But we can just use a list, right?


exercises/forth/src/Forth.hs
```diff
@@ -23,5 +23,5 @@ empty = undefined
 evalText :: Text -> ForthState -> Either ForthError ForthState
 evalText = undefined
 
-formatStack :: ForthState -> Text
-formatStack = undefined
+toList :: ForthState -> Text
```
rbasso: I think that you forgot to change this line to `[Int]`.

petertseng:
> provide a clean `package.yaml`, without a dependency on `containers`.

Sure! I will take the opportunity to do so.

> I think that you forgot to change this line to `[Int]`.

Oh dear, maybe I need to speed up the process of doing the "should-compile-but-fail-tests" and we should throw every stub solution into that category. Would prevent this sort of mistake.

petertseng:
okay, example is moved - note that I didn't know what to put after `success-` so I put... `success-standard`. Maybe just `success-example`? Or maybe simply `success` will do?

Also, maybe I should take a moment to do this for all package.yaml that have extra packages...

exercises/forth/examples/success-standard/package.yaml
```diff
@@ -0,0 +1,19 @@
...
+dependencies:
+  - base
+  - text
+
+library:
+  exposed-modules: Forth
+  source-dirs: src
+  dependencies:
+    - containers
+
```
rbasso: Great...you even remembered to remove the comments. :)

petertseng:
OK, test passed which is what I was waiting for. If we have no better suggestion than `success-standard` then I'll merge this one.

rbasso:
`success-standard` is good, don't worry! :+1:

PR #413, opened 2016-11-06
forth: Add new tests from x-common

petertseng:
First commit is same as #412, the rest should bring us to match https://github.com/exercism/x-common/pull/394

### DIFF
--- a/exercises/forth/test/Tests.hs
+++ b/exercises/forth/test/Tests.hs
@@ -17,62 +17,82 @@ specs = describe "forth" $ do
 
     let runTexts = fmap toList . foldM (flip evalText) empty
 
-    it "no input, no stack" $
-      toList empty `shouldBe` []
+    describe "parsing and numbers" $ do
+      it "empty input results in empty stack" $
+        toList empty `shouldBe` []
 
-    it "numbers just get pushed onto the stack" $
-      runTexts ["1 2 3 4 5"] `shouldBe` Right [1, 2, 3, 4, 5]
+      it "numbers just get pushed onto the stack" $
+        runTexts ["1 2 3 4 5"] `shouldBe` Right [1, 2, 3, 4, 5]
 
-    it "non-word characters are separators" $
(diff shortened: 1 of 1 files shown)


exercises/forth/test/Tests.hs
```diff
@@ -17,62 +17,82 @@ specs = describe "forth" $ do
...
+    describe "swap" $ do
+      it "swaps the top two values on the stack if they are the only ones" $
+        runTexts ["1 2 swap"  ] `shouldBe` Right [2, 1]
+      it "swaps the top two values on the stack if they are not the only ones" $
+        runTexts ["1 2 3 swap"] `shouldBe` Right [1, 3, 2]
+      it "errors if there is nothing on the stack" $
+        runTexts ["swap"      ] `shouldBe` Left StackUnderflow
+      it "errors if there is only one value on the stack" $
+        runTexts ["1 swap"    ] `shouldBe` Left StackUnderflow
+
```
rbasso: I see that here you swapped to order of those two test. I like the new ordering more. :+1:

petertseng: Oh, I was inconsistent. `1 over` is before `over`. I'll make it consistent, and note an order change in commit message.

exercises/forth/test/Tests.hs
```diff
@@ -17,62 +17,82 @@ specs = describe "forth" $ do
...
+    describe "over" $ do
+      it "copies the second element if there are only two" $
+        runTexts ["1 2 over"  ] `shouldBe` Right [1, 2, 1]
+      it "copies the second element if there are more than two" $
+        runTexts ["1 2 3 over"] `shouldBe` Right [1, 2, 3, 2]
+      it "errors if there is nothing on the stack" $
+        runTexts ["1 over"    ] `shouldBe` Left StackUnderflow
```
petertseng: actually it's even worse - the test message doesn't match the actual test!!!

petertseng:
Everything should be OK now - the only change is to rebase and also to fix the ordering of the `over` tests. I'll take a final look.

Issue #416, opened 2016-11-06
Ensuring we are up to date with x-common


petertseng:
How should we ensure we're up to date?

One idea is: Every time someone makes a change in x-common, that person is expected to file an issue against all tracks that implement the exercise. If we **rely** on this to happen, then we need take no further action.

Sometimes, that is not always adhered to. We mark dates in our test files, what if we used that to an advantage?

I wrote a short script in https://github.com/exercism/x-common/compare/master...petertseng:up-to-date . Sorry it is not in Haskell, but the idea should be clear:

* read out the date from our test file
* ask x-common if that canonical-data.json file has been changed since that date

In the short term: I will be making some PRs today based on what I've found.
In the long term: Which route shall we take?

rbasso:
The best solution I could find until now is to add the data to `config.json`:

```haskell
  "exercises": [
    {
      "slug": "leap",
      "difficulty": 1,
      "topics": [
      ],
      "x-common-file": "exercises/leap/canonical-data.json",
      "x-common-hash": "cda8f9800a33d997f8c6146a10b8caf66e25ec4b"
    },
    ...
 ```

With this information there, it would be easier to match against x-common:

```bash
$ git diff cda8f9800a33d997f8c6146a10b8caf66e25ec4b exercises/leap/canonical-data.json
```

The only thing we would have to keep in the test suite are the deviations from x-common. 

What do you think about it?

rbasso:
As an example, this shamefull bash script gets the filenames and hashes from `config.json` and checks which files changed in the repository:

#### Beware: Ugly script ahead!

###### checkChanges
```bash
#!/usr/bin/env bash


QUERY='.exercises|map(select(has("x-common-file")))|map({"x-common-hash","x-common-file"})|map("git diff --name-only " + ."x-common-hash" + " " + ."x-common-file" + ";") | .[]';

eval $(jq -r "$QUERY" "$@")
```

It must be run from the root of `x-common`, passing the `config.json` file as an argument:

```bash
.../x-common/ $
.../x-common/ $ checkChanges ../xhaskell/config.json
exercises/leap/canonical-data.json
exercises/space-age/canonical-data.json
...
```

petertseng:
It seems good to use a sha instead of a date - it allows us to check precisely whether the file has changed. That is good.

It may not be necessary to specify the file directly, since they are all `"exercises/" ++ slug ++ "/canonical-data.json"`. I would say it only needs to be specified if it breaks the pattern.

[...]

PR #417, opened 2016-11-06
grains: Declare compliance with x-common

petertseng:
Added in https://github.com/exercism/x-common/pull/420

We already match all cases.


petertseng:
If you ask "why only change the comment if no test content has changed?" Then I say, I'm reading in the date in a script (as in #416), so I simply need a signal that we comply with the new canonical-data.json that was added since that comment was added.

PR #419, opened 2016-11-06
word-count: Add more tests with punctuation

petertseng:
Added in https://github.com/exercism/x-common/pull/403

---

discussion before merging:

These new tests seem to raise difficulty slightly because now we need to be sensitive to the difference between `'` as an apostrophe versus used as single quotes. I expect the example solution to need fixing, and I might bump the difficulty to 5 based on what I see.

### DIFF
--- a/config.json
+++ b/config.json
@@ -149,12 +149,6 @@
       ]
     },
     {
-      "slug": "word-count",
-      "difficulty": 4,
-      "topics": [
-      ]
-    },
-    {
       "slug": "meetup",
       "difficulty": 4,
       "topics": [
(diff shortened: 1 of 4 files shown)


petertseng:
Right, `wordsBy` isn't going to work now, since we now need to know the characters surrounding a `'`.

All right, I'm probably not going to have time to work on this soon, so if nobody has an idea to deal with the new tests maybe it's better to create an issue

petertseng:
An idea I will work on soon: `wordsBy (\c -> isAlphaNum c || c == '\'')`, then strip all leading and trailing `'` from the words.

petertseng:
Okay, here's something that seems to work OK. I really think I should bump difficulty to 5 for this, so I did it.

PR #422, opened 2016-11-07
Reorganize example solutions

### LINKED ISSUE #398: readme: Fix instructions for running the tests.
After #395 and resulting PR #396, some exercises will have multiple examples. We should add instructions for that in the README once we finalise how that will look.

We may want to defer this until after we decide how to deal with #397.

rbasso:
- Move examples to `examples/success-something/ModuleName.hs`
- Copy `package.yaml`files to all example solutions.
- Remove comments and empty dependencies from the examples' `package.yaml` files.
- Fix Travis to find the example in the new location.
- Remove Travis support for examples in the source directory.
- Update documentation to reflect the new directory structure.

Closes #398.


petertseng:
I like where this is going. I probably won't check every single individual file but I would think that everything is in order because Travis still passes after requiring `examples` directory.

So now I have a question about how we want to structure these. Current example structure:

```
examples/success-standard/Accumulate.hs
examples/success-standard/package.yaml
```

what if instead it became like this:

```
examples/success-standard/src/Accumulate.hs
examples/success-standard/package.yaml
```

Why does it matter? It may make testing each example easier - we'd probably only need to symlink in the `stack.yaml`, `.cabal` file, and the `test` directory, instead of moving the files in to overwrite `src` and `package.yaml` every time.

This change doesn't necessarily need to be made in this PR, but I felt that if we are moving a bunch of around, might be a good suggestion to make now so we don't move them around as many times.

rbasso:
I'm updating the exercise to your proposed structure, @petertseng. Right now I messed everything, so please don't merge it.

rbasso:
I think it is ready now! :smile:

petertseng:
Very nice. I'll explore later whether we want to have Travis symlink files in, instead of moving files out

petertseng:
In my mind, the benefit would be that it leaves the stub solution intact... which we can then test for compilation.

rbasso:
I think I tried something with symbolic links but it didn't work for some reason I don't remember. Good luck!

Issue #421, opened 2016-11-07
We should probably enforce in Travis that our stub solutions compile


petertseng:
In #412 I made a type change to the tests + example solution but forgot to change the stub solution.

In #420 we discuss that we should keep packages in `package.yaml` if the stub solution needs them.

It would be good if we had a way to automatically check this. I think this means we should probably check that the stub solution compiles in Travis.

Are there any stub solutions that we expect not to compile? any other concerns with this approach? (It will increase the time taken, probably, but perhaps it is worth it. We will see when we do it)

rbasso:
I think that, ideally, the stub solution would compile and fail at the first test, but some problems have incomplete stub solutions (maybe because implementing a data type is part of the exercise), and cannot compile.

We could try to fix those exercises using incomplete data types, but I'm not sure if that's possible in all cases.

What we are facing in this issue is a fundamental incompatibility between *test-driven development* and "statically-typed languages". There is no way to run the tests before everything type-checks.

We had a lot of work trying to make test suites that accept multiple types to mitigate the fact that, in Haskell, the types usually come before writing the tests.

If we can "fix" each exercise to compile without ruining the challenge, then I think it would be **great** to test the stub solutions in Travis.

rbasso:
One way to improve in this direction is to first move all exercises to the new proposed format. Removing the examples from the `src` folder, it would be even easier to test (locally or in Travis) if the stub solution compiles.

petertseng:
All right, now that that's done, let's try to think about how it might be done. We'll take `allergies` exercise as an example, since that is the alphabetically first that has the problem.

The stub solution gives:

```haskell
module Allergies (Allergen(..), allergies, isAllergicTo) where

allergies :: Int -> [Allergen]
allergies = undefined

isAllergicTo :: Allergen -> Int -> Bool
isAllergicTo = undefined
```

And the problem is simply that `Allergen` isn't there.

So what can we do about this?

* Put in the stub solution something like `data Allergen = Dummy`
  * It is a very simple solution, and doesn't give away anything.
  * Potential risk: What if a student doesn't realize it is allowed to change `data` to `type` or `newtype` and suddenly we see less variety in solutions because everyone blindly uses the choice we handed to them?
* For now, we add it to a list of stub solutions that we expect to fail to build
  * Can easily be done now without taking any further decision
  * ... but we probably have to make a decision later.
[...]

PR #425, opened 2016-11-08
custom-set: change stub to make it compile.

rbasso:
This PR makes small changes to allow the stub solution to build and also compile the test suite.

The ability to run the tests with the stub solution is useful mainly for two reasons:

- It allows us to assert that the stub solution is "correct".
- It allows the students to have an easier start, also making it clear what needs to be implemented.

On the other hand, it has drawbacks:

- We are forced to create a dummy data definition, which makes the exercise a little less challenging.
- We cannot export the data constructors to the test suite, because they do not exist yet.

Considering that most of the students don't go too far in the track, I'm not too worried about the risk of making it too easy, so I think it is worthy to try this path.

Related to #421.

### DIFF
--- a/exercises/custom-set/HINTS.md
+++ b/exercises/custom-set/HINTS.md
@@ -17,8 +17,9 @@ with `Eq` and `Show` instances, and implement the following functions:
 - `toList`
 - `union`
 
-You will find the type signatures already in place, but it is up to you
-to define the functions.
+You will find a dummy data declaration and type signatures already in place,
+but it is up to you to define the functions and create a meaningful data type,
+newtype or type synonym.
 
 If you're interested in writing an efficient implementation but don't quite
 know where to start, the best primer I know of is Chris Okasaki's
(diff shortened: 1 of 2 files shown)


petertseng:
> We are forced to create a dummy data definition, which makes the exercise a little less challenging

I don't think this is too serious. The student still has to figure out what they want the type to be, because they certainly won't be keeping the `Dummy`. So the only work we have relieved them of is the step of figuring out that we need to define the `CustomSet` type.

> We cannot export the data constructors to the test suite, because they do not exist yet.

For this type `CustomSet`, a library writer might not want to export the data constructors anyway since it might show internal implementation details. So not being able to export constructors seems OK to me.

Regarding what a library writer might do, I think they would probably at least export `CustomSet` if not its constructors, otherwise nobody can refer to the type in a signature, etc. But it works for us since our tests don't refer to the type. I only make this comment because it's removed from the export list of the stub solution.

Overall, I feel the disadvantages pointed out are acceptable. The text in HINTS.md seems good. If you can't think of any other drawbacks I think we can get this moving.

PR #427, opened 2016-11-08
series: rewrite to be flexible and improve stub.

rbasso:
- Change the test suite to accept multiple signatures.
- Add HINTS.md explaining how to work the exercise.
- Add a example solution using ByteStrings and Sequences.
- Add type signature to the stub solution.
- Add variables to the stub implementation.

These change also have the effect of allowing to run the test suite with the stub solution.

Related to #421.

### DIFF
--- a/exercises/series/HINTS.md
+++ b/exercises/series/HINTS.md
@@ -1,0 +1,23 @@
+## Hints
+
+To complete this exercise you need to implement the function `slices`,
+that takes a *text* and returns the subsequences of digits with a
+specified size:
+
+If it is your first time solving this exercise, it is recommended that you
+stick to the provided signature:
+
+```haskell
+slices :: Int -> String -> [[Int]]
+```
+
+Later, it may be a good idea to revisit this problem and play with other data
+types and libraries:
+
+- `ByteString`, from package *bytestring*.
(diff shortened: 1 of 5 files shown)


petertseng:
> These change also have the effect of allowing to run the test suite with the stub solution.

Oh, was it unable to before...?

Oh, indeed it was.

```
    • Ambiguous type variable ‘t0’ arising from a use of ‘shouldBe’
      prevents the constraint ‘(Eq t0)’ from being solved.
```

I see.

petertseng:
I agree with every change I see here. Can I take a moment to be a stickler about separate commits?

I think one commit should be "improve stub", and the other commit everything else - everything else seems related, but "improve stub" seemed like the odd duck out in the big commit and seemed like it could go in its own commit. True, that would make the commit very small, but it seemed to make sense to me.

Unless you think the two I mentioned are inextricably linked, in which case sure, just tell me why and then merge it.

rbasso:
You are right, they belong to separated commits. Fixed! :smile:

petertseng:
Oh good, I can edit approve/request changes comments now (we were not able to before, which is why I always make it a separate comment)

PR #431, opened 2016-11-11
grains: add signatures to stub solution

rbasso:
Since we changed `grains` to expect a `Maybe` in https://github.com/exercism/xhaskell/pull/197, we had two users that opened issues about not being able to compile it: https://github.com/exercism/xhaskell/issues/305 and https://github.com/exercism/xhaskell/issues/430

The test suite seems correct, but users have a hard time trying to infer the correct types based on test suite.

This PR just add the expected type signatures to the stub solution.

### DIFF
--- a/exercises/grains/src/Grains.hs
+++ b/exercises/grains/src/Grains.hs
@@ -1,5 +1,7 @@
 module Grains (square, total) where
 
-square = undefined
+square :: Integer -> Maybe Integer
+square n = undefined
 
+total :: Integer
 total = undefined


rbasso:
As time passes I get more convinced that all exercises should obey the following:

- [X] Have a stub solution
- [ ] Compile and run the tests with the stub solution
- [ ] Fail the first test with the stub solution

Not sure yet about:

- [ ] Have a `HINTS.md` file

rbasso:
As time passes I get more convinced that all exercises should obey the following:

- [X] Have a stub solution
- [ ] Compile and run the tests with the stub solution
- [ ] Fail the first test with the stub solution

**Edit:**
- [ ] Have variable names in the stub implementation and a body of `undefined`.

Not sure yet about:

- [ ] Have a `HINTS.md` file

PR #432, opened 2016-11-11
add dominoes exercise

petertseng:
Just like our song-writing exercuses (beer-song, food-chain, house),
this one has to make heavy use of property-based tests since the number
of possible answers for even a small input can be very large.

### DIFF
--- a/exercises/dominoes/examples/success-standard/src/Dominoes.hs
+++ b/exercises/dominoes/examples/success-standard/src/Dominoes.hs
@@ -1,0 +1,15 @@
+module Dominoes (chain) where
+
+import Data.List  (delete)
+import Data.Maybe (listToMaybe, mapMaybe)
+
+chain :: [(Int, Int)] -> Maybe [(Int, Int)]
+chain [] = Just []
+chain (x@(left, right):xs) = fmap (x:) (subchain xs left right)
+
+subchain :: [(Int, Int)] -> Int -> Int -> Maybe [(Int, Int)]
+subchain [] chainLeft chainRight = if chainLeft == chainRight then Just [] else Nothing
+subchain l  chainLeft chainRight = listToMaybe (mapMaybe subchain' l)
+  where subchain' (a, b) | a == chainRight = fmap ((a, b):) (subchain (delete (a, b) l) chainLeft b)
+        subchain' (a, b) | b == chainRight = fmap ((b, a):) (subchain (delete (a, b) l) chainLeft a)
+        subchain' _                        = Nothing
--- a/exercises/dominoes/test/Tests.hs
+++ b/exercises/dominoes/test/Tests.hs
@@ -1,0 +1,108 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE RecordWildCards #-}
+
+import Control.Monad     (forM_, unless)
+import Data.Foldable     (for_)
+import Data.Function     (on)
+import Test.Hspec        (Spec, describe, expectationFailure, it, shouldBe, shouldMatchList)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+import Text.Printf       (printf)
+
+import Dominoes (chain)
+
+main :: IO ()
+main = hspecWith defaultConfig {configFastFail = True} specs
+
+specs :: Spec
+specs = describe "dominoes" $
+          describe "chain" $ for_ cases test
+  where
+
+    test Case{..} = it description assertion
+      where
+        assertion = if expected
+                      then shouldBeChain $ chain input
+                      else chain input `shouldBe` Nothing
+
+        shouldBeChain Nothing = expectationFailure "should have had a chain, but didn't"
+        shouldBeChain (Just output) = do
+          output `shouldMatchDominoesOf` input
+          consecutivesShouldMatch output
+          endsShouldMatch output
+
+        shouldMatchDominoesOf = shouldMatchList `on` map sortDomino
+
+        consecutivesShouldMatch l = forM_ (indexedPairs l) $ \(d1, d2, i) ->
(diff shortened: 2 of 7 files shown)


exercises/dominoes/examples/success-standard/src/Dominoes.hs
```diff
@@ -0,0 +1,15 @@
...
+
+chain :: [(Int, Int)] -> Maybe [(Int, Int)]
+chain [] = Just []
+chain (x@(left, right):xs) = fmap (x:) (subchain xs left right)
+
+subchain :: [(Int, Int)] -> Int -> Int -> Maybe [(Int, Int)]
+subchain [] chainLeft chainRight = if chainLeft == chainRight then Just [] else Nothing
+subchain l  chainLeft chainRight = listToMaybe (mapMaybe subchain' l)
+  where subchain' (a, b) | a == chainRight = fmap ((a, b):) (subchain (delete (a, b) l) chainLeft b)
+        subchain' (a, b) | b == chainRight = fmap ((b, a):) (subchain (delete (a, b) l) chainLeft a)
```
petertseng: a little repetitive

exercises/dominoes/test/Tests.hs
```diff
@@ -0,0 +1,108 @@
...
+        endsShouldMatch [] = return ()
+        endsShouldMatch l@(d1:_) =
+          shouldBeConsecutive l (" last domino", last l) ("first domino", d1)
+
+        indexedPairs :: [a] -> [(a, a, Int)]
+        indexedPairs l = zip3 l (drop 1 l) [1..]
+
+        sortDomino (a, b) = if a > b then (b, a) else (a, b)
+
+        shouldBeConsecutive l (name1, d1@(_, d1r)) (name2, d2@(d2l, _)) =
```
petertseng: was having trouble with the name here

exercises/dominoes/test/Tests.hs
```diff
@@ -0,0 +1,108 @@
...
+          unless (d1r == d2l) $
+            expectationFailure $
+              printf "In chain %s:\n\t   right end of %s (%s)\n\tand left end of %s (%s)\n\tdidn't match!" (show l) name1 (show d1) name2 (show d2)
```
petertseng: quite a long line!

exercises/dominoes/test/Tests.hs
```diff
@@ -0,0 +1,108 @@
...
+        shouldMatchDominoesOf = shouldMatchList `on` map sortDomino
+
+        consecutivesShouldMatch l = forM_ (indexedPairs l) $ \(d1, d2, i) ->
+          shouldBeConsecutive l ("domino " ++ show i, d1) ("domino " ++ show (i + 1), d2)
+
+        endsShouldMatch [] = return ()
+        endsShouldMatch l@(d1:_) =
+          shouldBeConsecutive l (" last domino", last l) ("first domino", d1)
+
+        indexedPairs :: [a] -> [(a, a, Int)]
```
petertseng: otherwise, `Defaulting the following constraints to type ‘Integer’`

exercises/dominoes/test/Tests.hs
```diff
@@ -0,0 +1,108 @@
...
+          endsShouldMatch output
+
+        shouldMatchDominoesOf = shouldMatchList `on` map sortDomino
+
+        consecutivesShouldMatch l = forM_ (indexedPairs l) $ \(d1, d2, i) ->
+          shouldBeConsecutive l ("domino " ++ show i, d1) ("domino " ++ show (i + 1), d2)
+
+        endsShouldMatch [] = return ()
+        endsShouldMatch l@(d1:_) =
+          shouldBeConsecutive l (" last domino", last l) ("first domino", d1)
```
petertseng: space before `last` intentional, for alignment with `first`!

exercises/dominoes/test/Tests.hs
petertseng: And I had to use a custom message because The simple "expected 2 got 1" message wouldn't have been helpful enough in this situation.

exercises/dominoes/test/Tests.hs
```diff
@@ -0,0 +1,108 @@
...
+
+        indexedPairs :: [a] -> [(a, a, Int)]
+        indexedPairs l = zip3 l (drop 1 l) [1..]
+
+        sortDomino (a, b) = if a > b then (b, a) else (a, b)
```
rbasso: I really tried to find something more readable without `if...then...else`, but the best I could do was:

```haskell
sortDomino = on (&&&) uncurry min max
```

If feel that maybe there a type class for homogeneous tuple allowing us to sort that, but this has nothing to do with this PR. :+1:

exercises/dominoes/test/Tests.hs
```diff
@@ -0,0 +1,108 @@
...
+specs :: Spec
+specs = describe "dominoes" $
+          describe "chain" $ for_ cases test
+  where
+
+    test Case{..} = it description assertion
+      where
+        assertion = if expected
+                      then shouldBeChain $ chain input
+                      else chain input `shouldBe` Nothing
```
rbasso: Nice solution to the problem of matching `Maybe [a]`. :+1:

rbasso: Nice solution to the problem of matching `Maybe [a]` disregarding the order BTW. :+1:

exercises/dominoes/test/Tests.hs
petertseng: I thought about something like that, or written out like `(min a b, max a b)`. Or maybe avoid the repetition of `(a, b)` with `sortDomino d@(a, b) = if a > b then (b, a) else d` but the asymmetry looked odd I think.

petertseng:
Looks like we have nothing to change for now, so it seems we should merge it to get it out to the students.

PR #438, opened 2016-11-20
list-ops: import/test filter function of exercise code

stesie:
Currently the list-ops exercise's test case neither hides `filter` from `Prelude` nor does it import that function from the exercise code. Hence the test cases don't test the users implementation but the Prelude one's ...


petertseng:
Oops. Nice catch. Guess we missed it in #229 !

petertseng:
Thanks!

rbasso:
> ...Guess we missed it in #229 !

Yeah...I probably added a lot of new bugs rewriting the tests to use `hspec`. Sorry! :disappointed:

PR #439, opened 2016-11-21
new exercise bowling

abo64:


exercises/bowling/src/Bowling.hs
```diff
@@ -0,0 +1,14 @@
+module Bowling (bowlingStart, roll, score) where
+
+-- modify this appropriately
+data Bowling = Bowling
```
rbasso: Since #302 , the stub solutions do not have comments anymore.

We did this because most people where submitting solutions without removing them, and `HINTS.md` seemed the most appropriate place, so people will have to get used to reading the `README.md` file in case of trouble.

Also, to mark data types that need to be changed by the user, we are using `Dummy` and adding instructions in `HINTS.md`. Take a look at the exercises `forth`, `custom-set` and `binary-search-tree` that we fixed; all of them now have a `HINTS.md` and a data type to be implemented.

We are trying to make the stub solutions allow compiling and running the tests when that doesn't ruin the exercise, so it would be nice if this exercise could allow that from the start #421.

abo64:
I appreciate your continuous efforts to improve the exercises!  
It makes it somewhat harder for contributors, but the end result is worth it.

I considered to change `bowlingStart` to `empty` but decided to keep the more telling name. Hope that's ok with you.

rbasso:
> It makes it somewhat harder for contributors, but the end result is worth it.

Sorry for that! We should definitely write instructions on how to implement an exercise to make things easier for contributors (.

The fact is that until now we didn't standardized a format for the exercises, because we are still evolving the infrastructure and some of the exercises do not fit in anything.

An alternative would be to avoid standardization, but with more than 70 exercises it is really hard to keep things updated and manageable.

Issue #440, opened 2016-11-22
bowling: should we be precise about which roll, specifically, is erroneous?


petertseng:
Take, for example, the sequence of rolls `[5, 6, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]`

In this, we would expect the 6 is the erroneous roll. By this, we mean that if `game = roll 6 (roll 5 bowlingStart)`, no matter what you do to `game` you cannot get a score out of it.

The tests don't test that precisely - they currently cannot, since `roll` returns a `Bowling`, and there is no way to determine the above-described property of a `Bowling` without trying to `roll` it to completion.

So my question is: Would we like to be more precise about which roll is erroneous?

(I have argued the same [in the Rust implementation](https://github.com/exercism/xrust/pull/213#issuecomment-250988227))

The gist is that I like to fail fast and close to the source of the error, so that it is easy to understand what caused the error.

If we find this desirable, some proposals to do it:

* Add `isValid :: Bowling -> Bool`. I find this a little unclean.
* Change `roll`'s type to `roll :: Bowling -> Int -> Either ___ Bowling` (the ___ can be any user-defined error they like; the tests only need to test `isLeft` on it)

And a final question:

Should this be brought to x-common? Note that there will be a slight challenge in how to represent the "this specific roll should fail" in JSON (my current leading thought is keeping the `rolls` array but then adding a `roll_should_fail` entry)

petertseng:
EBWODP: The correct type of `roll` in my second proposal should be `roll :: Bowling -> Int -> Either ___ Bowling`

rbasso:
**Disclaimer**: I didn't solve this exercises yet in any language.

I don't like to much the idea of having an `isValid` function, but you proposal to change the roll type seems sound, @petertseng.

If you want to be flexible in the type of error, maybe it can be generalized to any [MonadError m e](http://hackage.haskell.org/package/mtl-2.2.1/docs/Control-Monad-Except.html#t:MonadError), but I never tried to write a test suite like this one before.

I would start with the rigid type `Int -> Either String Bowling`, because we'll need to supply a signature anyway, but making it more general is always nice. :smile: 

The only problem that I see in this approach is that we may be restricting too much the way to calculate the final score. Considering that we don't score partial games, is it hard to justify an interface that forces the student to partially evaluate it.

[...]

PR #441, opened 2016-11-24
integrated score function

abo64:
Changes as discussed in issue #440 .
I found it difficult to rephrase `HINTS.md`. It assumes the current `Requirements` of x-common's `description.md`. Any improvements here are especially welcome.
The example changes are based on the earlier version. So it might not be overly elegant.

### DIFF
--- a/exercises/bowling/HINTS.md
+++ b/exercises/bowling/HINTS.md
@@ -1,12 +1,6 @@
 ## Hints
 
-To complete this exercise, you need to create the data type `Bowling`
-and implement the following functions:
+For a functional language like Haskell it makes sense to calculate the game result in one go from the list of rolls (without an extra `roll` function).
+This result can then either be the total game score or an error for a list of irregular rolls.
 
-- `bowlingStart` returns an empty `Bowling`.
-- `roll` evaluates an input roll, returning the new `Bowling`.
-- `score` calculates the total score at the end if it is regular game.
-
-You will find a dummy data declaration and type signatures already in place,
-but it is up to you to define the functions and create a meaningful data type,
-newtype or type synonym.
+You will find a data declaration for the possible errors as well as the type signature for the `score` function.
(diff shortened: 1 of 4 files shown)


exercises/bowling/HINTS.md
```diff
@@ -1,12 +1,6 @@
...
+This result can then either be the total game score or an error for a list of irregular rolls.
 
-- `bowlingStart` returns an empty `Bowling`.
-- `roll` evaluates an input roll, returning the new `Bowling`.
-- `score` calculates the total score at the end if it is regular game.
-
-You will find a dummy data declaration and type signatures already in place,
-but it is up to you to define the functions and create a meaningful data type,
-newtype or type synonym.
+You will find a data declaration for the possible errors as well as the type signature for the `score` function.
```
rbasso: What about not mentioning anything about us diverging from *x-common*?

Maybe just put the instructions and wait until `description.md` is fixed. Something like this:

## Hints

To complete this exercise you need to implement the function `score`,
that takes a sequence of bowling *rolls* and returns the final score or
the appropriate error:

```haskell
score :: [Int] -> Either BowlingError Int

data BowlingError = IncompleteGame
                  | InvalidRoll { rollIndex :: Int, rollValue :: Int }
                  deriving (Eq, Show)
```

You will find these definitions already in place, but it is up to you
to define the function.

Keep in mind that the test suite expects the rolls to be numbered
starting from zero.

exercises/bowling/HINTS.md
```diff
@@ -1,12 +1,6 @@
 ## Hints
 
-To complete this exercise, you need to create the data type `Bowling`
-and implement the following functions:
+For a functional language like Haskell it makes sense to calculate the game result in one go from the list of rolls (without an extra `roll` function).
+This result can then either be the total game score or an error for a list of irregular rolls.
```
rbasso: > ... or an error for a list of irregular rolls.

This is a little confusing. Seems like we are expecting a list of errors.

exercises/bowling/HINTS.md
abo64: thank you, @rbasso

PR #451, opened 2016-12-04
difference-of-square: Comment out last test

### LINKED ISSUE #449: difference-of-squares: Last test may be too hard for beginners
Since I reviewed [this iteration](http://exercism.io/submissions/167f5b869b1b43018a6f1335691a2326), I'm wondering if we should remove the last test case or move this exercise a little later in the track.

Most users try to solve this exercise using `sum`, and there is no way for that to work with a huge list.

Other comments about it:

- http://exercism.io/submissions/765f0083fde1479a96f3e40043b7ffab
- http://exercism.io/submissions/b92f05a157724dbdb336b42577fc259e
- http://exercism.io/submissions/e53c9da929444e88817b9bd6dfc0012b
- http://exercism.io/submissions/dfa730dd5a95460187c649a9f7b8070c

###### We are certainly missing more people reviewing the exercises...

There are two paths to pass the last test:

- using `foldl'`.
- using [more math](https://en.wikipedia.org/wiki/Square_pyramidal_number).

Should we do something about it? What?

rbasso:
The test `huge difference` has caused some difficulties
to users because the most common solutions use the `sum`
function, that seems to be implemented as a `foldr`.

A naive implementation would certainly fail the test or,
in the worst case scenario, allocate all the memory and
crash the operating system.

Closes #449.

### DIFF
--- a/exercises/difference-of-squares/test/Tests.hs
+++ b/exercises/difference-of-squares/test/Tests.hs
@@ -64,7 +64,9 @@ specs = describe "differenceOfSquares" $ do
           difference (12 :: Integer)
           `shouldBe` (5434 :: Integer)
 
+      {-
       describe "huge difference" $
         it "difference (1234567890 :: Integer)" $
           difference (1234567890 :: Integer)
           `shouldBe` (580764307309260838625720836817589660 :: Integer)
+      -}


rbasso:
> maybe with an explanation to the student that they can uncomment the test if they desire to push the limits + a warning?

I'm not against it, but I decided to just merge this asap to avoid more complaints. Feel free to change it! 👍

PR #443, opened 2016-11-28
common: Add `stack.yaml` to common files folder and link `pov` as a test.

rbasso:
As discussed in #442, it would be nice to have a single `stack.yaml` file and link all the others to it.

We expect this approach will work, but it is better to test fetching an exercise before making huge changes.

This PR does the following:

- Add `common/stack.yaml`
- Add `common` to ignore section in `config.json`
- Fix `.travis.yml` to dereference links when copying files for testing
- Symbolic link `stack.yaml` from exercism `pov`

Related to #442.

Issue #442, opened 2016-11-28
stack.yaml: Avoiding multiplicity


rbasso:
All exercises in the Haskell track have exactly the same `stack.yaml` file. We do that to minimize downloads and keep a minimum of coherence.

Having 76 identical files is so ugly....

There was a short discussion about a solution to that in https://github.com/exercism/x-api/issues/29#issuecomment-237160865, but it seems that the idea was abandoned, and the discussion is now only about having files to be deployed outside the exercise's directory.

Possibly this track is alone in the need for a feature to deploy the same file to all exercises.

An alternative solution, almost equally ugly, is to symbolic link all `stack.yaml` files to the same one, outside the exercises folder.

A short discussion in Gitter indicated that the symbolic links would work, delivering the contents of the targeted file.

This if far from perfect, because we have to be sure that all exercises have the same symbolic links in place, but is not harder than checking if all the exercises have identical files. The advantage would be to allow us to update just one `stack.yaml` instead of 76, every time that we need to update the *resolver*.

If we decide to follow this path, first we should make a test, just to be sure that the symbolic link works as expected.

What do you think about it? @exercism/haskell @abo64. 

---

**Edit:** *On hold* until https://github.com/exercism/trackler/issues/13 is solved.

rbasso:
All exercises in the Haskell track have exactly the same `stack.yaml` file. We do that to minimize downloads and keep a minimum of coherence.

Having 76 identical files is so ugly....

There was a short discussion about a solution to that in https://github.com/exercism/x-api/issues/29#issuecomment-237160865, but it seems that the idea was abandoned, and the discussion now only about having files to be deployed outside the exercise's directory.

Possibly this track is alone in the need for a feature to deploy the same file to all exercises.

An alternative solution, almost equally ugly, is to symbolic link all `stack.yaml` files to the same one, outside the exercises folder.

A short discussion in Gitter indicated that the symbolic links would work, delivering the contents of the targeted file.

This if far from perfect, because we have to be sure that all exercises have the same symbolic links in place, but is not harder than checking if all the exercises have identical files. The advantage would be to allow us to update just one `stack.yaml` instead of 76, every time that we need to update the *resolver*.

If we decide to follow this path, first we should make a test, just to be sure that the symbolic link works as expected.
 
What do you think about it? @exercism/haskell @abo64.

petertseng:
[...]

PR #454, opened 2016-12-04
pov: Add dummy symbolic link

rbasso:
In #442 we conjectured that using symbolic links to a common `stack.yaml` would be feasible, but the test (#443) didn't work as expected. 

The result of the experiment was a `pov` exercise without a `stack.yaml` file, and #453 was opened on fix that.

To keep testing symbolic links, we still need to use `pov` as a test bed, but this time without breaking it, so we use another symbolic link name, unrelated to the exercise.

### DIFF
--- a/exercises/pov/.dummylink
+++ b/exercises/pov/.dummylink
@@ -1,0 +1,1 @@
+../../common/stack.yaml


petertseng:
I considered using a file whose name starts with `.`, but I don't really care that much either way. Maybe it will keep a student from asking about it, but who knows whether they would actually miss it

rbasso:
> I considered using a file whose name starts with `.`

Makes perfect sense, @petertseng! Fixed!

Issue #459, opened 2016-12-17
stubs: Default implementations for functions


rbasso:
Currently, most, if not all, the exercises have stub solutions that have the form:

```haskell
foo :: ...
foo = undefined
```

This is great, but the are two minor probems:

- A beginner student may not see immediately that it needs to implement that function.
- The error message returned from the test suite may be a little confusing:

```
leap
  isLeapYear
    1996 - leap year FAILED [1]

Failures:

  test/Tests.hs:18: 
  1) leap.isLeapYear 1996 - leap year
       uncaught exception: ErrorCall (Prelude.undefined
       CallStack (from HasCallStack):
         error, called at libraries/base/GHC/Err.hs:79:14 in base:GHC.Err
         undefined, called at src/LeapYear.hs:4:19 in leap-0.0.0-4uYVmwBGJQhBRFIbJzqiyG:LeapYear)
```

An alternative would be to use `error` with a custom message...

```haskell
foo = error "You need to implement this function."
```

... which is clearer in the source code and gives a slightly better error message:

```
leap
  isLeapYear
    1996 - leap year FAILED [1]

Failures:

  test/Tests.hs:18: 
  1) leap.isLeapYear 1996 - leap year
       uncaught exception: ErrorCall (You need to implement this function.
       CallStack (from HasCallStack):
         error, called at src/LeapYear.hs:4:19 in leap-0.0.0-4uYVmwBGJQhBRFIbJzqiyG:LeapYear)
```

Another options would be to use...

```haskell
foo = errorWithoutStackTrace "You need to implement this function."
```

... which is not so beautiful in the source code, but gives the best error message:

```
leap
  isLeapYear
    1996 - leap year FAILED [1]

Failures:

  test/Tests.hs:18: 
  1) leap.isLeapYear 1996 - leap year
       uncaught exception: ErrorCall (You need to implement this function.)
```

So, which of the three solutions is better? What would be the best message to use?

petertseng:
Well the stack trace is still somewhat useful, right? Since the line number shows which function we need to implement. I wouldn't mind keeping it. So my leading option is `error "You need to implement this function."`.

rbasso:
Another related question!

**Should we also provide variable names in the stub implementations?**

For example, instead of ...

```haskell
isLeapYear = error "You need to implement this function."
```

...we could write...

```haskell
isLeapYear year = error "You need to implement this function."
```

petertseng:
I looked for past discussions of "eta" and got:

[...]

PR #461, opened 2016-12-26
dibs: I will implement exercise run-length-encoding

ed359:


exercises/run-length-encoding/src/RunLength.hs
```diff
@@ -0,0 +1,9 @@
+module RunLength (decode, encode) where
+
+import Data.Char (isDigit)
```
rbasso: There are many ways to solve this problem without using `isDigit`, and it is not need for the type signature.

This import would direct people to look for a specific solution, reducing diversity. I think that it would be removed.

exercises/run-length-encoding/test/Tests.hs
```diff
@@ -0,0 +1,75 @@
...
+specs = describe "run-length-encoding" $ do
+          describe "decode" $ for_ decodeCases $ test decode
+          describe "encode" $ for_ encodeCases $ test encode
+          describe "both"   $ for_ bothCases   $ test (decode . encode)
+  where
+    test f Case{..} = it description $ f input `shouldBe` expected
+
+-- Test cases adapted from file
+-- `exercism/x-common/exercises/run-length-encoding/canonical-data.json`
+-- on 2016-12-26.
```
rbasso: This test suite could be better "translated" as a big `do` block, reproducing the exact test ordering from `x-common`.

That said, the cases ordering doesn't seem to be so important in this exercise, so I think it is great the way it is.

exercises/run-length-encoding/src/RunLength.hs
```diff
@@ -0,0 +1,9 @@
...
+
+decode :: String -> String
+decode = undefined
+
+encode :: String -> String
+encode = undefined
```
rbasso: This is perfect the way it is and also matches exactly what we did in all the other exercises.

Recently, we discussed in #459 that it would be nice to use `error` to give better messages than `undefined`, so an alternative would be:

```haskell
encode xs  = error "You need to implement this function."
```

No change is needed here, and we are not yet in the processes of rewriting the stub solutions. Your choice!

config.json
```diff
@@ -23,6 +23,12 @@
       ]
     },
     {
+      "slug": "run-length-encoding",
+      "difficulty": 1,
+      "topics": [
```
rbasso: Not sure what is the best position for this exercise, so I think it is ok here.

rbasso:
### Questions

- How should the `decode` function behave when the input ends with a number?
- How should the `encode` function behave when the input contains numbers?!

It seems that, using `String` to represent the encoded data, there is no simple way to tell the numbers from the characters...

Well...I'm digressing. This is a problem in the design of the exercise, not in the implementation, that seems correct. So this discussion should not prevent merging.

Any ideas about how to deal with that? @exercism/haskell

abo64:
I found something [here](https://wiki.multimedia.cx/index.php/Run_Length_Encoding).  
So one could use negative numbers to indicate a non-encoded character sequence. Additionally one would need some separator like a blank space.

For example,
`123AAA45` -> `-3 123 3 A -2 45`
(one could also encode number sequences: `111` -> `3 1`)

But all this would of course complicate the exercise quite a bit - IMO in an interesting way! :)

rbasso:
Found this open issue in https://github.com/exercism/x-common/issues/238.

exercises/run-length-encoding/test/Tests.hs
ed359: I made a conscious decision to change the order of the tests. I like having tests for different methods separated so that I can fully implement `decode` and test it before I work on `encode`. I prefer this to testing that both `encode` and `decode` work for the simplest case, and then another simple case, etc. I feel this workflow is nicer for beginners, especially since we stop the test on first failure by default. On the other hand, it's a pain to manually reorder things. I assume that in the long run the exercism project will implement an automated system for generating the tests from the common json files though that's another story.

config.json
ed359: This depends. As it stands this is a very simple exercise, but if some of the more complicated error-handling and number-handling ideas (here https://github.com/exercism/x-common/issues/238) are implemented it would no longer be simple. I put it before `rna-transcription` because at the moment it only requires simple types and not `Maybe` for error handling.

ed359:
I think a simple problem where we explicitly state that handling errors isn't necessary and that the alphabet we encode doesn't contain digits (though any other unicode is fine) is best here. But I'm happy to wait for the discussion to settle before this exercise gets added to haskell.

exercises/run-length-encoding/src/RunLength.hs
ed359: If at a later date the decision is made to replace `undefined` with `error` (a decision that I'd support) the change can easily be implemented with a single find-replace.

PR #462, opened 2017-01-08
SETUP.md: move to exercises/TRACK_HINTS.md

petertseng:
As https://github.com/exercism/exercism.io/issues/3240 states, we would
like to rename SETUP.md.

https://github.com/exercism/trackler/pull/17 has added the necessary
support for TRACK_HINTS.md, so we can now move this file.


.travis.yml
```diff
@@ -49,7 +49,7 @@ script:
                    --pedantic      `# Enable -Wall and -Werror.    `
     }
 
-    for exercise in ${TRAVIS_BUILD_DIR}/exercises/* ; do
+    for exercise in ${TRAVIS_BUILD_DIR}/exercises/*/ ; do
```
rbasso: Nice solution! 👍

petertseng:
I know you didn't request this, but we are also able to take a small precaution in case this doesn't go as we expect. I will propose the precaution in a PR and we can decide then if the precaution is necessary.

PR #463, opened 2017-01-08
SETUP.md: Add for paranoia

petertseng:
In #462, we moved it to TRACK_HINTS.md

HOWEVER, we are the very first track to do so, so the code is untested.
It is possible something can go wrong.

If we want to be paranoid, we can add a SETUP.md so that if TRACK_HINTS
fails we don't leave our students in the dark.

This SETUP.md is slightly different from TRACK_HINTS.md: It has only one
space after the period on the last line, rather than two.

This way, we can tell which of the two files is being used.

---

Note to reviewers: We *do not necessarily* need to merge this. This all depends on how paranoid we are feeling. I have confidence in https://github.com/exercism/trackler/blob/master/lib/trackler/implementation.rb#L105 that our TRACK_HINTS.md is in the right place, but it is possible to be paranoid, and delete the SETUP.md file once we are sure it works.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -1,0 +1,52 @@
+
+## Getting Started
+
+For installation and learning resources, refer to the
+[exercism help page](http://exercism.io/languages/haskell).
+
+## Running the tests
+
+To run the test suite, execute the following command:
+
+```bash
+stack test
+```
+
+#### If you get an error message like this...
+
+```
+No .cabal file found in directory
+```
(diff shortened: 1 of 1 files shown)


petertseng:
I'll take it since we really are the first track to use TRACK_HINTS.md. It is a low-cost test, whereas the cost of leaving students without SETUP.md is a little higher.

PR #464, opened 2017-01-08
travis: Build all stubs, and test all stubs except for untestable ones

### LINKED ISSUE #421: We should probably enforce in Travis that our stub solutions compile

petertseng:
Closes #421 

Now all stubs work with `stack build` in Travis, and all stubs except for a few exceptions work with `stack test --no-run-tests`, which compiles with the tests.

We use the .meta directory of https://github.com/exercism/trackler/pull/4 to store a DONT-TEST-STUB marker. This entire directory is not served to students.

---

This PR makes the assumption: **We want all the stubs to build**.

We can relax this assumption, and not even run `stack build` on selected stubs, by adding a `.meta/DONT-BUILD-STUB` file, if we desire. If you believe any change in this PR was made to make the stub build but should not have been made, please state the case, and we could add `.meta/DONT-BUILD-STUB`.

This PR makes the acknowledgement that it would be **ideal** if we can get all the stubs to run the first test and then fail. It then also acknowleges that we couldn't get all stubs to meet this goal at once (in fact, there are some stubs for which we might never want this goal to be met?!), but lets us track how close/far we are from this goal by inspecting how many `DONT-TEST-STUB` files we have.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,27 @@ script:
         exercisename=$(basename "$exercise")
         pushd ${exercise}
 
+        examplename="stub"
+        buildfolder="${TRAVIS_BUILD_DIR}/build/${exercisename}/${examplename}"
+        mkdir -p "${buildfolder}"
+        cp -rL stack.yaml package.yaml src test "${buildfolder}"
+
+        pushd $buildfolder
+
+        examplecache="${HOME}/.foldercache/${exercisename}/${examplename}/.stack-work"
+        mkdir -p "$examplecache"
+        ln -f -s "$examplecache"
+
+        if [ -f "${exercise}/.meta/DONT-TEST-STUB" ]; then
+            echo "only building stub"
+            stack build ${SET_RESOLVER} --install-ghc --no-terminal
+        else
+            echo "testing stub"
+            stack test ${SET_RESOLVER} --install-ghc --no-terminal --no-run-tests
+        fi
+
+        popd
+
         if ! stat -t examples/*/ > /dev/null 2>&1; then
             echo "No examples for ${exercise}!"
             exit 1
--- a/exercises/allergies/HINTS.md
+++ b/exercises/allergies/HINTS.md
@@ -6,5 +6,6 @@ with `Eq` and `Show` instances, and implement the following functions:
 - `allergies`
 - `isAllergicTo`
 
-Your can use the provided signatures if you are unsure about the types, but
-don't let them restrict your creativity.
+You will find a dummy data declaration and type signatures already in place,
+but it is up to you to define the functions and create a meaningful data type,
+newtype or type synonym.
--- a/exercises/clock/HINTS.md
+++ b/exercises/clock/HINTS.md
@@ -12,5 +12,6 @@ The function `fromInteger`, from `Num`, must convert minutes
 to 24 hour clock time. It is not necessary to have a sensible
 implementation of `abs` or `signum`.
 
-Your can use the provided signatures if you are unsure about the types, but
-don't let them restrict your creativity.
+You will find a dummy data declaration and type signatures already in place,
+but it is up to you to define the functions and create a meaningful data type,
+newtype or type synonym.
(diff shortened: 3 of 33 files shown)


exercises/allergies/HINTS.md
```diff
@@ -6,5 +6,6 @@ with `Eq` and `Show` instances, and implement the following functions:
 - `allergies`
 - `isAllergicTo`
 
-Your can use the provided signatures if you are unsure about the types, but
-don't let them restrict your creativity.
+You will find a dummy data declaration and type signatures already in place,
```
petertseng: If this is *too* different from the original ("Don't let them restrict your creativity"), please help me to combine the two in a sensible manner.

exercises/clock/HINTS.md
```diff
@@ -12,5 +12,6 @@ The function `fromInteger`, from `Num`, must convert minutes
 to 24 hour clock time. It is not necessary to have a sensible
 implementation of `abs` or `signum`.
 
-Your can use the provided signatures if you are unsure about the types, but
-don't let them restrict your creativity.
+You will find a dummy data declaration and type signatures already in place,
```
petertseng: If this is *too* different from the original ("Don't let them restrict your creativity"), please help me to combine the two in a sensible manner.

exercises/grade-school/.meta/DONT-TEST-STUB
```diff
@@ -0,0 +1,2 @@
+No type signatures on functions (TODO: Should we have them?),
```
petertseng: if you believe we should have them, I will gladly add them and remove this `DONT-TEST-STUB`

exercises/sublist/src/Sublist.hs
```diff
@@ -1,3 +1,6 @@
 module Sublist (Sublist(..), sublist) where
 
+data Sublist = Equal | Sublist | Superlist | Unequal deriving (Eq, Show)
```
petertseng: Potentially controversial. If we don't do this though, we have to `DONT-TEST-STUB` Sublist. I think it's reasonable to do it because it's not the student really has a choice on how to do this, they'll just read the test file and do it this way.

compare to:

* `go-counting` where we define `data Color = Black | White`
* `connect` where we define `data Mark = Cross | Nought`
* `bowling` where we define `data BowlingError = IncompleteGame | InvalidRoll { rollIndex :: Int, rollValue :: Int }`
* `robot-simulator` where we define `data Bearing = North | East | South | West`

.travis.yml
```diff
@@ -53,6 +53,27 @@ script:
         exercisename=$(basename "$exercise")
         pushd ${exercise}
 
+        examplename="stub"
```
petertseng: these lines have very high duplication with the stub-testing code below. I'm not sure if it's desirable that we deduplicate that.

petertseng: these lines (*stub*-testing cdode) have very high duplication with the *example*-testing code below. I'm not sure if it's desirable that we deduplicate that.

petertseng:
Highly likely to increase our build times. Reviewers should check to make sure the increased time is worth the increased assurance.

exercises/grade-school/.meta/DONT-TEST-STUB
rbasso: I see no reason to not have them. 👍 
But we can also fix that later.

exercises/sublist/src/Sublist.hs
rbasso: Agreed!

I'm usually in favor of anything to avoid the students having to read the test file.

Haskell is already hard enough to learn, and it is better to allow the students to see a lot of data declarations until they get used to it than let them stuck reading the test suites.

.travis.yml
rbasso: Seems fine for me. If we change our mind we can rewrite it later. 👍

exercises/allergies/HINTS.md
rbasso: It is great the way it is. 👍

petertseng:
Looking over it one last time, and if I don't see anything I'm merging it.

PR #465, opened 2017-01-08
grade-school: Add signatures to stub so it runs tests

petertseng:
Lingering work from #421 and #464.

~~It seems acceptable to define the `Grade` and `Student` type synonyms in
the stub, so that the expected return types of the functions will all be
self-explanatory.~~

~~The alternative is, of course, not to define them, and replace all
instances of `Grade` with `Int` and `Student` with `String`, and let the
students figure it out on their own.~~

Although the example solution defines `Grade` and `Student` type
synonyms, we generally leave the definition of these type synonyms to
individual students, so the type signatures provided will just have
`Int` and `String`.

---

What say y'all? Should we define `Grade` and `Student`, or let the students do it? Do we have a precedent? I will say the only other `type` we ever define in a stub is `type Coord = (Int, Int)` in `go-counting`. So I could be convinced to leave this one out.

### DIFF
--- a/exercises/grade-school/src/School.hs
+++ b/exercises/grade-school/src/School.hs
@@ -2,10 +2,14 @@ module School (School, add, empty, grade, sorted) where
 
 data School = Dummy
 
-add = undefined
+add :: Int -> String -> School -> School
+add gradeNum student school = undefined
 
+empty :: School
 empty = undefined
 
-grade = undefined
+grade :: Int -> School -> [String]
+grade gradeNum school = undefined
 
-sorted = undefined
+sorted :: School -> [(Int, [String])]
+sorted school = undefined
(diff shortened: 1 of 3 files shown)


petertseng:
I'm leaning toward not defining them. There's lots of precedent for not doing it.

exercises/grade-school/src/School.hs
```diff
@@ -2,10 +2,14 @@ module School (School, add, empty, grade, sorted) where
 
 data School = Dummy
 
-add = undefined
+add :: Int -> String -> School -> School
+add gradeNum student school = undefined
```
petertseng: I would have said `grade` here, but shadowing the function `grade` seems bad.

exercises/grade-school/src/School.hs
```diff
@@ -2,10 +2,14 @@ module School (School, add, empty, grade, sorted) where
...
 
+empty :: School
 empty = undefined
 
-grade = undefined
+grade :: Int -> School -> [String]
+grade gradeNum school = undefined
```
petertseng: again, the arg should be something other than `grade`, so I chose `gradeNum` like the example solution.

exercises/grade-school/src/School.hs
rbasso: Makes sense. I can't think of a better name.

PR #467, opened 2017-01-08
allergies, space-age, meetup-day; Fully define data in stub

### LINKED ISSUE #466: data and constructors required by tests: Should we provide them in stubs?
See in https://github.com/exercism/xhaskell/pull/464#discussion_r95073688 how I argued:

* The test is expecting the four constructors
* If we don't provide it, the stub solution doesn't compile with the tests.
* If we don't provide it, the student has to read the test file, find all the possible constructors, and make the `data` type exactly as we want it anyway.

So, I argued to provide it to the student.

We can see that this is distinct from the case where the test demands an abstract type (such as a `Robot`) and wishes to perform some operations on it. In this case:

* the test doesn't care about the implementation of the type
* the student can choose the representation
* therefore we make a `data Robot = Dummy` in the stub file

In this issue I'll discuss a few exercises of the first type:

* allergies: `Allergen`
* meetup: `Weekday` and `Schedule`
* space-age: `Planet`

So, should we give these three exercises the same treatment, and fully define the above-mentioned `data` types, with constructors?

(If doing so, we can remove the `.meta/DONT-TEST-STUB` file for each of these exercises)

petertseng:
This is the concrete proposal to go along with #466.

If y'all think #466 is a good idea, then this is the way to go.

Closes #466.

### DIFF
--- a/exercises/meetup/.meta/DONT-TEST-STUB
+++ b/exercises/meetup/.meta/DONT-TEST-STUB
@@ -1,2 +1,0 @@
-Can't build because the parameters of meetupDay are unused
-Would also have to export all the individual days of the week in stub; let students figure that out.
--- a/exercises/allergies/.meta/DONT-TEST-STUB
+++ b/exercises/allergies/.meta/DONT-TEST-STUB
@@ -1,1 +1,0 @@
-We would have to export all the allergens in the stub; let students figure that out.
--- a/exercises/allergies/src/Allergies.hs
+++ b/exercises/allergies/src/Allergies.hs
@@ -1,6 +1,14 @@
 module Allergies (Allergen(..), allergies, isAllergicTo) where
 
-data Allergen = Dummy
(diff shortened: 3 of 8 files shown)


exercises/meetup/.meta/DONT-TEST-STUB
```diff
@@ -1,2 +0,0 @@
-Can't build because the parameters of meetupDay are unused
```
petertseng: also, this particular line (parameters) was back when I was trying to build --pedantic (it did not work out)

petertseng: but I forgot to remove the line

abo64:
Looks good to me, too.

Btw, in the Scala track we also decided to provide stubs at least for the first couple of exercises, see [this issue](https://github.com/exercism/xscala/issues/137) to not deter beginners.

PR #472, opened 2017-01-17
lens-person: Remove lens from example solution

petertseng:
Here is a proposal.

`lens` is quite a large library with many dependencies, right? And it is not strictly required to pass the tests of the `lens-person` exercise.

So, what if we were to reduce our dependency footprint by not using `lens` at all?

I'm not sure whether this will actually have a noticeable benefit for us and/or whether using `lens` is actually causing us any problems. Let's see.

### DIFF
--- a/exercises/lens-person/examples/success-standard/src/Person.hs
+++ b/exercises/lens-person/examples/success-standard/src/Person.hs
@@ -1,9 +1,10 @@
-{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE Rank2Types #-}
 
 module Person where
 
-import Control.Lens
-import Data.Time.Calendar
+import Control.Applicative (Const(..), getConst)
+import Data.Functor.Identity (Identity(..), runIdentity)
+import Data.Time.Calendar (Day, toGregorian, fromGregorian)
 
 data Person = Person {
       _name    :: Name,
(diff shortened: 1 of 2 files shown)


petertseng:
> `lens` is quite a large library, right?

I mostly mean that it has many dependencies.

exercises/lens-person/examples/success-standard/src/Person.hs
```diff
@@ -36,17 +37,44 @@ data Gregorian = Gregorian {
...
+type Getting s a = (a -> Const a a) -> s -> Const a s
+type Setting s t a b = (a -> Identity b) -> s -> Identity t
 
-gregorianDay :: Iso' Gregorian Day
-gregorianDay = iso toDay fromDay
-  where
-    toDay (Gregorian y m d) = fromGregorian y m d
-    fromDay d' = let (y, m, d) = toGregorian d' in Gregorian y m d
+lens :: (s -> a) -> (s -> b -> t) -> Lens s t a b
+lens sa sbt afb s = fmap (sbt s) (afb (sa s))
```
petertseng: these names are completely useless, but I'm not sure I can do better. I suppose `sa` can be named `get` and `sbt` can be `set`, but I'm not sure about `afb`

exercises/lens-person/examples/success-standard/src/Person.hs
```diff
@@ -36,17 +37,44 @@ data Gregorian = Gregorian {
...
+
+over :: Setting s t a b -> (a -> b) -> s -> t
+over l f = runIdentity . l (Identity . f)
+
+set :: Setting s t a b -> b -> s -> t
+set l v = over l (const v)
+
+-- boilerplate lenses:
+
+born :: Lens' Person Born
```
petertseng: all are boilerplate, but I didn't want to deal with template haskell here (should I? how is it done?)

rbasso: I definitely cannot help you with that. 😄

petertseng:
> If the lens library is not used anywhere else

It is indeed not.

PR #475, opened 2017-01-29
Add hello-world exercise

ErikSchierboom:
This PR adds the `hello-world` exercise, as requested in https://github.com/exercism/todo/issues/34.

This is my first PR in this language track, so I hope everything is in order. Let me know if it isn't.

### DIFF
--- a/exercises/hello-world/examples/success-standard/src/HelloWorld.hs
+++ b/exercises/hello-world/examples/success-standard/src/HelloWorld.hs
@@ -1,0 +1,4 @@
+module HelloWorld (hello) where
+
+hello :: String
+hello = "Hello, World!"
--- a/config.json
+++ b/config.json
@@ -5,6 +5,12 @@
   "active": true,
   "exercises": [
     {
+      "slug": "hello-world",
+      "difficulty": 1,
+      "topics": [
+      ]
+    },
+    {
       "slug": "leap",
       "difficulty": 1,
       "topics": [
--- a/exercises/hello-world/HINTS.md
+++ b/exercises/hello-world/HINTS.md
@@ -1,0 +1,6 @@
+## Hints
+
+To complete this exercise, you need to implement the `hello` function.
+
+You will find the type signature for `hello` already in place,
+but it is up to you to define the function.
--- a/exercises/hello-world/examples/success-standard/package.yaml
+++ b/exercises/hello-world/examples/success-standard/package.yaml
@@ -1,0 +1,16 @@
+name: hello-world
+
+dependencies:
+  - base
+
+library:
+  exposed-modules: HelloWorld
+  source-dirs: src
+
+tests:
+  test:
+    main: Tests.hs
+    source-dirs: test
+    dependencies:
+      - hello-world
+      - hspec
--- a/exercises/hello-world/package.yaml
+++ b/exercises/hello-world/package.yaml
@@ -1,0 +1,19 @@
+name: hello-world
+
+dependencies:
+  - base
+
+library:
+  exposed-modules: HelloWorld
+  source-dirs: src
+  dependencies:
+  # - foo       # List here the packages you
+  # - bar       # want to use in your solution.
+
+tests:
+  test:
+    main: Tests.hs
+    source-dirs: test
+    dependencies:
+      - hello-world
+      - hspec
--- a/exercises/hello-world/src/HelloWorld.hs
+++ b/exercises/hello-world/src/HelloWorld.hs
@@ -1,0 +1,4 @@
+module HelloWorld (hello) where
+
+hello :: String
+hello = error "You need to implement this function."
--- a/exercises/hello-world/stack.yaml
+++ b/exercises/hello-world/stack.yaml
@@ -1,0 +1,1 @@
+resolver: lts-7.16
--- a/exercises/hello-world/test/Tests.hs
+++ b/exercises/hello-world/test/Tests.hs
@@ -1,0 +1,13 @@
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
+import HelloWorld (hello)
+
+main :: IO ()
+main = hspecWith defaultConfig {configFastFail = True} specs
+
+specs :: Spec
+specs = describe "hello-world" $
+
+    it "hello" $
+      hello `shouldBe` "Hello, World!"


petertseng:
Don't forget to add to config.json.

The questions below assume hello world is intended to be the first exercise:

I have slight unease with introducing `Maybe` on the very first exercise: https://github.com/exercism/x-common/issues/290#issuecomment-237998505

Looking at Learn You a Haskell, the first-ever mention of `Maybe` is at http://learnyouahaskell.com/modules, the 7th chapter out of 14.

Then again, we also introduce `Maybe` relatively early: in exercise `rna-transcription`, `Maybe` appears.

Maybe I'm underestimating the audience's ability to understand `Maybe`, and maybe it'll all be fine to include `Maybe` in the first exercise. If anyone has evidence to support one side or the other, please help by sharing it.

ErikSchierboom:
@petertseng I can only agree, it's just that `hello-world` is supposed to be the first exercise. There was a discussion somewhere on this. @kytrinyx?

petertseng:
Maybe I need to look at https://github.com/exercism/x-common/issues/290 again. I want to know what the purpose of this first exercise should be.

1. Should it be just an introduction to Exercism, to demonstrate to a student how to fetch and submit? In that case, it's language-agnostic, and doesn't belong in any specific track.
2. Should it be an introduction to the language track, to demonstrate to a student how to run the testing framework for the language? In that case, maybe it's best if the function-under-test has as little logic in it as possible. I'm thinking a `the-answer` exercise that asks the function to do nothing more than return 42, unconditionally.
    * sub-question: How much experience with the language is assumed by the time the student gets to this first exercise? The concern of "Don't introduce `Maybe` in first exercise" is only relevant if we assume the student has some level of knowledge similar to "has read Learn You A Haskell at least through the first seven chapters".

petertseng:
Damn, inverted my logic.

Deleting:

~~The concern of "Don't introduce Maybe in first exercise" is only relevant~~

Replacing with (change in bold):

The concern of "Don't introduce Maybe in first exercise" is **not** relevant

petertseng:
> I can only agree, it's just that `hello-world` is supposed to be the first exercise.

Hmm... given that I have your agreement too, I am *very* inclined to reject the supposition entirely. Or if `hello-world` is being hard-required on all tracks (I note that I see no such initiative), I would highly suggest that at least *this* track make it "Hello, world!" unconditionally.

Of course, that's in contravention of the README, so now we have another problem to deal with. I wonder if we can stick something in HINTS.md that says something to the effect of "ha ha, ignore what it says above", but that's clumsy and indicates disharmony in the world.

Or another possibility is to go campaign to all tracks that unconditional "Hello, world!" is the right thing to do. And then all of them have to change their tests because otherwise *they*'re in contravention of a hypothetical README that says it's unconditional "Hello, world!". I'm slightly disinclined to be the one who spends time embarking on this campaign.

All this above is why I suspect `the-answer` may be the answer, so we don't have `hello-world` in an inconsistent state with some tracks doing it one way and some tracks another.

> I want to know what the purpose of this first exercise should be.

* https://github.com/exercism/todo/issues/12
* https://github.com/exercism/exercism.io/issues/2021

These suggest that it's a combination of *all* of these: intro-to-exercism, intro-to-track, intro-to-tdd.

Argh, this puts a hole into the `the-answer` idea. Since there is only one test in `the-answer`, there is only one "failing test -> passing test" transition, which isn't necessarily enough to show TDD.

ErikSchierboom:
I would also have thought that `the-answer` would be the correct answer.

kytrinyx:
My original idea for hello world was to give a very easy exercise to make sure that the person gets everything wired together. The secondary goal is to introduce people to how to run the test suite in a given language.

I don't see any particular reason to require each track to have a hello world exercise, and certainly not with any particular implementation. Especially where the implementation introduces concepts that make it frustrating to get started.

petertseng:
There is one potential implementation I can offer that is faithful to the README and does not introduce `Maybe`:

Use the same technique from `sum-of-multiples` (remember #100?):

```haskell
helloDefault :: String
helloDefault = error "You need to implement this"

hello :: String -> String
hello = error "You need to implement this"
```

If *that* is deemed usable, we could go with it.

* One could say that the `Default` is a bit suspicious-looking (would someone actually do that in real Haskell?)
* hello-world could be a place where we provide copious comments that explain everything that goes on - is any of that not served by our existing documents, including exercises/TRACK_HINTS.md, docs/TESTS.md, docs/INSTALLATION.md? Any chance it can fit in `leap` instead?

kytrinyx:
We could also considering making the README simpler and more—uhm—flexible.

petertseng:
I can try to do that, but I'm having a real hard. Currently I'm complaining that it's overspecified. But if I go too far in the other direction and make it underspecified, that's no good either. I'll make my case anyway, since there's no harm in asking.

rbasso:
It seems I'm a little late, but I'll try to summarize my thoughts about it:

#### About the implementation

The implementation seems perfect, @ErikSchierboom, but I'm also not comfortable with a `Maybe` in the first exercise.

#### About the `hello-world` exercise

I like the idea of a introductory exercise, but I think that`hello-world` misses the point when it tries to include an optional argument. To be faithful to the ["Hello, World!"](https://en.wikipedia.org/wiki/%22Hello,_World!%22_program) spirit, I think that it should be as simple as possible:

```haskell
hello :: String
```

But as @petertseng already pointed out, this would be against what is specified in `x-common`.

Also, the `description.md` says a lot about TTD, which is great, but not exactly the best match for Haskell, as it's static, strongly-typed system pushes a *Type-Driven Development* - where tests are still relevant, of course, but probably not as in dynamic languages.

ErikSchierboom:
My personal preference would be to simply the README of the `hello-world` exercise. I do think there is real value in a starting exercise that only requires you to simply return the "Hello, World!" string. As the implementation thus becomes trivial (which it isn't with optional values), it allows people to focus on the other things:

1. Figuring out how to write an implementation
2. Running the tests
3. Submitting the exercise

The `hello-world` exercise thus becomes just an introduction to the track's infrastructure. If one would then use `leap` as the second exercise, that would introduce a second aspect: creating an implementation by using multiple tests.

I think the main disdvantage of simplifying the `hello-world` exercise is that all tracks already implementing this exercise should also do this. I do think it's worth it though.

kytrinyx:
I think this discussion is important enough that we should promote it to the discussions repository. If we can get some sort of consensus we can draft an issue that can be submitted to all the tracks that have a hello-world currently.

petertseng:
> I think this discussion is important enough that we should promote it to the discussions repository. 

Indeed, such a discussion is at https://github.com/exercism/x-common/issues/520 (Sorry, I chose the x-common repository instead of the discussion repository).

I was hoping that you personally did not have any objection to it, because this is one case where curriculum matters and user experience are pretty inter-related. I will take the fact that you haven't immediately objected indignantly as one half of a good sign.

rbasso:
Considering that exercism/x-common#520 is going to be merged soon, I think there is no reason to keep this PR *on hold* anymore.

ErikSchierboom:
So I should update this PR to do the simple "Only return the 'Hello, World!'" string version, right?

rbasso:
That would be great!

> So I should update this PR to do the simple "Only return the 'Hello, World!'" string version, right?

ErikSchierboom:
I've just updated the PR. The updated exercise is very simple and a good first exercise now I guess.

exercises/hello-world/examples/success-standard/src/HelloWorld.hs
```diff
@@ -0,0 +1,4 @@
+module HelloWorld (hello) where
+
+hello :: String
+hello = "Hello, World!"
```
rbasso: I worry a little that maybe this way the student will be able to pass the tests without finding the source file that needs to be changed.

ErikSchierboom: I'm not sure I understand. Isn't this (putting the sample in an `examples` directory) the way other exercises also work?

rbasso: Oh...sorry! It is perfect. I just confused the files. 👍

PR #477, opened 2017-01-30
travis: extract functionality to scripts; behave differently based on test type. accumulate: add fail-reverse example

petertseng:
Individual commits have explanations for each.

### DIFF
--- a/bin/test-all-examples
+++ b/bin/test-all-examples
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+set -e
+
+if [ $# -eq 0 ]; then
+    echo "usage: $0 /path/to/exercise"
+    exit 1
+fi
+
+exercisedir=$1
+
+if ! stat -t ${exercisedir}/examples/*/ > /dev/null 2>&1; then
+    echo "No examples for ${exercisedir}!"
+    exit 1
+else
+    mydir=$(dirname $0)
+    for example in ${exercisedir}/examples/*/ ; do
+        $mydir/test-example $example
+    done
+fi
--- a/bin/test-example
+++ b/bin/test-example
@@ -1,0 +1,35 @@
+#!/bin/sh
+
+if [ $# -eq 0 ]; then
+    echo "usage: $0 /path/to/exercise/example"
+    exit 1
+fi
+
+exampledir=$(cd "$1" && pwd)
+
+examplename=$(basename "$exampledir")
+exercisedir=$(cd "$exampledir/../.." && pwd)
+exercisename=$(basename "$exercisedir")
+xhaskell=$(cd "$(dirname "$0")/.." && pwd)
+
+buildfolder="${TRAVIS_BUILD_DIR:-$xhaskell}/build/${exercisename}/${examplename}"
+mkdir -p "${buildfolder}"
+cp -rL ${exercisedir}/stack.yaml ${exercisedir}/test ${exampledir}/* "${buildfolder}"
+
+cd $buildfolder
+
+if [ -n "$TRAVIS" ]; then
+    cachedir="$HOME"
+else
+    cachedir="$xhaskell"
+fi
+examplecache="${cachedir}/.foldercache/${exercisename}/${examplename}/.stack-work"
+mkdir -p "$examplecache"
+ln -f -s "$examplecache"
+
+echo "testing ${exampledir}"
+# SET_RESOLVER passed by .travis.yml - sets --resolver if not current.
+stack test ${SET_RESOLVER} `# Select the correct resolver. `\
+           --install-ghc   `# Download GHC if not in cache.`\
+           --no-terminal   `# Terminal detection is broken.`\
+           --pedantic      `# Enable -Wall and -Werror.    `
--- a/bin/test-example
+++ b/bin/test-example
@@ -14,7 +14,7 @@ xhaskell=$(cd "$(dirname "$0")/.." && pwd)
 
 buildfolder="${TRAVIS_BUILD_DIR:-$xhaskell}/build/${exercisename}/${examplename}"
 mkdir -p "${buildfolder}"
-cp -rL ${exercisedir}/stack.yaml ${exercisedir}/test ${exampledir}/* "${buildfolder}"
+cp -R -L ${exercisedir}/stack.yaml ${exercisedir}/test ${exampledir}/* "${buildfolder}"
 
 cd $buildfolder
 
--- a/bin/test-example
+++ b/bin/test-example
@@ -12,8 +12,16 @@ exercisedir=$(cd "$exampledir/../.." && pwd)
 exercisename=$(basename "$exercisedir")
 xhaskell=$(cd "$(dirname "$0")/.." && pwd)
 
-buildfolder="${TRAVIS_BUILD_DIR:-$xhaskell}/build/${exercisename}/${examplename}"
+# We would ideally use a tmpdir here,
+# but stack will invalidate its cache if the path changes,
+# so the best alternative seems to be to put it in xhaskell.
+buildfolder="${xhaskell}/build/${exercisename}/${examplename}"
 mkdir -p "${buildfolder}"
+cleanup() {
+    rm -r "${xhaskell}/build"
+}
+trap cleanup EXIT INT TERM
+
 cp -R -L ${exercisedir}/stack.yaml ${exercisedir}/test ${exampledir}/* "${buildfolder}"
 
 cd $buildfolder
--- a/bin/test-example
+++ b/bin/test-example
@@ -35,9 +35,39 @@ examplecache="${cachedir}/.foldercache/${exercisename}/${examplename}/.stack-wor
 mkdir -p "$examplecache"
 ln -f -s "$examplecache"
 
-echo "testing ${exampledir}"
-# SET_RESOLVER passed by .travis.yml - sets --resolver if not current.
-stack test ${SET_RESOLVER} `# Select the correct resolver. `\
-           --install-ghc   `# Download GHC if not in cache.`\
-           --no-terminal   `# Terminal detection is broken.`\
-           --pedantic      `# Enable -Wall and -Werror.    `
+exampletype=$(echo "$examplename" | cut -d- -f1)
+
+runstack () {
+    # SET_RESOLVER passed by .travis.yml - sets --resolver if not current.
+    stack "$@" ${SET_RESOLVER} `# Select the correct resolver. `\
+               --install-ghc   `# Download GHC if not in cache.`\
+               --no-terminal   `# Terminal detection is broken.`\
+               --pedantic      `# Enable -Wall and -Werror.    `
+}
+
+if [ "$exampletype" = "success" ]; then
+    echo "testing ${exampledir} - should succeed"
+    # Implicit exit value: this is last thing in this path,
+    # so the script's exit value = stack's exit value.
+    # If statements are reordered, please preserve this property.
+    runstack "test"
+elif [ "$exampletype" = "fail" ]; then
+    echo "testing ${exampledir} - should build, but fail tests"
+    if ! runstack "build"; then
+        echo "${exampledir} build failed unexpectedly"
+        exit 1
+    fi
+    if runstack "test"; then
+        echo "${exampledir} test succeeded unexpectedly"
+        exit 1
+    fi
+elif [ "$exampletype" = "error" ]; then
+    echo "testing ${exampledir} - should fail to build"
+    if runstack "test" "--no-run-tests"; then
+        echo "${exampledir} build succeeded unexpectedly"
+        exit 1
+    fi
+else
+    echo "unknown example type: ${exampledir}"
+    exit 1
+fi
--- a/bin/test-stub
+++ b/bin/test-stub
@@ -1,0 +1,35 @@
+#!/bin/sh
+
+if [ $# -eq 0 ]; then
+    echo "usage: $0 /path/to/exercise"
+    exit 1
+fi
+
+exercisedir=$(cd "$1" && pwd)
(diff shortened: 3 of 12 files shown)


petertseng:
Open question - should the files have their .sh extensions or not

Also note I kept them in shell script because they're just straight ports of what was already in .travis.yml... despite me continually saying I'd write in Haskell. I kind of preferred to just get it done with what work has already been done. I guess I'm just not much of a rewriter.

bin/test-example
```diff
@@ -0,0 +1,48 @@
...
+if [ "$exampletype" = "success" ]; then
+    echo "testing ${exampledir} - should succeed"
+    # SET_RESOLVER passed by .travis.yml - sets --resolver if not current.
+    stack test ${SET_RESOLVER} `# Select the correct resolver. `\
+               --install-ghc   `# Download GHC if not in cache.`\
+               --no-terminal   `# Terminal detection is broken.`\
+               --pedantic      `# Enable -Wall and -Werror.    `
+elif [ "$exampletype" = "fail" ]; then
+    echo "testing ${exampledir} - should build, but fail tests"
+    if ! stack build ${SET_RESOLVER} --install-ghc --no-terminal --pedantic; then
```
petertseng: this duplication is ... not ideal. I will propose a way to deal with it.

bin/test-example
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+# need bash, not sh, for default variable ${var:-default}
+
+exampledir=$1
+
+examplename=$(basename "$exampledir")
+exercisedir=$(dirname $(dirname "$exampledir"))
+exercisename=$(basename "$exercisedir")
+
+buildfolder="${TRAVIS_BUILD_DIR:-.}/build/${exercisename}/${examplename}"
```
petertseng: hacky so that this can work locally even if TRAVIS_BUILD_DIR not specified.

bin/test-stub
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+# need bash, not sh, for default variable ${var:-default}
+
+exercisedir=$1
+exercisename=$(basename "$exercisedir")
+examplename="stub"
+
+buildfolder="${TRAVIS_BUILD_DIR:-.}/build/${exercisename}/${examplename}"
```
petertseng: hacky so that this can work locally even if TRAVIS_BUILD_DIR not specified.

bin/test-stub
```diff
@@ -0,0 +1,25 @@
...
+mkdir -p "${buildfolder}"
+cp -rL ${exercisedir}/stack.yaml ${exercisedir}/package.yaml ${exercisedir}/src ${exercisedir}/test "${buildfolder}"
+
+cd $buildfolder
+
+examplecache="${HOME}/.foldercache/${exercisename}/${examplename}/.stack-work"
```
petertseng: OK to create a .foldercache in user's personal HOME if running locally? If not, how do we deal with this? `if [ -n "$TRAVIS" ] examplecache="something-else"` perhaps?

bin/test-example
```diff
@@ -0,0 +1,53 @@
...
+mkdir -p "${buildfolder}"
+cp -rL ${exercisedir}/stack.yaml ${exercisedir}/test ${exampledir}/* "${buildfolder}"
+
+cd $buildfolder
+
+examplecache="${HOME}/.foldercache/${exercisename}/${examplename}/.stack-work"
```
petertseng: OK to create a .foldercache in user's personal HOME if running locally? If not, how do we deal with this? `if [ -n "$TRAVIS" ] examplecache="something-else"` perhaps?

bin/test-example
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+# need bash, not sh, for default variable ${var:-default}
+
+exampledir=$1
+
+examplename=$(basename "$exampledir")
+exercisedir=$(dirname $(dirname "$exampledir"))
```
petertseng: I would like t use `realpath`, but it doesn't exist on travis...???

The effect of not using `realpath` is that you cannot say in an `examples` dir `test-example success-standard` because then `exercisedir` becomes `.` instead of `..` as it should be.

petertseng:
Hmm, I should have a solution of `error` type, but I can't think of a good one that I want to include.

bin/test-example
rbasso: We probably need to cache in a separate folder, so I think it is ok to create it in the home folder.

bin/test-example
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
rbasso: I usually favor `#!/usr/bin/env bash` as a rule, so that it uses the bash it finds on the PATH. I don't know if I tested that in Travis.

Using `#!/bin/bash` will break the script in NixOS - which is not a problem for most of the tracks - but Nix is specially popular among Haskell programmers.

NobbZ: Using env is definitively the favored and most platform independent way. Also it works on Travis without problems.

petertseng: 👍 

One thing that I was not sure on was whether `/bin/sh` is OK to leave intact, but my quick research tells me that it is (please let me know if I was wrong).

Actually I will say, I would rather have scripts use `/bin/sh` rather than `bash`, so it would be best if it would not use bash-specific features. I believe the only one is the default-var. I'll probably add slightly more verbosity.

petertseng: Does anyone know whether http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02 is a standard that can be relied upon by all POSIX-compliant shells, not just bash? If so, default variable is POSIX-compliant, which is all I need.

It is important to for me to be able to assume that `/bin/sh` isn't bash, but possibly `dash` - this fact might be true for at least one machine I own.

petertseng: It works on the machines of mine where `/bin/sh` isn't bash, so I have changed all shebangs to `/bin/sh`. I would like to note for the record that it will remain important to me that these scripts do not use bash-specific features, maybe I should document that somewhere (don't know where to...)

bin/test-example
petertseng: I would suggest that in xhaskell is slightly better, if running locally.

bin/test-example
petertseng: Travis can help enforce that there are no Bash-specific features in /bin/sh files. /bin/sh is not bash on Travis. I hope it stays that way.

petertseng:
Now using the xhaskell directory as the destination of `build` and `.foldercache` if not in Travis. But I can't test this yet, and it will be hours until I can.

bin/test-example
```diff
@@ -0,0 +1,61 @@
...
+        echo "${exampledir} build failed unexpectedly"
+        exit 1
+    fi
+    if runstack "test"; then
+        echo "${exampledir} test succeeded unexpectedly"
+        exit 1
+    fi
+elif [ "$exampletype" = "error" ]; then
+    echo "testing ${exampledir} - should fail to build"
+    if runstack "build"; then
```
rbasso: Is it your intention not to build the tests in this case?

I guess that stack build will not build the tests, so a solution that cannot compile the test suite may not error when it should. Right?

I'll try to find time to take a better look at that soon.

petertseng: This is true. `test --no-run-tests` would be better.

bin/test-example
```diff
@@ -0,0 +1,60 @@
...
+        echo "${exampledir} build failed unexpectedly"
+        exit 1
+    fi
+    if runstack "test"; then
+        echo "${exampledir} test succeeded unexpectedly"
+        exit 1
+    fi
+elif [ "$exampletype" = "error" ]; then
+    echo "testing ${exampledir} - should fail to build"
+    if runstack "test" "--no-run-tests"; then
```
petertseng: this excessive quoting is not really necessary. it could be removed, if requested.

petertseng:
going to squash "test-example: test --no-run-test" into "test-example: behave differently based on test type" and "test-example, test-stub: Build in xhaskell if not Travis" into "travis: extract functionality to scripts"

petertseng:
I think the `error` example is too trivial, and would rather replace it with a non-trivial one, but I can't decide on one yet. The one I have now is just to make sure that things are working.

rbasso:
Maybe we should remove the build folder after testing. Keeping it can contaminate future tests without adding savings in time.

In Travis this already happens automatically at the end, but I wonder if we should remove it after each tested example:

- Create a build path that is unique to each exercise/[example,stub].
- Build and test there
- Remove the entire build folder

What do you think about it?

ps: I'm stil facing some issues here depending from where I'm running the scripts. I'll try to give a more detailed report maybe still today.

petertseng:
> depending on from where I'm running the scripts

This is a little troublesome - the script can't be run from within the `bin` directory right now (it will get the wrong value of xhaskell), nor within any `examples` directory (it will get the wrong value of `exercisename`)

This is very undesirable behaviour. Since travis doesn't have `realpath`, I may have to resort to a `cd && pwd` to get the paths I care about...!

I will also experiment in a few hours.

> Maybe we should remove the build folder after testing

Indeed, there are no time savings to be had. I am thinking cleaning up is the right thing to do.

petertseng:
Sometimes I wish we could test the tests. How do I know that this script correctly exits 1 when a `fail` solution passes the tests? etc.

petertseng:
(for this current version, I had to test that manually, but what if a change comes later on and breaks thsi property?)

petertseng:
I now have many `readlink -f`, which hopefully Travis has, to allow to get a full path.

This also fixed a problem with running `test-all-examples .` - the name of the exercise directory would be incorrectly called `.` instead of... the actual exercise's name.

**Note: now there is very little hope of compatibility with BSDs, including Mac OS X**. `readlink -f` doesn't work there. Then again, neither does `cp -rL`.

petertseng:
> Note: now there is very little hope of compatibility with BSDs, including Mac OS X

And really I would like to make that better, for example by using `realpath`, but Travis doesn't have it.

Maybe if we put in `.travis.yml` that we want to be Trusty VMs instead of Precise...? But I don't know if there is a longer wait time for those, or something.

petertseng:
Or maybe I could throw in a lot of `cd && pwd`...

petertseng:
> neither does `cp -rL`

that can be fixed with `cp -R -L`

petertseng:
Now it works on BSDs. And I really hope it works on Travis too.

bin/test-example
```diff
@@ -0,0 +1,69 @@
...
+fi
+
+exampledir=$(cd "$1" && pwd)
+
+examplename=$(basename "$exampledir")
+exercisedir=$(cd "$exampledir/../.." && pwd)
+exercisename=$(basename "$exercisedir")
+xhaskell="$(cd "$(dirname "$0")/.." && pwd)"
+
+buildfolder=$(mktemp -d)
```
rbasso: I needs testing, but I believe that using a temp directory, will invalidate the `.stack-work` cache.

rbasso: Elaborating...it seems that `.stack-work` keeps some information about the sources path and ignores the previous cache if the source tree moves to another path.

rbasso: If this is really the case, the build path needs to be a function from (exercise, example/stub).

petertseng: Looks like you're right. Interesting. Clearly there is much I don't understand.

petertseng: > the build path needs to be a function from (exercise, example/stub).

Oh wow, so I can't even place it in the current directory (like I planned to).

I'll try always putting it in xhaskell then.

rbasso: I met these problems last time I played with `.travis.yml`. I really wanted to use a temporary directory too...

If I remember correctly, `.stack-work` depends on the path and timestamps, but I never really tested the timestamp thing.

bin/test-example
```diff
@@ -0,0 +1,70 @@
...
+
+examplename=$(basename "$exampledir")
+exercisedir=$(cd "$exampledir/../.." && pwd)
+exercisename=$(basename "$exercisedir")
+xhaskell="$(cd "$(dirname "$0")/.." && pwd)"
+
+buildfolder="${xhaskell}/build/${exercisename}/${examplename}"
+mkdir -p "${buildfolder}"
+cleanup() {
+    rm -r "${xhaskell}/build"
```
petertseng: * If remove just `buildfolder`, then `build/$exercisename` will remain for each exercise, which isn't a good cleanup. So I remove all of `build`.

rbasso: Makes perfect sense! 👍

bin/test-all-examples
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+set -e
+
+if [ $# -eq 0 ]; then
+  echo "usage: $0 /path/to/exercise"
```
petertseng: going to make indentation consistent, at 4 spaces.

I don't care whether we choose 4 or 2 as long as we are consistent.

bin/test-example
```diff
@@ -0,0 +1,73 @@
...
+fi
+
+exampledir=$(cd "$1" && pwd)
+
+examplename=$(basename "$exampledir")
+exercisedir=$(cd "$exampledir/../.." && pwd)
+exercisename=$(basename "$exercisedir")
+xhaskell=$(cd "$(dirname "$0")/.." && pwd)
+
+# We would ideally use a tmpdir here,
```
petertseng: new comment added to explain.

petertseng:
> Sorry for taking so long...

No trouble at all, don't worry! Only thing that happens is if I want to test an example locally against any of my PRs I have to merge the two branches locally, and that is easy enough.

petertseng:
> I think the error example is too trivial, and would rather replace it with a non-trivial one, but I can't decide on one yet. The one I have now is just to make sure that things are working.

We can make one later if we think of one.

PR #478, opened 2017-01-30
DO NOT MERGE: travis sh check

petertseng:
Does travis /bin/sh script fail if given a bash-specific feature?


NobbZ:
No check necessary. Travis runs Ubuntu containers. On Ubuntu sh is replaced by a sim link to bash per default. 

It will get problematic when we test Mac instances though. There ah is sim linked to dash afaik.

petertseng:
> On Ubuntu sh is replaced by a sim link to bash per default.

The undesired result. I wanted Travis CI to fail.

petertseng:
Fails as desired.

```
/home/travis/build/exercism/xhaskell/bin/fetch-configlet: 5: /home/travis/build/exercism/xhaskell/bin/fetch-configlet: Syntax error: Bad for loop variable
```

So as I hoped, on travis /bin/sh is not bash.

petertseng:
Note for posterity: `if [[ "a" == "a" ]]; then echo "yes"; fi` does cause a `[[: command not found`, but didn't fail the Travis build for some reason.

rbasso:
I think that, when `sh` is a link to `bash`, the runtime detects it was called as `sh` and changes behavior accordingly:

https://www.gnu.org/software/bash/manual/bashref.html#Bash-POSIX-Mode

http://unix.stackexchange.com/questions/44836/when-sh-is-a-symlink-to-bash-or-dash-bash-limits-itself-to-posix-compliance-so

If I remember correctly from my times playing with [Linux from Scratch](http://www.linuxfromscratch.org/), a few other compiled program and scripts share the same behavior, so you cannot expect - as a rule - that linked executables behave as their targets.

petertseng:
As an aside (not relevant to Haskell track), that is a useful way to introduce extra behaviour into a family of executables, or simply have one executable act as many executables. Busybox takes this approach: https://busybox.net/FAQ.html

> If the Busybox executable is renamed to one of the commands it supports, it will act as that command automatically

I'll make one more check, but basically what I hear here is that the only safe thing to assume about `/bin/sh` is that it implements http://pubs.opengroup.org/onlinepubs/9699919799/ (is that right?), and scripts with `/bin/sh` shebang should be written accordingly.

petertseng:
dash.

PR #480, opened 2017-01-31
phone-number: simplify to number only

petertseng:
areaCode and prettyPrint are hereby removed.

https://github.com/exercism/x-common/pull/477

---

Reviewers: many cases were changed as a result of syncing to x-common. Are there any cases in particular you wish to keep?

In particular, note that all `Just` results are `1234567890`. If you wish for more variety, please speak up.

### DIFF
--- a/exercises/phone-number/examples/success-standard/src/Phone.hs
+++ b/exercises/phone-number/examples/success-standard/src/Phone.hs
@@ -1,28 +1,11 @@
-module Phone (number, areaCode, prettyPrint) where
-import Data.Char (isDigit)
+module Phone (number) where
+import Data.Char (isDigit, isLetter)
 
 number :: String -> Maybe String
 number input
+  | any isLetter input = Nothing
   | len == 10 = Just digits
   | len == 11 && head digits == '1' = Just $ tail digits
   | otherwise = Nothing
   where digits = filter isDigit input
         len = length digits
-
-parts :: String -> Maybe (String, String, String)
-parts input = case number input of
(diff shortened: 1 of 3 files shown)


petertseng:
Don't forget to change HINTS.

petertseng:
> Don't forget to change HINTS.

There is none.

exercises/phone-number/examples/success-standard/src/Phone.hs
```diff
@@ -1,28 +1,11 @@
...
+import Data.Char (isDigit, isLetter)
 
 number :: String -> Maybe String
 number input
+  | any isLetter input = Nothing
   | len == 10 = Just digits
   | len == 11 && head digits == '1' = Just $ tail digits
   | otherwise = Nothing
-  where digits = filter isDigit input
+  where digits = filter (\c -> isDigit c || isLetter c) input
```
abo64: The `isLetter` case is already handled in L6, so you can keep the original `digits` definition.

petertseng: oh, oops. I mixed up and thought for some reason I put `any isLetter digits` instead of `any isLetter input`. Of course, if `any isLetter input` works, then I'll take it. Coming up shortly.

abo64:
IMO "more variety" is not really needed.

petertseng:
OK, let's take it, given it's a faithful representation of x-common. If we want to add back any track-specific tests, someone send another PR.

PR #481, opened 2017-01-31
prime-factors: use cases of x-common

petertseng:
https://github.com/exercism/x-common/pull/513

---

reviewers: your opinion will be appreciated on whether to incorporate the test descriptions into the file.

1. Do nothing.
2. Include as comments.
3. Include as an element of the tuple, then use `it` on them so thye become test descriptions.

### DIFF
--- a/exercises/prime-factors/test/Tests.hs
+++ b/exercises/prime-factors/test/Tests.hs
@@ -19,17 +19,12 @@ specs = describe "prime-factors" $
         explanation = show n
         assertion   = primeFactors n `shouldBe` expected
 
-    -- As of 2016-07-31, there was no reference file
-    -- for the test cases in `exercism/x-common`.
+    -- Test cases adapted from `exercism/x-common/prime-factors` on 2017-02-01.
 
     cases = [ (          1,                 [] )
             , (          2,                [2] )
-            , (          3,                [3] )
-            , (          4,             [2, 2] )
-            , (          6,             [2, 3] )
-            , (          8,          [2, 2, 2] )
             , (          9,             [3, 3] )
-            , (         27,          [3, 3, 3] )
-            , (        625,       [5, 5, 5, 5] )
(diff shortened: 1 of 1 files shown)


petertseng:
(My inclination is go for 3, but I didn't want to spend the time if nobody else cared)

abo64:
3 is more consistent with most other test suites, but it is indeed questionable whether it is worth the effort.

Maybe create a separate issue, but then for checking all test suites if they include the descriptions?

petertseng:
I don't necessarily think we need descriptions for all exercises - roman-numerals doesn't really need any https://github.com/exercism/xhaskell/blob/master/exercises/roman-numerals/test/Tests.hs - unless https://github.com/exercism/x-common/pull/451 comes up with something good.

rbasso:
Considering that one of the goals of Exercism seems to be keeping some degree of cross-language homogeneity, I would favor option 3, even in cases where the descriptions do not add too much to the tests.

exercises/prime-factors/test/Tests.hs
```diff
@@ -14,22 +14,16 @@ specs = describe "prime-factors" $
...
-            , (          9,             [3, 3] )
-            , (         27,          [3, 3, 3] )
-            , (        625,       [5, 5, 5, 5] )
-            , (     901255,   [5, 17, 23, 461] )
-            , (93819012551, [11, 9539, 894119] ) ]
+    cases = [ ("no factors",                              1,                 [] )
+            , ("prime number",                            2,                [2] )
+            , ("square of a prime",                       9,             [3, 3] )
+            , ("cube of a prime",                         8,          [2, 2, 2] )
+            , ("product of primes and non-primes",       12,          [2, 2, 3] )
```
petertseng: someone who is looking closely will see that there is a shortcut being taken in the indentation: the comma overhangs the 3 two lines below.

PR #482, opened 2017-01-31
nucleotide-count: simplify to only nucleotideCounts

petertseng:
It was decided that the exercise was doing too many things, and count
really is just one element of nucleotideCounts anyway.

https://github.com/exercism/x-common/pull/505


rbasso:
I had problems merging this one after the "Update branch" two times, probably because we where merging things at the same time. Not a problem, but I'll wait for you to merge this one, @petertseng. 😄

PR #484, opened 2017-01-31
triangle: Add case of 2a == b + c, then mark divergence from x-common

petertseng:
reviewers: it is possible that this will be controversial. Please let me know if instead you think we should move to the new test model. I will not have time to do that.

### DIFF
--- a/exercises/triangle/test/Tests.hs
+++ b/exercises/triangle/test/Tests.hs
@@ -25,7 +25,8 @@ specs = describe "triangle" $
       where
         assertion = triangleType a b c `shouldBe` expected
 
-    -- Test cases adapted from `exercism/x-common/triangle.json` on 2016-08-03.
+    -- Test cases created from old version of `x-common` tests on 2017-01-31,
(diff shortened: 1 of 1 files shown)


exercises/triangle/test/Tests.hs
```diff
@@ -25,7 +25,8 @@ specs = describe "triangle" $
       where
         assertion = triangleType a b c `shouldBe` expected
 
-    -- Test cases adapted from `exercism/x-common/triangle.json` on 2016-08-03.
+    -- Test cases created from old version of `x-common` tests on 2017-01-31,
```
petertseng: our only other diverging test: https://github.com/exercism/xhaskell/blob/master/exercises/crypto-square/test/Tests.hs#L29

rbasso:
> reviewers: it is possible that this will be controversial. Please let me know if instead you think we should move to the new test model. I will not have time to do that.

We can always change again later... 👍 

Thanks, @petertseng!

Issue #473, opened 2017-01-19
stubs: Add explicit arguments to functions (good for git practice)


rbasso:
As discussed in #459, it seems to be a good idea to add explicit arguments to all the stub solutions, because...

- It complements the type signature, documenting what the arguments are.
- It is easier for beginners that are not yet used to [pointfree style](https://wiki.haskell.org/Pointfree) programming.

For example, in the following code,

```haskell
sayHello :: String -> String
sayHello = error "You need to implement this function."
```

... it is not immediately clear what the argument for the function is. Adding an explicit argument solves the problem:

```haskell
sayHello :: String -> String
sayHello name = "You need to implement this function."
```

Sometimes, the meaning is obvious and generic argument names can be used:

```haskell
concatenate :: [a] -> [a] -> [a]
concatenate xs ys = "You need to implement this function."
```

Considering that this is a great opportunity to expose students to good argument naming practices, we should be extremely careful when choosing the argument names. This is how I usually try to name the arguments:

- First I try to choose a name that help documenting the functions.
- If a more descriptive name is not possible or for some reason undesirable, I use generic argument names, trying to follow what I usually see in the *Prelude*:
  - `x`, `y` and `z` for arguments of the same type.
  - `xs`, `ys` and `zs` for lists and strings.
  - `xss`, `yss`, `zss` for lists of lists.
  - `f`, `g` and `h` for functions.
  - `p` for functions that return a `Bool`.

Below there is a list of exercises that do not have yet explicit arguments in the stubs. The changes are kind of trivial, so this is a good opportunity for anyone trying to get used to `git`:

 - [x] accumulate
 - [x] allergies
 - [x] anagram
 - [x] atbash-cipher
 - [x] bank-account
 - [x] binary-search-tree
 - [x] binary (deprecated)
 - [x] bob
 - [x] bowling
 - [x] clock
 - [x] connect
 - [x] crypto-square
 - [x] custom-set
 - [x] difference-of-squares
 - [x] etl
 - [x] forth
 - [ ] gigasecond (deprecated)
 - [x] go-counting
 - [x] grade-school
 - [x] grains
 - [x] hamming
 - [ ] hexadecimal (deprecated)
 - [x] kindergarten-garden
 - [x] largest-series-product
 - [x] lens-person
 - [x] linked-list
 - [x] list-ops
 - [x] luhn
 - [x] matrix
 - [x] minesweeper
 - [x] nth-prime
 - [x] nucleotide-count
[...]

PR #497, opened 2017-02-14
stubs: add explicit argument to allergies function

mttakai:
Anagram: add explicit arguments to stub function

Related to issue #473.
see also #496  #491

### DIFF
--- a/exercises/allergies/src/Allergies.hs
+++ b/exercises/allergies/src/Allergies.hs
@@ -14,4 +14,4 @@ allergies :: Int -> [Allergen]
 allergies = error "You need to implement this function."
 
 isAllergicTo :: Allergen -> Int -> Bool
-isAllergicTo = error "You need to implement this function."
+isAllergicTo allergen n = error "You need to implement this function."


exercises/allergies/src/Allergies.hs
```diff
@@ -14,4 +14,4 @@ allergies :: Int -> [Allergen]
 allergies = error "You need to implement this function."
 
 isAllergicTo :: Allergen -> Int -> Bool
-isAllergicTo = error "You need to implement this function."
+isAllergicTo allergen n = error "You need to implement this function."
```
rbasso: Seems great!

exercises/allergies/src/Allergies.hs
```diff
@@ -14,4 +14,4 @@ allergies :: Int -> [Allergen]
 allergies = error "You need to implement this function."
```
rbasso: This function also needs an explicit argument. You can call it `n` to keep consistency.

rbasso:
Considering that this PR had no activity in the last two weeks, I'll close it for now.

Of course, we can reopen it later if @mttakai decides to continue with it.

Issue #498, opened 2017-02-21
Property-based Testing?


abo64:
For some exercises it makes perfect sense to also add some QuickCheck tests.  
A good example would be `simple-linked-list`.

I am aware that this could be quite confusing for people who are unfamiliar with property-based testing.  
And not all languages support it which might be a killer criterion in itself.  
On the other hand the concept could be important enough to expose people to it?

I would like to hear your opinion about this.

rbasso:
I agree that they make sense in Haskell - we even have two exercises that use *QuickCheck* - but I never spent the time to get comfortable enough with it to add new properties. I guess it is time to stop postponing. 😄 

In practice, using property-based testing would makes us diverge at least a little from `x-common` (adding new tests). I don't see any problem in following this route.

In the future, using exclusively property-based tests in some exercise could allow writing less prescriptive test suites, so I support the experiment.

abo64:
Great.
I know Scala's [ScalaCheck](https://www.scalacheck.org/), but I wanted to get into QuickCheck anyways.
So I might just give it a try for `simple-linked-list`.

rbasso:
> Which are those two exercises?

`octal` and `trinary`. Since I know them, they use QuickCheck.

> So I might just give it a try for `simple-linked-list`.

It seems perfect to start with it!

petertseng:
Also don't forget the Diamond exercise, whose readme at https://github.com/exercism/x-common/blob/master/exercises/diamond/description.md has a list of properties that can be used to test. http://blog.ploeh.dk/2015/01/10/diamond-kata-with-fscheck/ has some ideas as well. https://github.com/exercism/x-common/pull/191#issuecomment-254589087 confirms that it would be good to try to test it with property-based tests.

petertseng:
We could also try to retrofit our Dominoes tests at https://github.com/exercism/xhaskell/blob/master/exercises/dominoes/test/Tests.hs#L28 to use QuickCheck?

abo64:
I am sure there are more exercises suitable for property-based testing. My guess would be most of them - if you are just willing to invest enough inventiveness and imagination. :)

One has to be careful, though: Some properties might reveal too much about the solution. Just consider `bob`.

petertseng:
Since this was a question (should we have it) and the question was answered (affirmatively), we are done here, right?

Unless we want to be specific about which exercises it would be good to add QuickCheck to.

petertseng:
[...]

PR #499, opened 2017-02-21
simple-linked-list: Add QuickCheck tests

abo64:
#498 Property-based Testing
This is just to have some concrete code for discussion.

I simply used the existing QuickCheck tests from `octal` and `trinary` to get started.

One immediate problem is the compiler warning about the orphanned instance declaration. Is there some easy way to solve this?

### DIFF
--- a/exercises/simple-linked-list/examples/success-standard/package.yaml
+++ b/exercises/simple-linked-list/examples/success-standard/package.yaml
@@ -14,3 +14,4 @@ tests:
     dependencies:
       - simple-linked-list
       - hspec
+      - QuickCheck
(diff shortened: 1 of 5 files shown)


abo64:
One more problem: `LinkedList` must be `deriving (Show, Eq)`.

petertseng:
I don't think it will be a big burden to ask for `Show` and `Eq`, because it is asked for in other exercises as well, such as https://github.com/exercism/xhaskell/blob/master/exercises/binary-search-tree/HINTS.md , allergies, clock, matrix, maybe there are others

PR #504, opened 2017-03-12
Add explicit args to Exercises - Part 1

lpalma:
Part of #473 

Perhaps the next PRs will be smaller, so they can be reviewed easier.

I was thinking about some of the exercises covered in this PR and how much more complex they are in comparison to the first 10 - 15 available exercises (the ones I've completed so far).
Once the student get there, he will have crossed a few Haskell / FP concepts, which is one of the goals of this Track.

Consequently, wouldn't [eta-conversion](https://wiki.haskell.org/Eta_conversion) also be part of the concepts he should familiarize with? Therefore, I believe that, starting from a complex enough exercise, keeping the function arguments out would create a better learning experience.

### DIFF
--- a/exercises/atbash-cipher/src/Atbash.hs
+++ b/exercises/atbash-cipher/src/Atbash.hs
@@ -1,7 +1,7 @@
 module Atbash (decode, encode) where
 
 decode :: String -> String
-decode = error "You need to implement this function."
+decode phrase = error "You need to implement this function."
 
 encode :: String -> String
-encode = error "You need to implement this function."
+encode phrase = error "You need to implement this function."
--- a/exercises/connect/src/Connect.hs
+++ b/exercises/connect/src/Connect.hs
@@ -3,4 +3,4 @@ module Connect (Mark(..), winner) where
 data Mark = Cross | Nought deriving (Eq, Show)
 
 winner :: [String] -> Maybe Mark
-winner = error "You need to implement this function."
+winner board = error "You need to implement this function."
--- a/exercises/crypto-square/src/CryptoSquare.hs
+++ b/exercises/crypto-square/src/CryptoSquare.hs
@@ -1,4 +1,4 @@
 module CryptoSquare (encode) where
 
 encode :: String -> String
-encode = error "You need to implement this function."
+encode xs = error "You need to implement this function."
--- a/exercises/custom-set/src/CustomSet.hs
+++ b/exercises/custom-set/src/CustomSet.hs
@@ -19,40 +19,40 @@ import Prelude hiding (null)
 data CustomSet a = Dummy deriving (Eq, Show)
 
 delete :: a -> CustomSet a -> CustomSet a
-delete = error "You need to implement this function."
+delete x set = error "You need to implement this function."
 
 difference :: CustomSet a -> CustomSet a -> CustomSet a
-difference = error "You need to implement this function."
+difference setA setB = error "You need to implement this function."
 
 empty :: CustomSet a
 empty = error "You need to implement this function."
 
 fromList :: [a] -> CustomSet a
-fromList = error "You need to implement this function."
+fromList xs = error "You need to implement this function."
 
 insert :: a -> CustomSet a -> CustomSet a
-insert = error "You need to implement this function."
+insert x set = error "You need to implement this function."
 
 intersection :: CustomSet a -> CustomSet a -> CustomSet a
-intersection = error "You need to implement this function."
+intersection setA setB = error "You need to implement this function."
 
 isDisjointFrom :: CustomSet a -> CustomSet a -> Bool
-isDisjointFrom = error "You need to implement this function."
+isDisjointFrom setA setB = error "You need to implement this function."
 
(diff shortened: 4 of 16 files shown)


rbasso:
> Consequently, wouldn't eta-conversion also be part of the concepts he should familiarize with? 
> Therefore, I believe that, starting from a complex enough exercise, keeping the function arguments out would create a better learning experience.

Is desirable that, after a few exercises, the student start developing point-free skills. Certainly the people solving the last exercises have a at least a basic competence at eta-reduction.

I thought about the explicit arguments for later exercises more as documentation to help the students identify what each argument is, and maybe also as an additional exposure to idiomatic argument naming, which is usually something people slowly learn reading the function's implementations in Hackage.

Following the same reasoning, we also have the pattern-matching, which isn't possible without changing some explicit arguments, and that is also a must-learn syntax.

I'm unsure about what is the best for the track, so I'll copy the other maintainers to see what they think about it. @exercism/haskell

exercises/connect/src/Connect.hs
```diff
@@ -3,4 +3,4 @@ module Connect (Mark(..), winner) where
 data Mark = Cross | Nought deriving (Eq, Show)
 
 winner :: [String] -> Maybe Mark
-winner = error "You need to implement this function."
+winner board = error "You need to implement this function."
```
rbasso: `board` is certainly more descriptive than `xss`. 👍

exercises/crypto-square/src/CryptoSquare.hs
```diff
@@ -1,4 +1,4 @@
 module CryptoSquare (encode) where
 
 encode :: String -> String
-encode = error "You need to implement this function."
+encode input = error "You need to implement this function."
```
rbasso: Here, `input` doesn't add anything. I would normally go with `xs`, or something more descriptive, like `plainText`.

exercises/custom-set/src/CustomSet.hs
```diff
@@ -19,40 +19,40 @@ import Prelude hiding (null)
 data CustomSet a = Dummy deriving (Eq, Show)
 
 delete :: a -> CustomSet a -> CustomSet a
-delete = error "You need to implement this function."
+delete elem set = error "You need to implement this function."
```
rbasso: I never thought about how to name sets...`set` seems ok. 👍

exercises/difference-of-squares/src/Squares.hs
```diff
@@ -1,10 +1,10 @@
 module Squares (difference, squareOfSums, sumOfSquares) where
 
 difference :: Integral a => a -> a
-difference = error "You need to implement this function."
+difference n = error "You need to implement this function."
```
rbasso: `n` fits perfectly!

exercises/atbash-cipher/src/Atbash.hs
```diff
@@ -1,7 +1,7 @@
 module Atbash (decode, encode) where
 
 decode :: String -> String
-decode = error "You need to implement this function."
+decode phrase = error "You need to implement this function."
```
lpalma: I also prefer following the convention (`f` or `g` for functions, `xs` `ys` for generic lists, etc.) and I'm happy to change.

The reason I choose `phrase` here instead, was to match the test suite, believing that it would be easier for the student to understand the test cases.

I followed this same approach for other exercises, trying to match the test suite terms whenever possible, but I'm happy to change them as well.

rbasso: Please feel free to keep or to change the names you chose.

lpalma:
I'm happy with the changes now 👍

rbasso:
Thanks! 😄 

Just one final question before merging.

Would you like to squash the commits that affect the same exercise together, to create a more clean history?

lpalma:
>Would you like to squash the commits that affect the same exercise together, to create a more clean history?

done 👍

exercises/crypto-square/src/CryptoSquare.hs
```diff
@@ -1,4 +1,4 @@
 module CryptoSquare (encode) where
 
 encode :: String -> String
-encode = error "You need to implement this function."
+encode xs = error "You need to implement this function."
```
petertseng: sorry that I know I'm contradicting https://github.com/exercism/xhaskell/pull/504#discussion_r105556991 here, but it seemed a bit inconsistent here to use `xs` but `phrase` in `atbash-cipher`?

exercises/largest-series-product/src/Series.hs
```diff
@@ -1,4 +1,4 @@
 module Series (largestProduct) where
 
 largestProduct :: Int -> String -> Maybe Integer
-largestProduct = error "You need to implement this function."
+largestProduct series digits = error "You need to implement this function."
```
petertseng: `series` seems confusing to me since an integer is not really a series. It is the size of the seires we are considering within the digits. I would think either `size` or `span`.

exercises/crypto-square/src/CryptoSquare.hs
rbasso: I agree!

PR #506, opened 2017-03-14
Add Explicit Args To Exercises - Part 2

lpalma:
Part of #473 

I also renamed arguments from some functions following https://github.com/exercism/xhaskell/pull/504#discussion_r105571421 and https://github.com/exercism/xhaskell/pull/504#discussion_r105571584

**Note:** Could we tick the master list (from #473) with the updated exercises?
PR #504 together with the current one cover up to `prime-factors` inclusive

### DIFF
--- a/exercises/atbash-cipher/src/Atbash.hs
+++ b/exercises/atbash-cipher/src/Atbash.hs
@@ -1,7 +1,7 @@
 module Atbash (decode, encode) where
 
 decode :: String -> String
-decode phrase = error "You need to implement this function."
+decode cipherText = error "You need to implement this function."
 
 encode :: String -> String
-encode phrase = error "You need to implement this function."
+encode plainText = error "You need to implement this function."
--- a/exercises/list-ops/src/ListOps.hs
+++ b/exercises/list-ops/src/ListOps.hs
@@ -13,25 +13,25 @@ import Prelude hiding
   ( length, reverse, map, filter, foldr, (++), concat )
 
 foldl' :: (b -> a -> b) -> b -> [a] -> b
-foldl' = error "You need to implement this function."
+foldl' f z xs = error "You need to implement this function."
 
 foldr :: (a -> b -> b) -> b -> [a] -> b
-foldr = error "You need to implement this function."
+foldr f z xs = error "You need to implement this function."
 
 length :: [a] -> Int
-length = error "You need to implement this function."
+length xs = error "You need to implement this function."
 
 reverse :: [a] -> [a]
-reverse = error "You need to implement this function."
+reverse xs = error "You need to implement this function."
 
 map :: (a -> b) -> [a] -> [b]
-map = error "You need to implement this function."
+map f xs = error "You need to implement this function."
 
 filter :: (a -> Bool) -> [a] -> [a]
-filter = error "You need to implement this function."
+filter p xs = error "You need to implement this function."
 
 (++) :: [a] -> [a] -> [a]
 xs ++ ys = error "You need to implement this function."
 
 concat :: [[a]] -> [a]
-concat = error "You need to implement this function."
+concat xss = error "You need to implement this function."
--- a/exercises/luhn/src/Luhn.hs
+++ b/exercises/luhn/src/Luhn.hs
@@ -1,11 +1,16 @@
 module Luhn (addends, checkDigit, checksum, create, isValid) where
 
-addends = error "You need to implement this function."
+addends :: Integer -> [Integer]
+addends n = error "You need to implement this function."
 
-checkDigit = error "You need to implement this function."
+checkDigit :: Integer -> Integer
+checkDigit n = error "You need to implement this function."
 
-checksum = error "You need to implement this function."
+checksum :: Integer -> Integer
+checksum n = error "You need to implement this function."
 
-create = error "You need to implement this function."
+create :: Integer -> Integer
+create n = error "You need to implement this function."
 
-isValid = error "You need to implement this function."
+isValid :: Integer -> Bool
+isValid n = error "You need to implement this function."
--- a/exercises/matrix/src/Matrix.hs
+++ b/exercises/matrix/src/Matrix.hs
@@ -17,31 +17,31 @@ import Data.Vector (Vector)
 data Matrix a = Dummy deriving (Eq, Show)
 
 cols :: Matrix a -> Int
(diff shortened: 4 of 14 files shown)


exercises/matrix/src/Matrix.hs
```diff
@@ -17,31 +17,31 @@ import Data.Vector (Vector)
...
-fromList = error "You need to implement this function."
+fromList xss = error "You need to implement this function."
 
 fromString :: Read a => String -> Matrix a
-fromString = error "You need to implement this function."
+fromString xs = error "You need to implement this function."
 
 reshape :: (Int, Int) -> Matrix a -> Matrix a
-reshape = error "You need to implement this function."
+reshape dimension matrix = error "You need to implement this function."
```
rbasso: Any idea here, @exercism/haskell?

exercises/nth-prime/src/Prime.hs
```diff
@@ -1,3 +1,4 @@
 module Prime (nth) where
 
-nth = error "You need to implement this function."
+nth :: Int -> Maybe Int
```
rbasso: The input can be any `Integral` type, but the test suite demands the output to be `Maybe Integer`.

This is possiblywhat makes this PR fail in Travis.

exercises/pascals-triangle/src/Triangle.hs
```diff
@@ -1,4 +1,4 @@
 module Triangle (rows) where
 
 rows :: Int -> [[Integer]]
-rows = error "You need to implement this function."
+rows x = error "You need to implement this function."
```
rbasso: Not sure what is better here, `n` or `x`.

exercises/matrix/src/Matrix.hs
petertseng: try `dimensions` since there the `(Int, Int)` would mean there are two dimensions.

exercises/atbash-cipher/src/Atbash.hs
```diff
@@ -1,7 +1,7 @@
 module Atbash (decode, encode) where
 
 decode :: String -> String
-decode phrase = error "You need to implement this function."
+decode xs = error "You need to implement this function."
```
petertseng: I don't know if it's just me, but I found `phrase` better, since `xs` doesn't really give a meaning to the string. So I think I in general didn't like `xs` as the name for any string. Even though strings are lists.

Then again, maybe the students are meant to find the meaning and rename the args to suit personal taste?

We can't please everyone so we just have to pick one, I suppose. I let y'all decide.

rbasso: `xs` is great when we don't know anything about the content, so there is not way to give more information. Which is not the case.

I think that `phrase` conveys meaning, but not necessarily the correct one:

> [phrase](http://www.thefreedictionary.com/phrase) n.
> A sequence of words that have meaning, especially when forming part of a sentence.

In the context of `decode`, the argument is encoded, so I don't think it fits the definition very sell, and that's why I prefer `xs` over `phrase` here.

Of course, in a cryptography-context, there are more descriptive names, like [plainText](https://en.wikipedia.org/wiki/Plaintext) and [cipherText](cipherText).

abo64: @petertseng I agree with you for this case:
The meaning of the parameter is fixed and cannot vary (as it can for some type `[a]`, say). So for me it makes sense to use a more meaningful name, too.

exercises/list-ops/src/ListOps.hs
```diff
@@ -13,25 +13,25 @@ import Prelude hiding
...
-reverse = error "You need to implement this function."
+reverse xs = error "You need to implement this function."
 
 map :: (a -> b) -> [a] -> [b]
-map = error "You need to implement this function."
+map f xs = error "You need to implement this function."
 
 filter :: (a -> Bool) -> [a] -> [a]
-filter = error "You need to implement this function."
+filter f xs = error "You need to implement this function."
```
lpalma: I'm used to `p` for _predicate_ as well, but thought `f` would make it more clear that this is a function. Happy to change 👍

exercises/luhn/src/Luhn.hs
```diff
@@ -1,11 +1,16 @@
 module Luhn (addends, checkDigit, checksum, create, isValid) where
 
-addends = error "You need to implement this function."
+addends :: Int -> [Int]
+addends n = error "You need to implement this function."
```
lpalma: >Also, I don't think it is a good idea to optimize so early.

not sure what you mean. Should I use `Integer` then, or get rid of the type declaration altogether?

exercises/atbash-cipher/src/Atbash.hs
lpalma: I'm going for `encode plainText` and `decode cipherText` then 👍

exercises/luhn/src/Luhn.hs
rbasso: Yes. You should use `Integer` to avoid causing problems in some platforms. The Haskell standard's minimal range for `Int` is too short for our tests.

lpalma:
I've addressed the feedbacks.
Happy for this PR to be merged if you guys agree 👍

exercises/list-ops/src/ListOps.hs
```diff
@@ -13,25 +13,25 @@ import Prelude hiding
   ( length, reverse, map, filter, foldr, (++), concat )
 
 foldl' :: (b -> a -> b) -> b -> [a] -> b
-foldl' = error "You need to implement this function."
+foldl' f z xs = error "You need to implement this function."
 
 foldr :: (a -> b -> b) -> b -> [a] -> b
-foldr = error "You need to implement this function."
+foldr f x xs = error "You need to implement this function."
```
rbasso: Oh! Sorry.

I didn't explicitly said earlier, but all the *folds* usually use `z`. Also, for consistency with the previous `foldl'`, this could use it too.

petertseng:
The two hard things - cache invalidation, naming things, and off-by-one errors. It was hard for me to come up with a better name for any of the `String` named `xs`, so I will believe we are ready when the `z` is done.

lpalma:
I think it's good to go now 👍

rbasso:
Thanks, @lpalma.

Note to myself: Everything ticked in #473.

PR #507, opened 2017-03-17
Add Explicit Args To Exercises - Part 3

lpalma:
Part of #473 

The [POV](http://exercism.io/exercises/haskell/pov/readme) page is broken. I can open an Issue if that's the case.

### Exercises covered:

- [x] POV
- [x] Pythagorean Triplet
- [x] Queen Attack
- [x] Raindrops
- [x] Robot Name
- [x] Robot Simulator
- [x] Roman Numerals
- [x] Run Length Encoding
- [x] Saddle Points
- [x] Say
- [x] Scrabble Score
- [x] Secret Handshake
- [x] SGF Parsing

### DIFF
--- a/exercises/pov/src/POV.hs
+++ b/exercises/pov/src/POV.hs
@@ -3,7 +3,7 @@ module POV (fromPOV, tracePathBetween) where
 import Data.Tree (Tree)
 
 fromPOV :: Eq a => a -> Tree a -> Maybe (Tree a)
-fromPOV = error "You need to implement this function."
+fromPOV x tree = error "You need to implement this function."
 
 tracePathBetween :: Eq a => a -> a -> Tree a -> Maybe [a]
-tracePathBetween = error "You need to implement this function."
+tracePathBetween from to tree = error "You need to implement this function."
--- a/exercises/pythagorean-triplet/src/Triplet.hs
+++ b/exercises/pythagorean-triplet/src/Triplet.hs
@@ -1,7 +1,10 @@
 module Triplet (isPythagorean, mkTriplet, pythagoreanTriplets) where
 
-isPythagorean = error "You need to implement this function."
+isPythagorean :: (Int, Int, Int) -> Bool
+isPythagorean triplet = error "You need to implement this function."
 
-mkTriplet = error "You need to implement this function."
+mkTriplet :: Int -> Int -> Int -> (Int, Int, Int)
+mkTriplet a b c = error "You need to implement this function."
 
-pythagoreanTriplets = error "You need to implement this function."
+pythagoreanTriplets :: Int -> Int -> [(Int, Int, Int)]
+pythagoreanTriplets minFactor maxFactor = error "You need to implement this function."
--- a/exercises/queen-attack/src/Queens.hs
+++ b/exercises/queen-attack/src/Queens.hs
@@ -4,4 +4,4 @@ boardString :: Maybe (Int, Int) -> Maybe (Int, Int) -> String
 boardString white black = error "You need to implement this function."
 
 canAttack :: (Int, Int) -> (Int, Int) -> Bool
-canAttack = error "You need to implement this function."
+canAttack queenA queenB = error "You need to implement this function."
--- a/exercises/raindrops/src/Raindrops.hs
+++ b/exercises/raindrops/src/Raindrops.hs
@@ -1,3 +1,4 @@
 module Raindrops (convert) where
 
-convert = error "You need to implement this function."
+convert :: Int -> String
+convert n = error "You need to implement this function."
--- a/exercises/robot-name/src/Robot.hs
+++ b/exercises/robot-name/src/Robot.hs
@@ -6,7 +6,7 @@ mkRobot :: IO Robot
 mkRobot = error "You need to implement this function."
 
 resetName :: Robot -> IO ()
-resetName = error "You need to implement this function."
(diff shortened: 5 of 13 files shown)


petertseng:
> The POV page is broken.

Unfortunately that means the canary we added in #454 died, and for the wrong reason.

```
$ ruby -r./lib/trackler -e "Trackler::Tracks.new('.')['haskell'].implementations['pov'].files"
/home/pt/.rvm/gems/ruby-2.4.0/gems/trackler-2.0.8.22/lib/trackler/implementation.rb:30:in `read': No such file or directory @ rb_sysopen - tracks/haskell/exercises/pov/.dummylink (Errno::ENOENT)
```

Guess it would be better to remove it.

At least http://x.exercism.io/v2/exercises/haskell/pov still works, eh?

exercises/pov/src/POV.hs
```diff
@@ -3,7 +3,7 @@ module POV (fromPOV, tracePathBetween) where
 import Data.Tree (Tree)
 
 fromPOV :: Eq a => a -> Tree a -> Maybe (Tree a)
-fromPOV = error "You need to implement this function."
+fromPOV node tree = error "You need to implement this function."
 
 tracePathBetween :: Eq a => a -> a -> Tree a -> Maybe [a]
-tracePathBetween = error "You need to implement this function."
+tracePathBetween nodeA nodeB tree = error "You need to implement this function."
```
petertseng: do we want to make clear which node between the two given should be first in the path?

exercises/pythagorean-triplet/src/Triplet.hs
```diff
@@ -1,7 +1,10 @@
 module Triplet (isPythagorean, mkTriplet, pythagoreanTriplets) where
 
-isPythagorean = error "You need to implement this function."
+isPythagorean :: (Int, Int, Int) -> Bool
```
petertseng: if this is done, then the reasoning in https://github.com/exercism/xhaskell/blob/master/exercises/pythagorean-triplet/.meta/DONT-TEST-STUB for why the tests would fail to build is false, so if we do this please delete that file so that the stub gets tested.

If we choose the type of `mkTriplet` as `(Int, Int, Int)` rather than letting the student choose, do we even need `mkTriplet`? Shouldn't we just delete `mkTriplet` then? I think the only point of it is to give the student the choice.

Why did we give this choice, anyway?

petertseng: > Why did we give this choice, anyway?

I didn't see explanation given in https://github.com/exercism/exercism.io/pull/741/files. I guess we can see that `mkTriplet` is invariant to ordering of the three numbers, since it will always order them in a consistent way.

So we still need `mkTriplet`.

exercises/pov/src/POV.hs
rbasso: In `fromPOV` we only know that the element **may** be in the `Tree`, but `node` suggests two things:

1. It is part of a tree.
2. It is a node, not a leaf.

I would suggest simply `x`, because the meaning is already clear in the function's name.

In `tracePathBetween` we have a similar problem but, as @petertseng mentioned, one is supposed to be the starting element and the other the ending element, so a think a pair of related names would be great. Some suggestions:

- `start`, `end`
- `from`, `to`

exercises/pythagorean-triplet/src/Triplet.hs
rbasso: I think we can consider redesigning this exercise without `mkTriplet`, @petertseng. The way it is now, I think it is over-specified.

exercises/pov/src/POV.hs
lpalma: Happy to make the changes. As the Haskell's POV page is broken, I checked the instructions for [Ruby](http://exercism.io/exercises/ruby/pov/readme), which says:

>This lets us more simply describe the paths between two nodes. So for example the path from 6-9 (which in the first graph goes up to the root and then down to a different leaf node) can be seen to follow the path 6-2-0-3-9
This exercise involves taking an input graph and re-orientating it from the point of view of one of the nodes.
 
Also, isn't a Leaf also considered a _node_ (often called _leaf node_) ?

exercises/pythagorean-triplet/src/Triplet.hs
lpalma: Yeah, I left this exercise for last because I couldn't figure out a way to declare the types without over-specifying it :/

I was completely unaware of https://github.com/exercism/xhaskell/blob/master/exercises/pythagorean-triplet/.meta/DONT-TEST-STUB though

What about get rid of this commit so this exercise can get redesigned as suggested by @rbasso ?

exercises/pov/src/POV.hs
rbasso: True. A leaf is a node. 😄

exercises/pythagorean-triplet/src/Triplet.hs
rbasso: I don't think we should stop things here because of that. We don't have anything ready yet.

lpalma: cool. I've applied the feedback for the POV exercise, so I'm just waiting for the decision here.

petertseng:
> Unfortunately that means the canary we added in #454 died

Er? it works again? http://exercism.io/exercises/haskell/pov/readme

Okay fine guess we can leave the canary in...

PR #511, opened 2017-03-28
Add Pangram Exercise

lpalma:
I think this exercise is fairly trivial to be implemented. Plus, it is good to have more exercises between Bob and Run Length Encoding, since the first is much simpler than the latter.

### DIFF
--- a/exercises/pangram/stack.yaml
+++ b/exercises/pangram/stack.yaml
@@ -1,0 +1,1 @@
+resolver: lts-8.2
--- a/exercises/pangram/test/Tests.hs
+++ b/exercises/pangram/test/Tests.hs
@@ -1,0 +1,65 @@
+{-# LANGUAGE RecordWildCards #-}
+
+import Data.Foldable     (for_)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
+import Pangram (isPangram)
+
+main :: IO ()
+main = hspecWith defaultConfig {configFastFail = True} specs
+
+specs :: Spec
+specs = describe "pangram" $
+          describe "isPangram" $ for_ cases test
+  where
+    test Case{..} = it description $ isPangram input `shouldBe` expected
+
+-- Adapted from
+-- Source: exercism/x-common/exercises/pangram/canonical-data.json
+-- Version: 1.0.0
+-- Date: 2017-03-28.
+
+data Case = Case { description :: String
+                 , input       :: String
+                 , expected    :: Bool
+                 }
+
+cases :: [Case]
+cases = [ Case { description = "sentence empty"
+               , input       = ""
+               , expected    = False
+               }
+        , Case { description = "pangram with only lower case"
+               , input       = "the quick brown fox jumps over the lazy dog"
+               , expected    = True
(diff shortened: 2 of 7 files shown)


exercises/pangram/stack.yaml
```diff
@@ -0,0 +1 @@
+resolver: lts-8.6
```
petertseng: good time to ask now about whether there is a policy:

* Do we want to keep the same version across the same track?
* Or is it acceptable to let each exercise have its own version?

exercises/pangram/test/Tests.hs
```diff
@@ -0,0 +1,64 @@
...
+
+specs :: Spec
+specs = describe "pangram" $
+          describe "isPangram" $ for_ cases test
+  where
+    test Case{..} = it description $ isPangram input `shouldBe` expected
+
+-- Test cases adapted from
+-- `exercism/x-common/exercises/pangram/canonical-data.json`
+-- on 2017-03-28.
```
petertseng: I think at this point since there is a version at https://github.com/exercism/x-common/blob/master/exercises/pangram/canonical-data.json#L6 we should include it somehow, just not sure what format we want. "x-common version: 1.0.0" perhaps?

lpalma: I'm not sure if the structure I used here was explicitly made standard/convention or simply accepted by adapting from other tests (I simply followed/copied from other tests).

That said, now that we have more data to include, I'm going to propose the following:

```
Adapted from
Source: exercism/x-common/exercises/pangram/canonical-data.json
Version: 1.0.0
Date: 2017-03-28
```

`Test cases` seems redundant since that's what the file is about.

exercises/pangram/stack.yaml
lpalma: Ah, I just read the [instructions](https://github.com/exercism/xhaskell#exercise-structure) again and realised I may have misunderstood this bit:

>* `stack.yaml` has just one line specifying the current [Stack snapshot](https://www.stackage.org/snapshots). We use the same resolver for all the exercises.

The first part suggests we should include the current Snapshot from the link given, but the second part says we should use the same resolver for all exercises!!!11one

I'd prefer consistency within the track, so either downgrade this one to `lts-8.2`, or upgrade the others.

Perhaps we could fix the resolver version in a config file within `xhaskell` and then create a bin  in `xhaskell/bin` that will generate the `stack.yaml` for a given path? Starting with `stack.yaml`, this could evolve to a file that would generate the full exercise structure.

petertseng: > the current Stack snapshot

Yes, I guess that is ambiguous. Perhaps "the track's current Snack snapshot" would be more accurate.

> Starting with `stack.yaml`, this could evolve to a file that would generate the full exercise structure.

Nice idea. We could start simple with a _template directory like https://github.com/exercism/xjava/tree/master/exercises/_template or https://github.com/exercism/xkotlin/tree/master/exercises/_template

petertseng: > either downgrade this one to lts-8.2, or upgrade the others

Let's save the upgrade for its own PR that affects all exercises like #500. I think this will be ready to merge.

lpalma: done 👍

PR #513, opened 2017-03-30
Add Bracket Push Exercise

lpalma:
Port [Bracket Push][1] to the Haskell Track.

For the [Example Solution][2], I tried to explore Data constructors and the Maybe type.

[1]: https://github.com/exercism/x-common/tree/master/exercises/bracket-push
[2]: https://github.com/lpalma/xhaskell/blob/94ea00b9a6823a332772436ed6a8a723a30efaef/exercises/bracket-push/examples/success-standard/src/BracketPush.hs

### DIFF
--- a/config.json
+++ b/config.json
@@ -280,6 +280,13 @@
       ]
     },
     {
+      "slug": "bracket-push",
+      "difficulty": 5,
+      "topics": [
+          "Stack"
+      ]
+    },
+    {
       "slug": "crypto-square",
       "difficulty": 5,
       "topics": [
--- a/config.json
+++ b/config.json
@@ -497,7 +497,8 @@
       "difficulty": 10,
       "topics": [
         "Define type",
-        "Either"
+        "Either",
+        "Stack"
       ]
     },
     {
--- a/exercises/bracket-push/stack.yaml
+++ b/exercises/bracket-push/stack.yaml
@@ -1,0 +1,1 @@
+resolver: lts-8.2
--- a/exercises/bracket-push/test/Tests.hs
+++ b/exercises/bracket-push/test/Tests.hs
@@ -1,0 +1,85 @@
+{-# LANGUAGE RecordWildCards #-}
+
+import Data.Foldable     (for_)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
+import Brackets (arePaired)
+
+main :: IO ()
+main = hspecWith defaultConfig {configFastFail = True} specs
+
+specs :: Spec
+specs = describe "bracket-push" $
+          describe "isPaired" $ for_ cases test
+  where
+    test Case{..} = it description $ arePaired input `shouldBe` expected
+
+-- Adapted from
+-- Source: exercism/x-common/exercises/bracket-push/canonical-data.json
+-- Version: 1.1.0
+-- Date: 2017-04-07.
+
+data Case = Case { description :: String
+                 , input       :: String
+                 , expected    :: Bool
+                 }
+
+cases :: [Case]
+cases = [ Case { description = "paired square brackets"
+               , input       = "[]"
+               , expected    = True
+               }
+        , Case { description = "empty string"
+               , input       = ""
+               , expected    = True
+               }
+        , Case { description = "unpaired brackets"
+               , input       = "[["
+               , expected    = False
+               }
+        , Case { description = "wrong ordered brackets"
+               , input       = "}{"
+               , expected    = False
+               }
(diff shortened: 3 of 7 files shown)


config.json
```diff
@@ -110,6 +110,12 @@
       ]
     },
     {
+      "slug": "bracket-push",
+      "difficulty": 3,
```
lpalma: Not 100% sure it is difficulty 3. Perhaps I'm overcomplicating it?

petertseng: It's true that once one realizes one can use a stack, it is easy. So the hard part is not writing the implementation. It is figuring out how to tackle the problem at all.

Because of this, I would actually rate *higher*, maybe 4 or 5.

If it helps to see how other tracks rated it:

`for i in tracks/*/config.json; do echo $i; jq 'if any(.exercises[]; .difficulty != 1) then .exercises | map(select(.slug == "bracket-push")) else [] end' < $i; done`

```
tracks/ceylon/config.json
[
  {
    "slug": "bracket-push",
    "difficulty": 6,
    "topics": [
      "Parsing",
      "Stacks",
      "Strings"
    ]
  }
]

tracks/csharp/config.json
[
  {
    "slug": "bracket-push",
    "difficulty": 5,
    "topics": [
      "Parsing",
      "Strings"
    ]
  }
]

tracks/elixir/config.json
[
  {
    "slug": "bracket-push",
    "difficulty": 3,
    "topics": []
  }
]

tracks/fsharp/config.json
[
  {
    "slug": "bracket-push",
    "difficulty": 7,
    "topics": [
      "Parsing",
      "Strings"
    ]
  }
]

tracks/java/config.json
[
  {
    "slug": "bracket-push",
    "difficulty": 5,
    "topics": []
  }
]

tracks/lua/config.json
[
  {
    "slug": "bracket-push",
    "difficulty": 5,
    "topics": [
      "stacks",
      "strings",
      "algorithms",
      "control-flow (if-else statements)",
      "control-flow (loops)"
    ]
  }
]

tracks/objective-c/config.json
[
  {
    "difficulty": 5,
    "slug": "bracket-push",
    "topics": [
      "Strings",
      "Conditionals",
      "Looping"
    ]
  }
]

tracks/ocaml/config.json
[
  {
    "slug": "bracket-push",
    "difficulty": 4,
    "topics": [
      "Stacks"
    ]
  }
]

tracks/rust/config.json
[
  {
    "slug": "bracket-push",
    "difficulty": 4,
    "topics": [
      "From trait",
      "stack or recursion"
    ]
  }
]

tracks/swift/config.json
[
  {
    "difficulty": 7,
    "slug": "bracket-push",
    "topics": [
      "Parsing",
      "Strings"
    ]
  }
]


```

exercises/bracket-push/examples/success-standard/src/BracketPush.hs
```diff
@@ -0,0 +1,36 @@
+module BracketPush (isPaired) where
```
petertseng: I would just call the module `Brackets`. the `Push` is too much of an implementation detail and I don't even think it should go in the name: https://github.com/exercism/x-common/issues/693

I would maybe call the function `arePaired` so that can read like `Brackets.arePaired`, and this makes sense since it handles multiple brackets not just one pair.

exercises/bracket-push/stack.yaml
```diff
@@ -0,0 +1 @@
+resolver: lts-8.6
```
petertseng: same q as https://github.com/exercism/xhaskell/pull/511#discussion_r109295114 - same across all exercises, or allow to be different?

exercises/bracket-push/test/Tests.hs
```diff
@@ -0,0 +1,80 @@
...
+
+specs :: Spec
+specs = describe "bracket-push" $
+          describe "isPaired" $ for_ cases test
+  where
+    test Case{..} = it description $ isPaired input `shouldBe` expected
+
+-- Test cases adapted from
+-- `exercism/x-common/exercises/bracket-push/canonical-data.json`
+-- on 2017-03-29.
```
petertseng: same q as https://github.com/exercism/xhaskell/pull/511#discussion_r109295131 - how should we put the version in here?

config.json
lpalma: In that case, I'll include `Stack` in the list of topics and increase its difficulty to **5**. I agree it shouldn't go any higher than that.

petertseng: OK.

The only other exercise I can think of that would get `Stack` in its topic list at the same time would be `forth`.

lpalma:
thanks for the feedback @petertseng, they have been addressed.

petertseng:
I think this will be ready if we make the Stack version uniform.

Issue #522, opened 2017-05-06
Add exercise's version to package.yaml and remove redundant information


rbasso:
### Rationale

As [discussed earlier](https://github.com/exercism/xhaskell/issues/416#issuecomment-294103455), it may be a good idea to have the exercise's implementation version in the `package.yaml` files. There are a few reasons to do that:

- It will allow us to version the exercises.
- It will be displayed to the users when running the tests.
- It will allow us to check it against the `canonical-data.json` versions, signaling when there is new test data. #416

###### exercises/leap/package.yaml

```yaml
name: leap
version: 1.0.0.1
...
```

Running the tests would give something like this:

```
leap-1.0.0.1: test (suite: test)


year not divisible by 4: common year
year divisible by 4, not divisible by 100: leap year
year divisible by 100, not divisible by 400: common year
year divisible by 400: leap year

Finished in 0.0001 seconds
4 examples, 0 failures
```

### Suggested changes

- Add the version to the main `package.yaml` of each exercise:
  - All exercises must have a four-digits version: MAJOR.MINOR.PATCH.SERIAL
  - Exercises based on a **non-versioned** `canonical-data.json` should use version **0.9.0.1**. If there is a comment containing the date from the data file used as a reference, it should be added to the YAML file, to ease transition, like this:

    ```yaml
    name: all-your-base
    version: 0.9.0.1 # 2016-06-29
    ````

  - Exercises based on a **versioned** `canonical-data.json` should use its version plus a *serial number*, without any comment.
  - Exercises not following a `canonical-data.json` should use **0.1.0** plus a *serial number*, without any comment.
  - The **serial number** must start at **1** and **always** increase a unit when changed, regardless of the changes in the other version numbers.
  - The `package.yaml` files in the example solutions should not be versioned, because it would be too hard to keep them synchronized.
- Remove redundant data from the test suites:
  - Keep comments describing the divergences from `x-common`.
  - Remove references to the `canonical-data.json` files and version.
  - ~~Remove the test group containing the name of the exercise, if existent, because this information is already displayed with the version when running the test suite.~~ (**edit:** as pointed out by @petertseng, this change doesn't belong here) 👍 
- Update the documentation to include the new versioning system.

### Should we do it?

This will take time and demand some work, so we shouldn't start unless the @exercism/haskell maintainers agree with it.

So... **Objections? Comments?**

**Edit:** Approved!

---

### Progress

- [x] Check if we have a consensus about it.
[...]

PR #523, opened 2017-05-06
all-exercises: Add/move versioning to package.yaml

rbasso:
Move the exercise's versioning information from the test suites to the `package.yaml` files, including sensible default versions when no information is available.

Checking if the contents of the test suite matches the `canonical-data.json` file is not in the scope of this PR.

Related to #522.

### DIFF
--- a/exercises/connect/package.yaml
+++ b/exercises/connect/package.yaml
@@ -1,4 +1,5 @@
 name: connect
+version: 0.9.0.1 # 2016-09-16
 
 dependencies:
   - base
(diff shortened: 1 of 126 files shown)


petertseng:
> Checking if the contents of the test suite matches the `canonical-data.json` file

It will be useful to do that so we can declare ourselves up to date if we are in fact up to date. Any plans to do it? If I do not hear that someone else is doing it sometime, I may start doing it, but it would be a shame if I duplicated work.

petertseng:
... I ask because I just did the same for another track in https://github.com/exercism/xrust/pull/282

rbasso:
It would be great if you could do that, @petertseng! 😄

PR #525, opened 2017-05-08
kindergarten-garden 1.0.0.2: edit descriptions and elide middle students

petertseng:
Descriptions are clearer as to the significance of the garden size, or
the fact that the given student list is non-alphabetical.

The tests of the middle children in the full garden are elided.

The canonical data file was added in https://github.com/exercism/x-common/pull/770

### DIFF
--- a/exercises/kindergarten-garden/test/Tests.hs
+++ b/exercises/kindergarten-garden/test/Tests.hs
@@ -18,21 +18,21 @@ main = hspecWith defaultConfig {configFastFail = True} specs
 specs :: Spec
 specs = describe "kindergarten-garden" $ do
 
-    it "alice tests" $ do
+    it "garden with single student" $ do
 
       let alicePlants = lookupPlants "Alice" . defaultGarden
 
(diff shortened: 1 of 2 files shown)


exercises/kindergarten-garden/test/Tests.hs
```diff
@@ -47,18 +47,10 @@ specs = describe "kindergarten-garden" $ do
 
       plants "Alice"   `shouldBe` [ Violets , Radishes, Violets , Radishes ]
       plants "Bob"     `shouldBe` [ Clover  , Grass   , Clover  , Clover   ]
-      plants "Charlie" `shouldBe` [ Violets , Violets , Clover  , Grass    ]
```
petertseng: we could keep these since we already have them, and just declare that they're Haskell specific. don't really care (but note that this is the only place that many of these names are tested!)

rbasso: I don't care too much about the track-specific cases. If they are important, we should include them in `x-common`, otherwise, I think we should remove them.

PR #535, opened 2017-05-09
roman-numerals 1.0.0.2: add descriptions

petertseng:
https://github.com/exercism/x-common/pull/707

### DIFF
--- a/exercises/roman-numerals/test/Tests.hs
+++ b/exercises/roman-numerals/test/Tests.hs
@@ -17,66 +17,85 @@ specs = describe "roman-numerals" $
 
     test Case{..} = it explanation assertion
       where
-        explanation = show number
+        explanation = show number ++ ": " ++ description
         assertion   = numerals (fromIntegral number) `shouldBe` expected
 
-data Case = Case { number   :: Integer
-                 , expected :: Maybe String
+data Case = Case { description :: String
+                 , number      :: Integer
+                 , expected    :: Maybe String
                  }
 
 cases :: [Case]
-cases = [ Case { number   = 1
-               , expected = Just "I"
+cases = [ Case { description = "1 is a single I"
+               , number      = 1
+               , expected    = Just "I"
                }
-        , Case { number   = 2
-               , expected = Just "II"
+        , Case { description = "2 is two I's"
+               , number      = 2
+               , expected    = Just "II"
                }
-        , Case { number   = 3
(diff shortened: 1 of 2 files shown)


petertseng:
I'm not sure if there was a reasonable way to split the commit apart (whitespace change) to make ti easier. I generated these programmatically.

```ruby
require 'json'

json = JSON.parse(File.read(File.join(__dir__, 'canonical-data.json')))

json['cases'].each { |c|
  puts "        , Case { description = \"#{c['description']}\""
  puts "               , number      = #{c['number']}"
  puts "               , expected    = Just \"#{c['expected']}\""
  puts "               }"
}
```

... has anyone been doing these for any of the other tests?

rbasso:
> ... has anyone been doing these for any of the other tests?

No, but I envy your magic code snippet. 👍

exercises/roman-numerals/test/Tests.hs
```diff
@@ -17,66 +17,85 @@ specs = describe "roman-numerals" $
 
     test Case{..} = it explanation assertion
       where
-        explanation = show number
+        explanation = show number ++ ": " ++ description
```
petertseng: like "27: 20 is two X's"

I wanted to keep the actual number being tested because in some cases the number in the description is not the input. Thought it might have been confusing to have the test output "20 is two X's" when the input is 27

rbasso: Yeah! The `27`'s description is confusing...

I still prefer the [assertEqual solution](https://github.com/exercism/xhaskell/issues/517), instead of changing the test's `description`, because it is shown only when it fails.

Anyway, you solution definitely solves the problem here! 👍

petertseng:
Thanks for the reviews! Octomerging soon, then seeing what to do about pov.

petertseng:
36 minutes from declaration of "Octomerging soon" to actual merge for 5 PRs, so it is not sure whether it is a time savings (depends on how long each Travis build would have taken), but at least it saves me from having to rebase and click buttons all the time.

petertseng:
I still maintain that it's a win.

petertseng:
ebwodp: that's 38 minutes.

petertseng:
I don't expect to have to do that too often though. Only when there are a lot of simultaneous PRs in flight, and since the last one I'm dealing with now is POV, we will be OK.

PR #536, opened 2017-05-09
pov 1.1.1.3: simplify "Can find path from nodes other than x"

petertseng:
This was the main change in https://github.com/exercism/x-common/pull/704

https://github.com/exercism/x-common/pull/710 is an inconsequential
change that does not affect our tests.

Two other changes in 1.1.0 that aren't exactly followed:

* As a part of conforming to the canonical schema, trees were embedded
  into the test data, but we have no need to do it - let's reduce
  duplication, and besides we are safe from accidental mutation.
* The "cannot trace" cases were changed to use simpler trees than
  cousins, but we would have had to introduce another tree to do that
  (one with kids and siblings). We'll stick with cousins.

### DIFF
--- a/exercises/pov/package.yaml
+++ b/exercises/pov/package.yaml
@@ -1,5 +1,5 @@
 name: pov
-version: 1.0.0.2
+version: 1.1.1.3
 
 dependencies:
   - base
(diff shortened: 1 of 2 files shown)


petertseng:
I'm honestly not sure about the decisions I made, so take a look.

PR #540, opened 2017-05-09
all: Remove test group with the exercise's name

rbasso:
The exercise's name is already displayed when running the test suite, so there is no need to have a test group just to display it.

This PR remove those labels that I, unfortunately, added to all exercises when porting the exercises to use `hspec`.

### DIFF
--- a/exercises/clock/package.yaml
+++ b/exercises/clock/package.yaml
@@ -1,5 +1,5 @@
 name: clock
-version: 1.0.1.2
+version: 1.0.1.3
 
 dependencies:
   - base
--- a/exercises/clock/test/Tests.hs
+++ b/exercises/clock/test/Tests.hs
@@ -10,7 +10,7 @@ main :: IO ()
 main = hspecWith defaultConfig {configFastFail = True} specs
 
 specs :: Spec
-specs = describe "clock" $ do
+specs = do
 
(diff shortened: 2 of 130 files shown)


petertseng:
Before I review this! See how the test version was not updated as they were in #539!

I don't actually care whether they get updated, but it is your opportunity to update them before I get too far in review. Several hours before I review.

rbasso:
I guess it is fixed now, @petertseng!

I updated the versions of all exercises - except for `change` and `luhn` - that where not changed here. The version changes where added to each individual commit.

Everything was edited by hand, but the commits where generated using a loop, so they should have identical messages, except for the exercise's name.

To help in the review, here are the totals:

- We have 81 exercises. 79 where changed in this PR. `change` and `luhn` where changed previously.
- There should be two files changed in each commit: `package.yaml` and `test/Tests.hs`. They where added to the commits in a loop, so I expect not to have committed changes to other files.
- In some exercises I had to remove the `import` of `describe` to avoid problems with `-Wall`.

petertseng:
I am sorry to say... you said "Remote" in all the commit messages (probably meant "Remove")

rbasso:
Hahahaha! Don't be sorry!

Nice catch, @petertseng!

Fixed!

PR #542, opened 2017-05-11
parallel-letter-frequency: Add benchmark

unmanbearpig:
Fixes https://github.com/exercism/xhaskell/issues/519

I've added a benchmark test that makes sure that `frequency` function runs at least 10% faster (arbitrary number) with multiple workers that with a single worker.

It increases the time it takes to run the tests by about 5 seconds on my machine, not sure if it's acceptable. It might be possible to make it faster by tweaking Criterion settings.

Another concern is the benchmark test would only work correctly on a machine with at least 2 cores.

It prints the benchmark results into the console:

```
frequency
  no texts mean no letters
  one letter
  case insensitivity
  many empty texts still mean no letters
  many times the same text gives a predictable result
  punctuation doesn't count
  numbers don't count
  all three anthems, together, 1 worker
  all three anthems, together, 4 workers
Run with a single worker
benchmarking...
time                 4.487 ms   (4.398 ms .. 4.544 ms)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 4.566 ms   (4.557 ms .. 4.577 ms)
std dev              17.05 μs   (11.71 μs .. 25.01 μs)

Run with 4 workers
benchmarking...
time                 1.425 ms   (1.345 ms .. 1.490 ms)
                     0.977 R²   (0.951 R² .. 0.992 R²)
mean                 1.556 ms   (1.539 ms .. 1.585 ms)
std dev              50.19 μs   (33.35 μs .. 71.29 μs)
variance introduced by outliers: 13% (moderately inflated)

  multiple workers run faster than 1

Finished in 4.9143 seconds
10 examples, 0 failures

```

I guess another option would be to add a separate benchmark and have users check the speed themselves, not sure which option is better. If it looks good I'm going to add some information about it into `HINTS.md`. 

Maybe we could make it optional somehow?

### DIFF
--- a/exercises/parallel-letter-frequency/examples/success-standard/package.yaml
+++ b/exercises/parallel-letter-frequency/examples/success-standard/package.yaml
@@ -1,7 +1,5 @@
 name: parallel-letter-frequency
 
-ghc-options: -threaded -with-rtsopts=-N -O2
-
 dependencies:
   - base
   - containers
@@ -23,6 +21,8 @@ tests:
 
 benchmarks:
   bench:
+    ghc-options: -threaded -with-rtsopts=-N -O2
+
     main: Benchmark.hs
     source-dirs: bench
     dependencies:
--- a/exercises/parallel-letter-frequency/examples/success-standard/package.yaml
+++ b/exercises/parallel-letter-frequency/examples/success-standard/package.yaml
@@ -22,7 +22,7 @@ tests:
 
 benchmarks:
   bench:
-    main: Benchmarks.hs
+    main: Benchmark.hs
     source-dirs: bench
     dependencies:
       - parallel-letter-frequency
--- a/exercises/parallel-letter-frequency/examples/success-standard/package.yaml
+++ b/exercises/parallel-letter-frequency/examples/success-standard/package.yaml
@@ -1,4 +1,5 @@
 name: parallel-letter-frequency
+version: 0.1.0.3
 
 dependencies:
   - base
@@ -18,3 +19,11 @@ tests:
     dependencies:
       - parallel-letter-frequency
       - hspec
+
+benchmarks:
+  bench:
+    main: Benchmarks.hs
+    source-dirs: bench
+    dependencies:
+      - parallel-letter-frequency
+      - criterion
--- a/exercises/parallel-letter-frequency/examples/success-standard/package.yaml
+++ b/exercises/parallel-letter-frequency/examples/success-standard/package.yaml
@@ -1,5 +1,7 @@
 name: parallel-letter-frequency
 
+ghc-options: -threaded -rtsopts -O2
+
 dependencies:
   - base
   - containers
--- a/exercises/parallel-letter-frequency/examples/success-standard/package.yaml
+++ b/exercises/parallel-letter-frequency/examples/success-standard/package.yaml
@@ -1,5 +1,4 @@
 name: parallel-letter-frequency
-version: 0.1.0.3
 
 dependencies:
   - base
--- a/exercises/parallel-letter-frequency/examples/success-standard/package.yaml
+++ b/exercises/parallel-letter-frequency/examples/success-standard/package.yaml
@@ -1,6 +1,6 @@
 name: parallel-letter-frequency
 
-ghc-options: -threaded -rtsopts -O2
+ghc-options: -threaded -with-rtsopts=-N -O2
 
 dependencies:
   - base
--- a/exercises/parallel-letter-frequency/package.yaml
+++ b/exercises/parallel-letter-frequency/package.yaml
@@ -1,6 +1,8 @@
 name: parallel-letter-frequency
 version: 0.1.0.3
 
+ghc-options: -threaded -rtsopts -O2
+
 dependencies:
   - base
   - containers
--- a/exercises/parallel-letter-frequency/package.yaml
+++ b/exercises/parallel-letter-frequency/package.yaml
@@ -1,7 +1,7 @@
 name: parallel-letter-frequency
 version: 0.1.0.3
 
-ghc-options: -threaded -rtsopts -O2
+ghc-options: -threaded -with-rtsopts=-N -O2
 
 dependencies:
   - base
--- a/exercises/parallel-letter-frequency/package.yaml
+++ b/exercises/parallel-letter-frequency/package.yaml
@@ -1,5 +1,5 @@
 name: parallel-letter-frequency
-version: 0.1.0.2
+version: 0.1.0.3
 
 dependencies:
   - base
@@ -20,3 +20,11 @@ tests:
     dependencies:
       - parallel-letter-frequency
       - hspec
+
+benchmarks:
+  bench:
+    main: Benchmarks.hs
+    source-dirs: bench
+    dependencies:
+      - parallel-letter-frequency
+      - criterion
--- a/exercises/parallel-letter-frequency/package.yaml
+++ b/exercises/parallel-letter-frequency/package.yaml
@@ -23,7 +23,7 @@ tests:
 
 benchmarks:
   bench:
-    main: Benchmarks.hs
+    main: Benchmark.hs
     source-dirs: bench
     dependencies:
       - parallel-letter-frequency
--- a/exercises/parallel-letter-frequency/package.yaml
+++ b/exercises/parallel-letter-frequency/package.yaml
@@ -1,8 +1,6 @@
 name: parallel-letter-frequency
 version: 0.1.0.3
 
-ghc-options: -threaded -with-rtsopts=-N -O2
-
 dependencies:
   - base
   - containers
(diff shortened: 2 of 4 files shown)


unmanbearpig:
I'm not certain on whether I should use `nf` or `whnf` here https://github.com/exercism/xhaskell/pull/542/files#diff-63fa4ff5bfd229a3aed31a4232a2f3f8R110

rbasso:
Thanks for the PR!

I think we shouldn't make here a clear distinction about what belongs to tests and what belongs to benchmarks. I usually divide like this:

- Tests give us a reasonable assurance that our functions have the correct behavior.
- Benchmarks give feedback about how fast the code runs.

> I guess another option would be to add a separate benchmark and have users check the speed themselves, not sure which option is better.

I don't knows what the other maintainers think about it, but I definitely think it would be a better solution.

> Maybe we could make it optional somehow?

If it where not inside the test suite, but in a separate benchmark suite in `package.yaml`, it would be automatically optional.

Because we still need to decide where does the benchmark code belongs, I'll wait for the other maintainers position and not review the benchmark code a the moment.

unmanbearpig:
@rbasso, I get what you're saying.  My reasoning for having it as a test instead of a separate benchmark is: since the exercise is to write parallel code then tests should check if it runs in parallel. Sequential code shouldn't pass the tests, because users might think that their solution is correct when it actually is not. A benchmark is the only way to test it in Haskell, as far as I know.
It might be more straightforward for users, as they wouldn't have to think about running a separate `stack bench` command. Instead they would be able to keep running `stack test` until it passes, like in any other exercise. Seems like a bit less friction and slightly better user experience. 

I agree that we should wait for other people's opinions on it.

petertseng:
For any other exercise I would suggest the benchmarks go in `stack bench`. Convince me why it should be otherwise here! Ah, because the exercise is **specifically** about parallelism? That does in fact make the exercise slightly unique. I have noted it as such in x-common.

rbasso:
> Another concern is the benchmark test would only work correctly on a machine with at least 2 cores.

This is really problematic, because there are multiples situations where the tests could be run on a single core:

- Users with core machines (uncommon nowadays)
- Development environment inside a virtual machine (some developers like that)
- Continuous Integration in single core machines (I guess that Travis-CI gives us two cores ATM)

Even if the number of cores where not a problem, I think that testing for speed improvements misses the point:

- It is in general impossible, because parallel execution isn't a property of the code. Haskell is non-strict, so it is up to the compiler and run-time system to decide when/how to evaluate expressions.
- Testing speed is a proxy for testing parallel execution, not the same thing.
- A busy host/guest OS could make a normally faster parallel benchmark be scored as slower than the non-parallel version. 
- A slower, parallel solution is correct IMHO, so it shouldn't be refused in any circumstance.
- It is a really smart hack to measure the variation in speed, but hacks tend to compose in exponentially harder to manage projects.

While I'm willing to accept a benchmark suite, I strongly oppose using the tests suites to check for speed.

exercises/parallel-letter-frequency/test/Tests.hs
```diff
@@ -80,3 +86,25 @@ specs = do
...
+        putStrLn $ "Run with " ++ show workers ++ " workers"
+        multiWorkerMetric <- measuredMetric <$> runBench workers
+
+        (singleWorkerMetric / multiWorkerMetric) `shouldSatisfy` (> 1.1)
+
+measuredMetric :: Report -> Double
+measuredMetric = B.estPoint . anMean . reportAnalysis
+
+makeBench :: [Text] -> Int -> Benchmarkable
+makeBench texts n = nf (frequency n) texts -- am I right that I should use nf and not whnf?
```
rbasso: `nf` seems preferable, but it accepts a smaller set of types than `whnf`.

In this case, because the returned value is a lazy `Map`, I guess that using `whfn` could give artificially lower and incorrect benchmark values.

I didn't tested though.

unmanbearpig:
I've added a benchmark with cases for running it on 4 and 500 anthems. Each number of anthems would run with a single worker and physical-number-of-cores number of workers. All cases run first on a single core and then on all available cores, as I found that results are often different.

It runs about 45 seconds on my machine. It's probably too many cases, but I'm not sure which we cases could get rid of.

Here is the results of the sample solution: https://gist.github.com/unmanbearpig/124297ee38cc35117205679895924c87 and here is my solution: https://gist.github.com/unmanbearpig/fbe706e99b450d313b5bd160a79bcb65 (I've added a case with 256 workers as it's sometimes faster even than the best case of the sample solution)

rbasso:
Just as a reference to anyone interested in reviewing this or another PR with benchmarks, the following commands may help:

###### Clean the cache

```
stack clean
```

###### Build everything with `-Wall -Werror`

```
stack build --pedantic --test --no-run-tests --bench --no-run-benchmarks
```

###### Run the tests

```
stack test
```

###### Run the benchmarks and generate a HTML report

```
stack bench --benchmark-arguments '--output=benchmark.html'
```

If you are using *Stack* with *Nix*, you may also need to add `--no-nix-pure` to the last command.

bin/test-example
```diff
@@ -50,7 +58,9 @@ if [ "$exampletype" = "success" ]; then
     # Implicit exit value: this is last thing in this path,
     # so the script's exit value = stack's exit value.
     # If statements are reordered, please preserve this property.
-    runstack "test"
+    if ! runstack "test"; then
+        exit 1
+    fi
```
rbasso: @petertseng, could you please take a look at these changes in the test scripts when you get some time. I have some kind fever right now, so my brain isn't 100%, and I don't want to merge anything that could break the tests scripts you wrote.

petertseng: Well, it has preserved the intent, but in that case the three lines above **need to be removed** since now it's no longer an implicit exit value.

exercises/parallel-letter-frequency/examples/success-standard/package.yaml
```diff
@@ -1,4 +1,5 @@
 name: parallel-letter-frequency
+version: 0.1.0.3
```
petertseng: I recommend not adding it. You never know what might look upon this as an example. So we should be consistent if we're not adding it.

bin/test-example
petertseng: > three lines above

the three lines being mentioned are the comments - the comments must be removed.

petertseng: I think I accidentally removed this, sorry.

bin/test-example
```diff
@@ -50,7 +58,9 @@ if [ "$exampletype" = "success" ]; then
     # Implicit exit value: this is last thing in this path,
```
petertseng: these comments now must be removed, since they are otherwise inaccurate.

unmanbearpig:
I think I've fixed all of the issues. I've also added a note about the benchmark to `HINTS.md`. What do you think?

rbasso:
I met a related problem with Travis-CI while trying to evaluate this PR, so we'll have to wait for #550.

petertseng:
I'm personally thinking things are OK. At least I don't see anything that I would ask to change. Haven't run it yet to try it out, that'll have to happen tomorrow. But because of this, no need to wait for me if someone beats me to it.

exercises/parallel-letter-frequency/examples/success-standard/package.yaml
```diff
@@ -1,5 +1,7 @@
 name: parallel-letter-frequency
 
+ghc-options: -threaded -with-rtsopts=-N -O2
```
rbasso: Vide comment for `package.yaml`.

exercises/parallel-letter-frequency/package.yaml
```diff
@@ -1,5 +1,7 @@
 name: parallel-letter-frequency
-version: 0.1.0.2
+version: 0.1.0.3
+
+ghc-options: -threaded -with-rtsopts=-N -O2
```
rbasso: This lines causes the following warnings:

```
Configuring parallel-letter-frequency-0.1.0.3...
Warning: 'ghc-options: -threaded' has no effect for libraries. It should only be used for executables.
Warning: 'ghc-options: -with-rtsopts' has no effect for libraries. It should only be used for executables.
```

Besides the warnings, the only real side effect of this is enabling `-threaded -with-rtsopts=-N -O2` for the test suite, which doesn't need it, but the slower compilation from `-O2` probably has an insignificant effect in the overall building time.

unmanbearpig:
@rbasso, I've just rebased it on top of master, deleted my script changes, and moved `ghc-options` into `bench` section. 
Should I clean up and squash some commits before we do the merge into master or should we keep the history as it is?

rbasso:
> Should I clean up and squash some commits before we do the merge into master or should we keep the history as it is?

We usually don't care about the PRs history, and leave only a few commits to make the repository's history nice and clean. In this case, I guess all the changes belong together, so a single commit seems a good idea.

To avoid giving you more trouble, I'll squash merge it here. 👍 

Thanks a lot!

PR #543, opened 2017-05-12
matrix: Remove string tests

### LINKED ISSUE #541: I think the Matrix problem would be better without the 'string' tests.
As a beginner-intermediate Haskeller making my way through the exercises, this was the first time I had to submit an incomplete result and look up the answer. When I looked at it I didn't understand it, and I'm still not sure I understand it. [Basically, I think the type inference is coming from the other side of the `shouldBe` in the tests, is that right?] I never would have figured this out for myself.

I also don't feel it makes much sense in the context of the problem. The direct text of the problem is "Given a string representing a matrix of numbers, return the rows and columns of that matrix." Expecting that to work with a matrix of strings is not reasonable, particularly given that it's a much more difficult parsing problem than anything seen so far.

rbasso:
~~It was argued in #541 that the the exercise would be more interesting without the `fromString` function, which is not an essential part of of it.~~

- ~~Rewrite test suite to use `[[Int]]` instead of `String`.~~
- ~~Remove test that wouldn't make sense without `fromString`.~~
- ~~Remove all other references to `fromString`.~~
- ~~Fix `HINTS.md` to remove reference to invalid characters.~~

---

**Edit**:

The scope of this PR was restricted to just removing the last three tests, based on the discussion below. If in the future we decided to completely remove `fromString` from the test suite, `description.md` will demand some rewriting.

I think that it is preferable to completely rewrite or remove this exercise. The way it is now is lacks focus and also exposes users to a really unidiomatic interface.

---

Closes #541.

### DIFF
--- a/exercises/matrix/test/Tests.hs
+++ b/exercises/matrix/test/Tests.hs
@@ -69,16 +69,3 @@ specs = do
 
     it "flatten" $
       flatten (intMatrix "1 2\n3 4") `shouldBe` vector [1, 2, 3, 4]
-
-    it "matrix of chars" $
-      fromString "'f' 'o' 'o'\n'b' 'a' 'r'" `shouldBe` fromList ["foo", "bar"]
-
-    it "matrix of strings" $ 
-       fromString "\"this one\"\n\"may be tricky!\""
-       `shouldBe` fromList [ ["this one"      ]
-                           , ["may be tricky!"] ]
-
-    it "matrix of strings 2" $ 
-       fromString "\"this one\" \"one\" \n\"may be tricky!\" \"really tricky\""
-       `shouldBe` fromList [ ["this one"      , "one"          ]
-                           , ["may be tricky!", "really tricky"] ]


petertseng:
I drew a different conclusion from #541 - I thought that the parsing (in particular, parsing a matrix of ints) was okay, it was simply the tests of **parsing a matrix of strings** (the two that say "may be tricky") that should have been removed. The matrix of chars probably could go too.

The https://github.com/exercism/x-common/blob/master/exercises/matrix/description.md **does** mention "given a string with embedded newlines", which does imply that the string parsing was an intended part... just probably not matrices containing non-numbers.

It is true that I wouldn't mind removing the parsing completely (there is already a lot in this exercise) and leaving it for another exercise - please indicate if that is an intended route.

rbasso:
Oh! When I was reading the exercise I thought that having `fromString` and representing all matrices as strings in the test suite was so ugly in Haskell that I almost automatically assumed that the complaint was about it.

Right now I'm a little busy, but I can rewrite the PR in a in a day or two.

rbasso:
Oh! When I was reading the exercise I thought that having `fromString` and representing all matrices as strings in the test suite was so ugly in Haskell that I almost automatically assumed that the complaint was about it.

Right now I'm a little busy, but I can rewrite the PR in a in a day or two.

If anyone want to fix it before that, fell free to open another PR and close this one.

rbasso:
PR Updated to just remove the two final tests.

rbasso:
PR Updated to just remove the three final tests.

rbasso:
> you commented saying remove the final three, but removed only the final two. 

Nice catch, @petertseng. Thanks for checking that!

I guess I forgot to push after deciding to remove the third one.

> however, the commit contents are consistent with the commit message (removing tricky string). intentional?

I considered that the third one was also in the trick, so I'm keeping the description. Ok?

petertseng:
We forgot to increment the test version (the fourth component)

PR #552, opened 2017-05-25
travis: Build exercise's benchmarks

rbasso:
- Enable building benchmarks when testing the exercises.

While it is undesirable to run benchmarks as part of the tests, checking that the stubs and example solutions can compile with them is needed to assure that the user will be able to run the benchmarks.

### DIFF
--- a/bin/test-example
+++ b/bin/test-example
@@ -22,7 +22,13 @@ cleanup() {
 }
 trap cleanup EXIT INT TERM
 
-cp -R -L ${exercisedir}/stack.yaml ${exercisedir}/test ${exampledir}/* "${buildfolder}"
+cp -R -L                      \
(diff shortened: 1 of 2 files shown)


rbasso:
> seems OK I suppose. since there are no bench now, is this to support #542 ? Or something else?

The solution in #542 had problems, so I decided to separate in another PR. I'll explain in more details there when reviewing the latest changes.

> Can/should it be tested in conjunction with #542?

It will have to be rebased after this one is merged.

I used [a branch in my fork](https://github.com/rbasso/xhaskell/tree/leap-benchmark) to test everything together.

> It's a shame that we'll no longer error on the cp if, say, a stack.yaml is missing, but I guess it's OK that the error happens later.

I completely agree! If you find a better alternative we can certainly update it.

PR #558, opened 2017-06-10
Normalize format of config file

kytrinyx:
This will let us script changes to it with less noise in the diffs.

### DIFF
--- a/config.json
+++ b/config.json
@@ -289,7 +289,7 @@
       "slug": "bracket-push",
       "difficulty": 5,
       "topics": [
-          "Stack"
+        "Stack"
       ]
     },
     {


petertseng:
It will not offend me if this gets merged, but in anticipation of it offending others, I'll discuss it.

It is good for actions to be scriptable without causing extra noise.

We'd have to be care about adding these extra newlines if we would have to remove them in the future if these problems get topics. That, too, is noise.

There is the possibility of: normalise, generate diff, normalise + apply desired change (remove `ignored` key), apply normalisation diff in reverse to get just the desired change (remove `ignored`) but it seems too much to ask others to do this.

I'm not bothered by the solution of "I just remove the whitespace noise from the scripted changes". Given my current estimated frequency of the changes I think this remains sustainable.

Probably don't need to spend too long thinking about this since it's just a few newlines. If nobody cares, should just merge.

kytrinyx:
First: I don't feel strongly either way, but I *really* want to understand the pros/cons!

> We'd have to be care about adding these extra newlines if we would have to remove them in the future if these problems get topics. That, too, is noise.

Would you elaborate? I'm not sure I understand what you mean.

petertseng:
> >  We'd have to be care about adding these extra newlines if we would have to remove them in the future if these problems get topics. That, too, is noise.
>
> Would you elaborate? I'm not sure I understand what you mean.

Ah right. See how an empty line was added between the opening and closing bracket of, say, hello-world:

```
    {
      "slug": "hello-world",
      "difficulty": 1,
      "topics": [

      ]
},
```

Maybe one day hello-world would have a topic:

```
    {
      "slug": "hello-world",
      "difficulty": 1,
      "topics": [
        "Some topic"
      ]
},
```

The diff would involve removing the empty line and adding `"Some topic"`, so we've gone +1, -1, +1.

If the new line were not added, the diff would simply be adding `"Some topic"`, just a +1.

NobbZ:
You will have similar noise, whenever you append an item to the list
instead of prepending or random insert.

Peter Tseng <notifications@github.com> schrieb am Sa., 10. Juni 2017 um
23:20 Uhr:

kytrinyx:
Let me clarify what I mean by _noise_:

Imagine that I am removing the "ignored" key from the config.json, and because I generated the change, every single empty `topics` key for every single entry in the `exercises` array gets changed. Here's an example: https://github.com/exercism/xperl6/pull/170/files

These changes have nothing to do with the removal of the "ignored" key. That qualifies as noise to me.

In the case of adding topics to hello-world, I don't qualify it as noise, even though there are whitespace changes that perhaps would not have happened in the case of keeping the square brackets on the same line. This is because the whitespace changes are legitimately related to the change in hello-world. 

Likewise, when adding something to the end of an array, there will always be a bit of additional change, adding a comma to the item that was previously the last. This doesn't feel like noise to me, as the change is directly related to the addition of the item.

Does that make sense?

petertseng:
Oh! Okay. That makes sense. So it means the definition is just about *unrelated* changes, as opposed to the more emcompassing definition of *non-minimal* changes. I think? Maybe not the best words. No value judgments here.

So to use these terms, I guess the question is whether we wish to accept the non-minimal changes (one-time) in order to prevent unrelated changes on all auto-generated commits. For me the answer is: As long as I'm willing to personally pay the costs of removing unrelated changes (as I did in #557), I do wish to avoid non-minimal changes.

kytrinyx:
> So it means the definition is just about unrelated changes, as opposed to the more emcompassing definition of non-minimal changes.

Yes! That's definitely what I was going for.

kytrinyx:
> As long as I'm willing to personally pay the costs of removing unrelated changes (as I did in #557), I do wish to avoid non-minimal changes.

If we were able to tell the generator to keep `[]` on a single line, how to indent, and what types of line endings to use, then we wouldn't have this problem, I think. Not that indentation and line endings are a problem in this repository. They were in other repositories.

kytrinyx:
Nah, no trouble :)

Thanks @petertseng

PR #561, opened 2017-06-18
Remove leading 'x' from repository name

kytrinyx:
The leading 'x' is kind of arbitrary. Especially now that we can set
topics on the repositories, we don't need a pattern to distinguish what
is a track or not.

The repository itself has already been renamed. GitHub redirects from
the old name to the new name, so we do not have to rush to fix links to
the old repository name, though we should update them for the sake of
clarity.

See exercism/meta#1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# xHaskell
+# Exercism Haskell Track
 
-[![Build Status](https://travis-ci.org/exercism/xhaskell.png?branch=master)](https://travis-ci.org/exercism/xhaskell)
+[![Build Status](https://travis-ci.org/exercism/haskell.png?branch=master)](https://travis-ci.org/exercism/haskell)
 
 Exercism exercises in Haskell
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -107,10 +107,10 @@ format that has all dependencies and build instructions for an exercise.
 - `test/Tests.hs` is the [test suite](#test-suite).
 
 ### Writing an issue
-To report a bug you should [create an issue](https://help.github.com/articles/creating-an-issue/) [here](https://github.com/exercism/xhaskell/issues).
+To report a bug you should [create an issue](https://help.github.com/articles/creating-an-issue/) [here](https://github.com/exercism/haskell/issues).
 
 ### Writing a pull request
-To fix a bug you should [create a pull request from a fork](https://help.github.com/articles/creating-a-pull-request-from-a-fork/) [here](https://github.com/exercism/xhaskell/pulls). See also [here](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#git-basics) for more information.
+To fix a bug you should [create a pull request from a fork](https://help.github.com/articles/creating-a-pull-request-from-a-fork/) [here](https://github.com/exercism/haskell/pulls). See also [here](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#git-basics) for more information.
```
petertseng: in a separate PR:

* I want a different text than "here".
* should link to docs repo instead.

README.md
```diff
@@ -107,10 +107,10 @@ format that has all dependencies and build instructions for an exercise.
 - `test/Tests.hs` is the [test suite](#test-suite).
 
 ### Writing an issue
-To report a bug you should [create an issue](https://help.github.com/articles/creating-an-issue/) [here](https://github.com/exercism/xhaskell/issues).
+To report a bug you should [create an issue](https://help.github.com/articles/creating-an-issue/) [here](https://github.com/exercism/haskell/issues).
```
petertseng: in a separate PR: I want a different text than "here".

PR #564, opened 2017-06-21
spiral-matrix: Add exercise

rbasso:
This is the implementation of the exercise as initially proposed in exercism/x-common#840.

I'm not sure about the difficulty, so I inserted it just after `food-chain`, for contrast.

~~We shouldn't merge this before `x-common`, as this exercises references the `1.0.0` version of the test cases. I'm just opening this PR so that we can have fix major problems beforehand.~~

### DIFF
--- a/config.json
+++ b/config.json
@@ -346,6 +346,12 @@
       ]
     },
     {
+      "slug": "spiral-matrix",
+      "difficulty": 5,
+      "topics": [
+      ]
+    },
+    {
       "slug": "custom-set",
       "difficulty": 5,
       "topics": [
--- a/exercises/spiral-matrix/examples/success-standard/src/Spiral.hs
+++ b/exercises/spiral-matrix/examples/success-standard/src/Spiral.hs
@@ -1,0 +1,17 @@
+module Spiral (spiral) where
+
+-- | Create a square matrix of natural numbers in a inward, clockwise spiral.
+spiral :: Int -> [[Int]]
+spiral l = matrix (spiralIndex (l, l)) (l, l)
+  where
+    -- If an index isn't in the first row of a matrix, it is in the
+    -- 90 degress-rotated, complementary matrix.
+    spiralIndex :: Integral a => (a, a) -> (a, a) -> a
+    spiralIndex     _  (1, j) = j
+    spiralIndex (h, w) (i, j) = w + spiralIndex (w, h - 1) (w - j + 1, i - 1)
+
+-- Build a lazy, list-based matrix based on a formula.
+matrix :: ((Int, Int) -> a) -> (Int, Int) -> [[a]]
+matrix f (rows, columns) = [[ f (i, j)
+                            | j <- [1..columns] ]
+                            | i <- [1..rows   ] ]
--- a/exercises/spiral-matrix/test/Tests.hs
+++ b/exercises/spiral-matrix/test/Tests.hs
@@ -1,0 +1,40 @@
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
+
+import Spiral (spiral)
+
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
+main :: IO ()
+main = hspecWith defaultConfig {configFastFail = True} specs
+
+specs :: Spec
+specs = describe "spiral" $ do
+
+  it "empty spiral" $
+    spiral 0 `shouldBe` []
+
+  it "trivial spiral" $
+    spiral 1 `shouldBe` [ [1] ]
+
+  it "spiral of side length 2" $
+    spiral 2 `shouldBe` [ [1, 2]
+                        , [4, 3] ]
+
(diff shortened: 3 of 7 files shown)


config.json
```diff
@@ -346,6 +346,12 @@
       ]
     },
     {
+      "slug": "spiral-matrix",
+      "difficulty": 5,
```
abo64: Interesting exercise!
Maybe difficulty 5 could be a bit too high if you consider a possible simple recursive solution with shrinking squares? (although your sample solution is of course much more interesting from a theoretical point of view).

rbasso: > Maybe difficulty 5 could be a bit too high...

My reasoning was that is seems more difficult than `anagram` and `series`, but  is easier than `bank-account`.

Also, I don't think that exercises should be ordered just based on difficulty, but also to alternate in types, making the experience more enjoyable, so I put it after `food-chain`.

If you have another place/difficulty in mind, please tell me.

> ...if you consider a possible simple recursive solution with shrinking squares? (although your sample solution is of course much more interesting from a theoretical point of view).

I'll curious about how people will approach this exercise...

abo64: Then I agree on both the difficulty and the position according to fun factor. :)

`bank-account` is another matter:  I was never too happy with its current form as it enforces a non-FP solution (although one should probably also know how to code such side effects in Haskell, too). But that is another discussion.

exercises/spiral-matrix/examples/success-standard/src/Spiral.hs
```diff
@@ -0,0 +1,17 @@
+module Spiral (spiral) where
+
+-- | Create a square matrix of natural numbers in a inward, clockwise spiral.
+spiral :: Int -> [[Int]]
+spiral l = matrix (spiralIndex (l, l)) (l,l)
```
petertseng: why no space in the second `(l,l)` but a space in the first `(l, l)`?

exercises/spiral-matrix/test/Tests.hs
```diff
@@ -0,0 +1,40 @@
...
+specs :: Spec
+specs = describe "spiral" $ do
+
+  it "empty spiral" $
+    spiral 0 `shouldBe` []
+
+  it "trivial spiral" $
+    spiral 1 `shouldBe` [ [1] ]
+
+  it "spiral of side 2" $
```
petertseng: side or size? (1/4)

exercises/spiral-matrix/test/Tests.hs
```diff
@@ -0,0 +1,40 @@
...
+    spiral 2 `shouldBe` [ [1, 2]
+                        , [4, 3] ]
+
+  it "spiral of side 3" $
```
petertseng: side or size? (2/4)

exercises/spiral-matrix/test/Tests.hs
```diff
@@ -0,0 +1,40 @@
...
+    spiral 3 `shouldBe` [ [1, 2, 3]
+                        , [8, 9, 4]
+                        , [7, 6, 5] ]
+
+  it "spiral of side 4" $
```
petertseng: side or size? (3/4)

exercises/spiral-matrix/test/Tests.hs
```diff
@@ -0,0 +1,40 @@
...
+    spiral 4 `shouldBe` [ [ 1,  2,  3, 4]
+                        , [12, 13, 14, 5]
+                        , [11, 16, 15, 6]
+                        , [10,  9,  8, 7] ]
+
+  it "spiral of side 5" $
```
petertseng: side or size? (4/4)

exercises/spiral-matrix/test/Tests.hs
rbasso: I used `side` because I imagined that `size` could be confused with the number of elements. Which one do you think is better?

- `spiral of size 2`
- `spiral of side 2`
- `spiral of 2x2`
- `spiral of size 2x2`

Any other suggestion?

exercises/spiral-matrix/examples/success-standard/src/Spiral.hs
rbasso: Nice catch!

exercises/spiral-matrix/test/Tests.hs
petertseng: Oh, I guess you might want to change it in problem-specifications too, then.

If just saying its size, maybe something like `spiral of side length 2`?

I was going to try to give some thought to "what shall the larger sizes test that smaller sizes don't" and I might have thought of some answer like "size 3 has a middle-of-edge cell", "size 4 has multiple middle-of-edge cells", "size 5 is the smallest matrix that turns back to the starting direction a second time" for those solutions that are fundamentally going from number -> index rather than index -> number but I don't think that distinction is very meaningful for an index -> number solution and might even lead students away from an index -> number solution.

So I have no suggestion other than that.

rbasso: Fixed!

Issue #538, opened 2017-05-09
docs: Document exercises' versioning policy


rbasso:
In #522 we introduced a way to version implementations directly in the `package.yaml` files.

In #523 we converted all the exercises to this new format.

We need to update the repository documentation to reflect those changes!

PR #565, opened 2017-06-24
fix pangram test for duplicate mixed-case chars

rsslldnphy:
From what I can see, the last test case is inteded to catch incorrect
programs such as this, which don't take into account different-cased
duplicates:

```haskell
isPangram = (== 26) . length . nub . filter isAlpha
```

However the above program passed the final test as the count of distinct
upper and lower case characters is 27, rather than 26.

This commit changes the test text to a non-pangram that has exactly 26
distinct upper and lower case characters, which causes the above,
incorrect program to fail.

### DIFF
--- a/exercises/pangram/package.yaml
+++ b/exercises/pangram/package.yaml
@@ -1,5 +1,5 @@
 name: pangram
-version: 1.0.0.3
+version: 1.1.0.3
 
 dependencies:
   - base
--- a/exercises/pangram/package.yaml
+++ b/exercises/pangram/package.yaml
@@ -1,5 +1,5 @@
 name: pangram
-version: 1.0.0.2
+version: 1.0.0.3
 
 dependencies:
   - base
--- a/exercises/pangram/test/Tests.hs
+++ b/exercises/pangram/test/Tests.hs
@@ -53,7 +53,7 @@ cases = [ Case { description = "sentence empty"
                , expected    = True
                }
         , Case { description = "upper and lower case versions of the same character should not be counted separately"
-               , input       = "the quick brown fox jumped over the lazy FOX"
+               , input       = "the quick brown fox jumps over with lazy FX"
                , expected    = False
                }
         ]


rbasso:
Hi!

Thanks for taking the time to open this PR.

> From what I can see, the last test case is inteded to catch incorrect
programs such as this, which don't take into account different-cased
duplicates:

That test case was implemented by exercism/problem-specifications#439, as a solution to the issue exercism/problem-specifications#266:

> Check out: https://github.com/exercism/xkotlin/pull/10
>
> The implementation that passed all existing tests didn't need to lowercase the input and was able to double count the same upper-case and lower-case characters. A new test was implemented to swap out `dog` in `the quick brown fox jumped over the lazy dog` with `FOX` which catches the error after asserting it's _not_ a pangram.

The implementation mentioned in exercism/kotlin#10 was the following:

```kotlin
fun isPangram(sentence: String): Boolean {
    return sentence
            .filter { it.isLetter() }
            .toList()
            .distinct()
            .count() >= 26
}
```

...which is very similar to your code:

```haskell
isPangram = (== 26) . length . nub . filter isAlpha
```

> However the above program passed the final test as the count of distinct
upper and lower case characters is 27, rather than 26.

I agree that it is undesirable that the solution passes the test, but the test cases that we used in this exercise came from [this file](https://github.com/exercism/problem-specifications/blob/master/exercises/pangram/canonical-data.json) in the [exercism/problem-specifications](https://github.com/exercism/problem-specifications) repository, which contains a pool of test cases that all languages can share.

For that reason, it would be great if you could open an issue there, so that maintainers from other languages could discuss it, and the change would help more people.

After the change is discussed and merged there, then we'll also change it here. 👍

petertseng:
Ready to go here. Can we ask that https://github.com/exercism/haskell/blob/master/exercises/pangram/package.yaml#L2 change to 1.1.0.3?

rsslldnphy:
👍 have rebased and bumped version

exercises/pangram/package.yaml
```diff
@@ -1,5 +1,5 @@
 name: pangram
-version: 1.0.0.2
+version: 1.0.0.3
```
petertseng: first three digits signify x-common version we are following. Since we are following 1.1.0, please 1.1.0.3 not 1.0.0.3

I am sorry that we did not document this (https://github.com/exercism/haskell/issues/538)

petertseng:
also, I considered asking for a https://github.com/exercism/haskell#example-solution `fail` example solution that would ensure that we have a test that catches this incorrect solution, but perhaps that problem is better solved on the problem-specifications side. The current fail solutions we have are checking for fails on Haskell-specific things (strictness)

petertseng:
Very good, thanks.

PR #566, opened 2017-06-28
stack.yaml: lts-8.2 -> lts-8.20

rbasso:
It's been a while since the last time we upgraded it, and the new snapshot has 215 (2428 - 2213) more packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ cache:
     - $HOME/.foldercache # Per exercise `.stack-work` cache.
 
 env:
- - RESOLVER="lts-8.2" CURRENT="YES"  # Equal to each stack.yaml.
+ - RESOLVER="lts-8.20" CURRENT="YES" # Equal to each stack.yaml.
  - RESOLVER="nightly"                # Latest nightly snapshot.
 
 matrix:
--- a/exercises/accumulate/stack.yaml
+++ b/exercises/accumulate/stack.yaml
@@ -1,1 +1,1 @@
-resolver: lts-8.2
+resolver: lts-8.20
--- a/exercises/acronym/stack.yaml
+++ b/exercises/acronym/stack.yaml
@@ -1,1 +1,1 @@
-resolver: lts-8.2
+resolver: lts-8.20
--- a/exercises/all-your-base/stack.yaml
+++ b/exercises/all-your-base/stack.yaml
@@ -1,1 +1,1 @@
-resolver: lts-8.2
+resolver: lts-8.20
--- a/exercises/allergies/stack.yaml
+++ b/exercises/allergies/stack.yaml
@@ -1,1 +1,1 @@
-resolver: lts-8.2
+resolver: lts-8.20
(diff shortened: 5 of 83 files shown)


petertseng:
You have indeed changed every stack.yaml

what about https://github.com/exercism/haskell/blob/master/.travis.yml#L15 ?

I think this upgrade is good.

rbasso:
> what about https://github.com/exercism/haskell/blob/master/.travis.yml#L15 ?

Ops. I really need to change that. Sorry!

rbasso:
It seems that we are not able to build using lts-8.20 under 50 minutes. I'll try to rebuild the lts-8.2 cache and hold/close this PR while we think in an alternative.

Some possible solutions:

- Do not build `hlint`.
- Do not build the benchmarks (`criterion` is a heavy dependency).
- Pass `--fast` to stack.

rbasso:
We are stuck in a situation where it is not possible building `master` because with `lts-8.2` is also taking more than 50 minutes.

> The job exceeded the maximum time limit for jobs, and has been terminated.

rbasso:
I was able to rebuild everything after disabling compiling the benchmarks, @petertseng.

While I think this is undesirable, I couldn't find another immediate solution to restore our ability to use Travis-CI.

petertseng:
Hmm this is interesting. So is the idea that every time we want to change the stack snapshot and/or clear the cache, we would have to temporarily disable benchmarks and then enable them after it is merged? Well, seems like some work.

So what would you say to the following ideas:

* Build benchmarks in their own separate Travis CI job
* Split up the exercises into N groups, run each group in separate Travis CI job.

The disadvantage is of course that we would then be taking up extra job slots and the entire Exercism organisation still has only 5 slots across all its repositories.

rbasso:
I was considering leaving the benchmarks disabled until we find a solution. The way things are now, we can't review PRs.

Enabling benchmarks again would be really inconvenient.

NobbZ:
Perhaps a dockerized approach which has all necessary dependencies prebuilt would suffice?

I already touched the ID in exercism/idris#25.

rbasso:
We could do that for some dependencies, @NobbZ, like `stack` and `hlint`, but that would be a problem for dependencies from exercises, as they change with each commit. Our major problem now is the benchmark library `criterion`, that takes too much time to build or upgrade the snapshot.

I'm currently testing a solution using `nix` to avoid building `hlint` to see if the savings in time are enough to get us below 50 minutes. Seems promising...

rbasso:
The major problem with the Nix approach is that it requires `sudo`, and this seems to force VM-based builds, instead of the - apparently faster - container-based ones. This is why we still need to disable the benchmarks, even after reducing the building time.

I'll see if we can use Nix with PRoot to keep our builds in the container-based infrastructure.

rbasso:
I guess I went overboard in this PR, as it now has little to do with bumping the stack's snapshot.

Considering that we are unable to build our tests at the moment, I'll close this PR until we solve #568. Later we can decide if we reopen it or create a new one.

Issue #570, opened 2017-07-01
Should we have a custom Prelude for tests?


rbasso:
Sometimes the standard `Prelude` makes writing beautiful code a really painful experience, forcing us to import a lot of stuff even to write basic code. Programmers tend to avoid imports, and this favors some bad coding practices that use functions already in the standard Prelude.

Most of the times, the imports we have depend only on `base`...

##### Package dependencies

```plain
 1 array      (saddle-points)
 1 criterion  (parallel-letter-frequency)
 1 multiset   (go-counting)
 1 vector     (matrix)
 3 QuickCheck (trinary, simple-linked-list, octal)
 3 text       (parallel-letter-frequency, forth, sgf-parsing)
 3 time       (lens-person, gigasecond, meetup)
 6 containers (parallel-letter-frequency, go-counting, etl, nucleotide-count, sgf-parsing, pov)
82 base       all
82 hspec      all
```

###### Available in `BasePrelude`, from `base-prelude` package

```haskell
Control.Arrow      ((&&&))
Control.Concurrent (forkIO, newEmptyMVar, putMVar, takeMVar)
Control.Exception  (Exception, throw, evaluate)
Control.Monad      (foldM, forM_, replicateM, unless)
Data.Bifunctor     (bimap, first)
Data.Char          (intToDigit, isSpace, toLower, toUpper)
Data.Either        (isLeft)
Data.Foldable      (for_)
Data.Function      (on)
Data.Ix            (inRange)
Data.Maybe         (fromJust)
Data.Tuple         (swap)
Data.List          (isPrefixOf, nub, sort)
Text.Printf        (printf)
Numeric            (showIntAtBase, showOct)
```

###### Not in `BasePrelude`

GHC.Exts (toList)

##### Imports from other packages

```haskell
array       Data.Array                (listArray)
containers  Data.Map                  (empty, fromList, lookup, singleton)
containers  Data.Set                  (toAscList)
containers  Data.Tree                 (Tree(Node), rootLabel)
multiset    Data.MultiSet             (fromOccurList, toOccurList)
text        Data.Text                 (concat)
time        Data.Time.Calendar        (fromGregorian)
time        Data.Time.Clock           (UTCTime)
time        Data.Time.Format          (ParseTime, TimeLocale, defaultTimeLocale,
                                       iso8601DateFormat, parseTimeOrError)
vector      Data.Vector               (fromList)
hspec       Test.Hspec.Runner         (configFastFail, defaultConfig, hspecWith)
[...]

PR #572, opened 2017-07-01
(WIP) pov: Use custom test suite Prelude

rbasso:
This is just a proof-of-concept mentioned in #570. See that issue for more information.

The custom `Prelude` is in the `common/test-prelude` folder. I guess it could also be hosted in GitHub if desired.

### DIFF
--- a/exercises/pov/stack.yaml
+++ b/exercises/pov/stack.yaml
@@ -1,1 +1,6 @@
 resolver: lts-8.2
+
+packages:
+  - location: .
+  - location: ./common/test-prelude
+    extra-dep: true
--- a/bin/test-example
+++ b/bin/test-example
@@ -28,6 +28,10 @@ cp -R -L                      \
     ${exampledir}/*           \
     "${buildfolder}"
 
+if [ -d ${exercisedir}/common ]; then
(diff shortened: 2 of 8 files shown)


exercises/pov/stack.yaml
```diff
@@ -1 +1,6 @@
 resolver: lts-8.2
+
+packages:
+  - location: .
+  - location: ./common/test-prelude
```
petertseng: if this will allow `..`, then an idea I have might work.

rbasso:
Note: The way it is now, the `test-prelude` is rebuilt each time the test runs.

exercises/pov/stack.yaml
rbasso: I was thinking about it, and maybe using a package outside the exercise's folder could be problematic. Two exercises fetched at different times could need distinct versions of the library.

I seems that this would only work would be with versioned packages.

rbasso:
Considering that there are technical problem in my proposal (rebuilds the custom prelude for each exercise) and that I don't have time to deal with that for a while, I'll close this PR and the corresponding issue.

PR #581, opened 2017-07-21
Add default maintainers file

kytrinyx:
We wish to feature the maintainers of each track on the new (nextercism) website. Ideally, the bios will be tailored to each track, giving a friendly view into each maintainer's background and interest in the language, and even (perhaps) what they enjoy most about maintaining the track, or what type of things they tend to focus on.

The main reasons for including these are to:

* provide a stronger basis for empathy with maintainers
* give well-deserved recognition
* plant a seed in people's minds that _Exercism tracks are maintained by people like me... I could do that!_

Being featured on the Exercism website is entirely optional, and we're generating all the data with show_on_website=false so that if you don't want to bother, you don't have to change anything.

If you've moved on, or decide to do so, please submit and merge a PR (for visibility to the others involved) that sets **alumnus** to _true_ for your maintainer entry. I'm working on a bot that will update the GitHub team to reflect the current state of this file.

We can merge this as is, and any changes can be added in subsequent pull requests.

See https://github.com/exercism/meta/issues/20

FYI @etrepum, @petertseng, @abo64, @rbasso - you are in the maintainer team for this track.

PR #582, opened 2017-07-25
maintainers: Set etrepum as alumnus

rbasso:
As suggested in #581.

@petertseng wrote:

> Probably makes sense to say that etrepum is alumnus.

@kytrinyx wrote:

Probably. If he doesn't pop in here I'll send him an email to see if he's OK with that.

### DIFF
--- a/config/maintainers.json
+++ b/config/maintainers.json
@@ -3,7 +3,7 @@
     {
       "github_username": "etrepum",
       "show_on_website": false,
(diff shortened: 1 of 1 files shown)


kytrinyx:
> We might consider deleting all fields, leaving only username, show_on_website, and alumnus.

That seems like a good call when 'alumnus' is true.

PR #585, opened 2017-08-02
Implement exercise protein-translation

clementi:


levex:
Why is this commit empty?

clementi:
The section ["Avoiding Duplicate Work"](https://github.com/exercism/docs/blob/master/you-can-help/implement-an-exercise-from-specification.md#avoiding-duplicate-work) in the document ["Porting an Exercise to Another Language Track"](https://github.com/exercism/docs/blob/master/you-can-help/implement-an-exercise-from-specification.md) indicates that one should create an empty WIP PR to indicate that you're working on the particular exercise.

levex:
Apologies, I wasn't aware of that!

clementi:
This exercise has now been implemented and the PR is ready for review.

clementi:
@kytrinyx or @levex, this is ready for review.

_Sent from my Motorola XT1575 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

kytrinyx:
Thank you @clementi! I've pinged the Haskell track maintainers so that they can take a look at it.

@levex If you have a few moments to spare to provide your input that would be valuable as well. ✨

exercises/protein-translation/examples/success-standard/src/ProteinTranslation.hs
```diff
@@ -0,0 +1,26 @@
+module ProteinTranslation (toProtein) where
+
+import Data.List.Split (chunksOf)
+
+toProtein :: String -> [String]
+toProtein strand = takeWhile ("STOP" /=) $ map codonToProtein $ chunksOf 3 strand
+
+codonToProtein :: String -> String
```
levex: Just a quick question, maybe I'm missing something, but to me the description is a little misleading. It seems that the correct solution would be just to do the map operation written in the description, but judging by this and the test cases it seems that it actually expects the protein names instead of the codons?

clementi: I was basing this off the Scala version: http://exercism.io/exercises/scala/protein-translation/test-suite

config.json
```diff
@@ -782,7 +791,7 @@
       "uuid": "d5997b60-e54c-4caa-beae-8614f0da3fb3",
       "slug": "trinary",
       "deprecated": true
-    }
+    }
```
petertseng: please remove these spaces, as they have nothing to do with adding the exercise

config.json
```diff
@@ -87,6 +87,15 @@
       ]
     },
     {
+      "uuid": "13154065-a5dc-4824-92ed-b2fae41c4dd6",
+      "slug": "protein-translation",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
```
petertseng: I was going to ask about difficulty 1 because I didn't think this exercise is difficulty 1, but given that run-length-encoding is also difficulty 1, I guess I can't complain.

exercises/protein-translation/README.md
```diff
@@ -0,0 +1,79 @@
+# Rna Transcription
```
petertseng: this is the description for the RNA transcription exercise (https://github.com/exercism/problem-specifications/blob/master/exercises/rna-transcription/description.md), whereas given that this is in the `protein-translation` directory it needs to be the description of the protein translation exercise: https://github.com/exercism/problem-specifications/blob/master/exercises/protein-translation/description.md

exercises/protein-translation/package.yaml
```diff
@@ -0,0 +1,20 @@
+name: protein-translation
+version: 1.0.0.0
```
petertseng: given that https://github.com/exercism/problem-specifications/tree/master/exercises/protein-translation/canonical-data.json does not exist (that link will lead to a 404), the version number that is in accordance with the versioning policy (https://github.com/exercism/haskell/issues/522) will be 0.1.0.1 rather than 1.0.0.0

exercises/protein-translation/test/Tests.hs
```diff
@@ -0,0 +1,95 @@
...
+               }
+        , Case { description = "identifies leucine codon (UUA)"
+               , strand      = "UUA"
+               , expected    = ["Leucine"]
+               }
+        , Case { description = "identifies leucine codon (UUG)"
+               , strand      = "UUG"
+               , expected    = ["Leucine"]
+               }
+        , Case { description = "identifies leucine codon (UUG)"
```
petertseng: is this case any different from the one immediately above? I see two `"identifies leucine codon (UUG)"`

exercises/protein-translation/test/Tests.hs
```diff
@@ -0,0 +1,95 @@
...
+               }
+        , Case { description = "translate RNA strand into correct protein"
+               , strand      = "AUGUUUUGG"
+               , expected    = ["Methionine", "Phenylalanine", "Tryptophan"]
+               }
+        , Case { description = "stops translation if the STOP codon is present"
+               , strand      = "AUGUUUUAA"
+               , expected    = ["Methionine", "Phenylalanine"]
+               }
+        , Case { description = "stops translation of longest strand"
```
petertseng: I don't understand what "longest" means in this context. Can you clarify? I know that it is the longest strand used in the test cases so far, but I don't understand why that must be specified in the description. To me the important part of this test case is that even if there are codons after STOP, they are not included in the output. Shouldn't that be the description?

petertseng:
I regret that between the dates of August 26 and September 15 (a period of 20 days), it was ready for review, but no maintainer of this track reviewed it.

An >= equivalent period of time (24 days) has passed since my review, and we had unanswered questions about the duplicate description `"identifies leucine codon (UUG)"` and the meaning of `"stops translation of longest strand"`, which we need answered before being able to merge it. We await your answer, but to get the PR out of the queue for now, we are closing it.

Despite it being closed, we are still open to receiving the answers, and would gladly reopen it upon receiving them. Feel free to reopen it by yourself if the permissions allow that (not familiar enough to know whether it is) or ask us to and we will.

Average-user:
Can I implement  this?

petertseng:
Yes. Note that https://github.com/exercism/problem-specifications/blob/master/exercises/protein-translation/canonical-data.json exists now, so take note of it appropriately.

Average-user:
ok, But I'll make a new PR.

petertseng:
Indeed, unfortunately it is not possible to transfer ownership of a PR to someone else, so you will indeed have to make a new one.

Average-user:
Oh, didn't now that

Average-user:
@petertseng  Should this be ```String -> [String]``` or ```String -> Maybe [String]```?

petertseng:
Ah, because there are strings that could not translate to any sequence of proteins? Okay, I suggest `String -> Maybe [String]` then. I note the canonical data doesn't have any case that would result in `None`, so you are not obligated to add any, but would still use `Maybe` nevertheless.

Thinking about `String -> Maybe [Protein]` though? Note sure whether I feel like making an ADT for these proteins. Try it out and see what the two options look like, then say why you decided on your choice.

Average-user:
Mmm ADT are fun, I'll see

PR #587, opened 2017-08-22
Add default introductory code example

kytrinyx:
Instead of having logic for a fallback in the backend, we're choosing a
default for the snippet file.

For tracks that have core exercises, we pick the first core exercise.
For tracks without these, we pick the first exercise listed in the config.

See https://github.com/exercism/meta/issues/89

### DIFF
--- a/docs/SNIPPET.txt
+++ b/docs/SNIPPET.txt
@@ -1,0 +1,4 @@
+module HelloWorld (hello) where
+
+hello :: String
+hello = "Hello, World!"


petertseng:
Currently in this PR, the contents of docs/SNIPPET.txt are not Haskell code. This is inadvisable.

kytrinyx:
Ouch. This appears to have matched the example_pattern from the config.json. Can we improve the sample_pattern to pick up the actual example?

kytrinyx:
Updated example pattern in https://github.com/exercism/haskell/pull/588

PR #588, opened 2017-08-22
Add an explicit example pattern to the track config

kytrinyx:
The default example pattern matches more than just the reference solution.
It also matches some of the support files that live within the example directories, such
as yaml files.

This has been tested against configlet to ensure that it detects a reference solution
for each exercise.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,6 +1,8 @@
 {
   "language": "Haskell",
   "active": true,
+  "solution_pattern": "example.*[.]hs",
+  "ignore_pattern": "example",
   "exercises": [
     {
       "uuid": "c97d9dda-c7ee-4595-b447-693559d03ba5",


petertseng:
> It also matches some of the support files that live within the example directories, such
as yaml files.

I would now like to ask: What negative effect does this have?

I am also obligated to ask: If this causes the yaml files to no longer be considered an example file, will they then be delivered to students? That is undesirable. They should not be delivered.

kytrinyx:
> I would now like to ask: What negative effect does this have?

This means that we would not be able to detect if we're actually missing an example solution.

>  They should not be delivered.

Got it. I'll add the explicit `ignore_pattern` as well.

petertseng:
> This means that we would not be able to detect if we're actually missing an example solution.

Then https://github.com/exercism/haskell/blob/master/bin/test-example would be unable to compile and run the example, so Travis CI will fail as expected

kytrinyx:
I checked: the default ignore pattern would prevent them from being delivered. That said, I think it's worth making this explicit in any case, so I'm adding the value to the config.

petertseng:
> This means that we would not be able to detect if we're actually missing an example solution.

If we were missing an .hs file for any example, https://github.com/exercism/haskell/blob/master/bin/test-example would be unable to compile and run it, so Travis CI will fail as expected

kytrinyx:
> If we were missing an .hs file [...] Travis CI will fail as expected

OK, this is good.

However, the reason I submitted YAML to the snippets PR rather than the Haskell solution, is because the YAML was incorrectly being detected as a solution file when it is not. I don't know that we need this on a regular basis, but I would prefer that it were correct.

petertseng:
I can't find any `example_pattern` in https://github.com/exercism/configlet/search?utf8=%E2%9C%93&q=example_pattern&type= nor https://github.com/exercism/trackler/search?utf8=%E2%9C%93&q=example_pattern&type= - can you show me the code that expects `example_pattern` to be present?

petertseng:
EBWODP:

> that expects example_pattern to be present

... or makes decisions based on the value of `example_pattern`?

kytrinyx:
Oh! That's because I'm completely wrong. It's `solution_pattern` https://github.com/exercism/configlet/blob/2c49de1e9f4a404047fa9483ae8d76a45d195ba6/track/config.go

Sheesh.

Updating the PR. Thank you for paying attention!

kytrinyx:
Alright, I've updated that and combined it into a single commit, as that felt more correct.

config.json
```diff
@@ -1,6 +1,8 @@
 {
   "language": "Haskell",
   "active": true,
+  "solution_pattern": "example.*[.]hs",
+  "ignore_pattern": "[Ee]xample",
```
petertseng: because of https://github.com/exercism/trackler/blob/58b942a857b61de29c9d799e0c61ec57bbfa564a/lib/trackler/implementation.rb#L71, the `[Ee]` is unnecessary. I think it should just be `example`.

petertseng:
EBWODP

> I can

I can additionally guarantee that all examples of this track contain `examples` in their path, not just `example`

config.json
kytrinyx: Trackler is about to be retired. That said, if this track always uses `example` and never `Example` then I think we should simplify to what is actually used.

petertseng: Indeed, `example` only.

kytrinyx: Ok, updated. Thanks for you help on this one!

PR #593, opened 2017-09-19
collatz-conjecture

Average-user:
I'm implementing the collatz conjecture of the Unimplmented excercises of Haskell (excercism.io). But i wasn't sure of what to put on the config.json, uuid part of the excercise.


petertseng:
> config.json, uuid part of the excercise.

You guessed correctly (I think) that you generate any arbitrary UUID and use it.

I notice that https://github.com/exercism/docs/blob/master/you-can-help/implement-an-exercise-from-specification.md#configuring-the-exercise did not tell you you are supposed to generate a UUID. That is a deficiency in the documentation, which should be reported.

Please notify if anything else was unclear!

exercises/collatz-conjecture/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+The [exercism/haskell](https://github.com/exercism/haskell) repository on
+GitHub is the home for all of the Haskell exercises.
+
+If you have feedback about an exercise, or want to help implementing a new
+one, head over there and create an issue.  We'll do our best to help you!
+
+## Source
+
+All of Computer Science [http://www.wolframalpha.com/input/?i=binary&a=*C.binary-_*MathWorld-](http://www.wolframalpha.com/input/?i=binary&a=*C.binary-_*MathWorld-)
```
petertseng: this is not the source listed in https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/metadata.yml

exercises/collatz-conjecture/package.yaml
```diff
@@ -0,0 +1,21 @@
+
+name: collatz-conjecture
+version: 1.0.0.3
```
petertseng: as https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/canonical-data.json has version 1.1.1, the version that would comply with our versioning policy in https://github.com/exercism/haskell/issues/522 is `1.1.1.1`, and ensuring that the tests in the Haskell test file match those in the JSON file

exercises/collatz-conjecture/package.yaml
```diff
@@ -0,0 +1,21 @@
+
```
petertseng: extra newline at the beginning of this file, it appears? I would remove it.

exercises/collatz-conjecture/README.md
Average-user: ups, sorry. Im changing it

exercises/collatz-conjecture/package.yaml
petertseng: (or, alternatively, if you believe those tests are unsuitable for the Haskell track for any reason,  let's discuss why they are unsuitable, and then use version`0.1.0.1`)

exercises/collatz-conjecture/package.yaml
petertseng: (it has been removed, now github put my comment on the wrong line, so I'll delete this comment)

exercises/collatz-conjecture/package.yaml
```diff
@@ -0,0 +1,20 @@
+name: collatz-conjecture
+version: 1.1.1
```
petertseng: needs an additional .1 serial number, resulting in 1.1.1.1

Average-user:
Well. I'm glad to have guessed. But i still without understand Travis, Will ever let me free?

Average-user:
Travis gives me an error of another excercise 
```/home/travis/build/exercism/haskell/build/accumulate/fail-reverse/.stack-work/install/x86_64-linux/lts-8.21/8.0.2/lib/x86_64-linux-ghc-8.0.2/accumulate-0.0.0-BR5YsFtF45zACYc92bgN8F
Registering accumulate-0.0.0...
accumulate-0.0.0: test (suite: test)
empty accumulation
accumulate squares
accumulate upcases
accumulate reversed strings
accumulate recursively
accumulate non-strict FAILED [1]
Failures:
  test/Tests.hs:35: 
  1) accumulate non-strict
       uncaught exception: ErrorCall (accumulate should be even lazier, don't use reverse!
       CallStack (from HasCallStack):
         error, called at test/Tests.hs:36:45 in main:Main)
Randomized with seed 1712205088
Finished in 0.0005 seconds
6 examples, 1 failure
Completed 2 action(s).
Test suite failure for package accumulate-0.0.0
    test:  exited with: ExitFailure 1
Logs printed to console```

**Comment by petertseng** [comment]
> Travis gives me an error of another excercise

**Comment by petertseng** [comment]
> Travis gives me an error of another excercise

> testing /home/travis/build/exercism/haskell/exercises/accumulate/examples/fail-reverse - should build, but fail tests

It should fail, and it did, so no problem. Please look for problems related to collatz-conjecture

**Comment by Average-user** [comment]
But why still a red X on my pull request? 

And i was thinking that Knuth shuffle algorithm, might be a good excercise

**Comment by petertseng** [comment]
> But why still a red X on my pull request?

I don't know, what does the Travis CI output at https://travis-ci.org/exercism/haskell/builds/277117302 say? Alternatively, what do you get if you try to `stack test` in the `collatz-conjecture` directory on your computer? Alternatively, what do you get if you try to run https://github.com/exercism/haskell/blob/master/bin/test-all-examples ? Something like `bin/test-all-examples exercises/collatz-conjecture` ?

> And i was thinking that Knuth shuffle algorithm, might be a good excercise

Okay, it can be suggested in https://github.com/exercism/problem-specifications . My advice is that based on https://github.com/exercism/problem-specifications/issues/761 exercises that test a *specific algorithm* are difficult to write tests for because there is no good way to verify that a student's implementation uses that specific algorithm, rather than any other algorithm that achieves the same effect.

**Comment by Average-user** [comment]
Yeah you are right about the Knuth shuffle algorithm, and it's even worts if you take on count that is a SHUFFLE algorithm, not always returns the same result for the same input (well, strictly yes but ...)

**Comment by Average-user** [comment]
At the end the problem was that i misspelled a file name.    

And i followed your advice and wrote an issue on: [https://github.com/exercism/problem-specifications/issues/906](https://github.com/exercism/problem-specifications/issues/906).
But it's a different  exercise, not the Knuth algorithm.

File: exercises/collatz-conjecture/test/Tests.hs [hunk 5]
```diff
@@ -0,0 +1,60 @@
...
+                 , expected    :: Integer
+                 }
+
+cases :: [Case]
+cases = [ Case { description = "collatz 1 don't require any steps."
+               , number      = 1
+               , expected    = 0
+               }
+        , Case { description = "collatz 2 require 1 step."
+               , number      = 2
```
petertseng: on my advice, you did change the problem version to 1.1.1.1, to match the 1.1.1 of https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/canonical-data.json, but you did not then use the test cases in https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/canonical-data.json. Can you do one of the two:

1. Use the test cases that appear in https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/canonical-data.json
2. Confirm that the test cases in https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/canonical-data.json are unsuitable for the Haskell track, and thus we cannot use that file, and thus we must change our version to 0.1.0.1

Average-user: so it's ok if i change it from 1.1.1.1 to 1.1.1 and use [this cases](https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/canonical-data.json) ?

petertseng: 1.1.1 is not a valid Haskell track version. The two choices are:

1. Use 1.1.1.1 and the cases in https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/canonical-data.json
2. Use 0.1.0.1 and whatever cases desired, with an explanation of why https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/canonical-data.json cannot be used.

Average-user: yeah, i forgot the extra 1 again. sorry about that. i'll use the first one

exercises/collatz-conjecture/test/Tests.hs
```diff
@@ -0,0 +1,59 @@
...
+               }
+        , Case { description = "divide if even"
+               , property    = "steps"
+               , number      = 16
+               , expected    = Just 4
+               }
+        , Case { description = "even and odd steps"
+               , property    = "steps"
+               , number      = 12
+               , expected    =Just 9
```
petertseng: add a space after the = here?

config.json
```diff
@@ -59,6 +59,15 @@
       ]
     },
     {
+      "uuid": "6eba4eac-8395-4ad6-ac21-3651a11ab4b4",
+      "slug": "collatz-conjecture",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
```
petertseng: since this exercise now uses Maybe, please add it to the topics list (see "rna-transcription" for example of how it is done)

exercises/collatz-conjecture/examples/success-standard/src/CollatzConjecture.hs
```diff
@@ -0,0 +1,9 @@
+module CollatzConjecture (collatz) where
+
+collatzHelper :: Integer -> Integer -> Maybe Integer
+collatzHelper t x | x == 1    = Just t
+                  | even x    = collatzHelper (t+1) (x `div` 2)
+                  | otherwise = collatzHelper (t+1) (x*3 + 1)
+
+collatz :: Integer -> Maybe Integer
+collatz x = if  x <= 0 then Nothing else collatzHelper 0 x
```
petertseng: there are two spaces after the `if`, you could remove one of them, right?

exercises/collatz-conjecture/test/Tests.hs
```diff
@@ -0,0 +1,59 @@
...
+
+data Case = Case { description :: String
+                 , property    :: String
+                 , number      :: Integer
+                 , expected    :: Maybe Integer
+                 }
+
+cases :: [Case]
+cases = [ Case { description = "zero steps for one"
+               , property    = "steps"
```
petertseng: the `property` is not being used. It should be removed.

config.json
Average-user: ok

exercises/collatz-conjecture/test/Tests.hs
Average-user: i will, but maybe it has to be removed from [https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/canonical-data.json)

petertseng: It needs to be in problem-specifications since it tells what is being tested.

It must not be here since we already know what we are testing (it's specified in code)

Average-user:
No, do it. I'm ok with that

petertseng:
Thank you for implementing this exercise!

exercises/collatz-conjecture/test/Tests.hs
```diff
@@ -0,0 +1,52 @@
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
```
petertseng: I don't think this line is necessary. Can you check? If you confirm it is not necessary, please send a PR that deletes it.

petertseng: I retract. I think https://github.com/exercism/haskell/pull/144 tells me it is desirable to have for exercises with numbers.

PR #595, opened 2017-09-20
adding isogram

Average-user:
Another Unimplemented exercise

### DIFF
--- a/config.json
+++ b/config.json
@@ -75,7 +75,6 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-        "Data.Char"
       ]
     },
     {
--- a/config.json
+++ b/config.json
@@ -69,6 +69,16 @@
       ]
     },
     {
+      "uuid": "6eba4eac-6665-4ad6-ac21-3651a11ab4b4",
+      "slug": "isogram",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "Data.Char"
+      ]
+    },
+    {
       "uuid": "85d77b8e-9b87-4d02-9fba-81f843bd66f1",
       "slug": "rna-transcription",
       "core": false,
--- a/exercises/isogram/README.md
+++ b/exercises/isogram/README.md
@@ -6,10 +6,12 @@ An isogram (also known as a "nonpattern word") is a word or phrase without a rep
 
 Examples of isograms:
 
-lumberjacks
-background
-downstream
-The word isograms, however, is not an isogram, because the s repeats.
+- lumberjacks
+- background
+- downstream
+
+The word *isograms*, however, is not an isogram, because the s repeats.
+
 
 ## Getting Started
 
@@ -65,8 +67,7 @@ one, head over there and create an issue.  We'll do our best to help you!
 
 ## Source
 
-see more  at [wikipedia page of isograms]("https://en.wikipedia.org/wiki/Isogram")
-
+Wikipedia [https://en.wikipedia.org/wiki/Isogram](https://en.wikipedia.org/wiki/Isogram)
 
 ## Submitting Incomplete Solutions
 It's possible to submit an incomplete solution so you can see how others have completed the exercise.
--- a/exercises/isogram/README.md
+++ b/exercises/isogram/README.md
@@ -1,0 +1,72 @@
+# Isogram
+
+Determine if a word or phrase is an isogram.
+
+An isogram (also known as a "nonpattern word") is a word or phrase without a repeating letter.
+
+Examples of isograms:
+
+lumberjacks
+background
+downstream
+The word isograms, however, is not an isogram, because the s repeats.
+
+## Getting Started
+
+For installation and learning resources, refer to the
+[exercism help page](http://exercism.io/languages/haskell).
+
+## Running the tests
+
+To run the test suite, execute the following command:
+
+```bash
+stack test
+```
+
+#### If you get an error message like this...
+
+```
+No .cabal file found in directory
+```
+
+You are probably running an old stack version and need
+to upgrade it.
+
+#### Otherwise, if you get an error message like this...
+
+```
+No compiler found, expected minor version match with...
+Try running "stack setup" to install the correct GHC...
+```
+
+Just do as it says and it will download and install
+the correct compiler version:
+
+```bash
+stack setup
+```
+
+## Running *GHCi*
+
+If you want to play with your solution in GHCi, just run the command:
+
+```bash
+stack ghci
+```
+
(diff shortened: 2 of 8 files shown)


Average-user:
This one is ok. right?

config.json
```diff
@@ -69,6 +69,16 @@
       ]
     },
     {
+      "uuid": "6eba4eac-6665-4ad6-ac21-3651a11ab4b4",
+      "slug": "isogram",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "Data.Char"
```
petertseng: in contrast to Maybe, I don't think this:

1. affects where we would choose to place the exercise
2. is necessarliy going to be in everyone's solutions

so I would remove it

exercises/isogram/package.yaml
```diff
@@ -0,0 +1,20 @@
+name: isogram
+version: 1.1.0
```
petertseng: 1.1.0.1

exercises/isogram/test/Tests.hs
```diff
@@ -0,0 +1,64 @@
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
```
petertseng: given that #144 tells us this line helps for exercises with numbers but this is not one, I think it can be removed?

exercises/isogram/src/Isogram.hs
```diff
@@ -0,0 +1,6 @@
+module Isogram (isIsogram) where
+
+import Data.Char
```
petertseng: since

1. Not every solution wants or needs this
2. I wouldn't want to encourage the habit of importing everything rather than the specific list of functions desired

I would remove this line

config.json
Average-user: right

exercises/isogram/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+The [exercism/haskell](https://github.com/exercism/haskell) repository on
+GitHub is the home for all of the Haskell exercises.
+
+If you have feedback about an exercise, or want to help implementing a new
+one, head over there and create an issue.  We'll do our best to help you!
+
+## Source
+
+see more  at [wikipedia page of isograms]("https://en.wikipedia.org/wiki/Isogram")
```
petertseng: as you can see in https://github.com/Average-user/haskell/blob/d1556eadc748688089e3261700ab9678f2caf245/exercises/isogram/README.md, the quotes mean it is not a link.

I'm going to regenerate this README anyway so it doesn't matter whether you fix this, but don't put quotes in the brackets

exercises/isogram/src/Isogram.hs
```diff
@@ -0,0 +1,5 @@
+module Isogram (isIsogram) where
+
```
petertseng: I would remove one newline here so there is only one between `module` and `isIsogram`, not two.

exercises/isogram/README.md
Average-user: ok, so the next time i make PR like this, do i even have to put the README.md?

petertseng: I guess not! I will just do it. However, it would be good if you just put a README.md that just says "please regenerate this for me" so that I don't forget to!

Average-user: ok, also, dont merge yet, i want to implement something

Average-user:
Now you can merge if you want

Average-user:
How can i do if i want to do another PR before this one gets merge?

petertseng:
> How can i do if i want to do another PR before this one gets merge?

Make a branch.

Average-user:
But Github will still catch the differences with the (in this case) Isogram commits

petertseng:
> But Github will still catch the differences with the (in this case) Isogram commits

Github shows all commits that are not on the target branch. The Isogram commits should not be on your new branch. Achieve that using whatever means you prefer. Ask if you want a suggestion of a preferred means.

petertseng:
> But Github will still catch the differences with the (in this case) Isogram commits

Github shows all commits that are not on the target branch (Exercism's master branch) but are on the PR branch (your new branch). The Isogram commits should not be on your new branch. Achieve that using whatever means you prefer. Ask if you want a suggestion of a preferred means.

Average-user:
Thanks, i'll take that suggestion you offer .

petertseng:
`git checkout -b your_new_branch_name 7dc8cf2f0403857a79d13c4916d4b26f99af15e4`

(why, because `7dc8cf2f0403857a79d13c4916d4b26f99af15e4` is the latest commit shown on https://github.com/exercism/haskell/commits/master )

petertseng:
Case 1:  You do not already have a branch.

You decide on a branch name, and then you `git checkout -b your_new_branch_name 7dc8cf2f0403857a79d13c4916d4b26f99af15e4`

(why, because `7dc8cf2f0403857a79d13c4916d4b26f99af15e4` is the latest commit shown on https://github.com/exercism/haskell/commits/master )

Case 2: You already have a branch.

You `git checkout your_existing_branch_name` if you are not already on that branch, then you `git rebase -i 7dc8cf2f0403857a79d13c4916d4b26f99af15e4`, then you delete all lines that are related to Isogram.

petertseng:
Case 1:  You do not already have a branch.

You decide on a branch name, and then you `git checkout -b your_new_branch_name 7dc8cf2f0403857a79d13c4916d4b26f99af15e4`

(why, because `7dc8cf2f0403857a79d13c4916d4b26f99af15e4` is the latest commit shown on https://github.com/exercism/haskell/commits/master )

Case 2: You already have a branch.

You `git checkout your_existing_branch_name` if you are not already on that branch, then you `git rebase -i 7dc8cf2f0403857a79d13c4916d4b26f99af15e4`, then you delete all lines that are related to Isogram.

Average-user:
what  does this ```7dc8cf2f0403857a79d13c4916d4b26f99af15e4```?

Average-user:
Instead of this key (or whatever) ```7dc8cf2f0403857a79d13c4916d4b26f99af15e4``` can I use the link ``` https://github.com/exercism/haskell```?

petertseng:
> Instead of this key (or whatever) ```7dc8cf2f0403857a79d13c4916d4b26f99af15e4``` can I use the link ``` https://github.com/exercism/haskell```?

If the repository https://github.com/exercism/haskell is a remote on your local copy of the git repo, you may instead use `your_name_for_the_exercism_remote/master` instead of `7dc8cf2f0403857a79d13c4916d4b26f99af15e4`.

petertseng:
Thank you for implementing this exercise!

PR #596, opened 2017-09-22
Implementing diamond

Average-user:
Sorry for all this PRs, but I'm absolutely not hurry. But, there are to many unimplemented exercises on Haskell , makes me sad.

### DIFF
--- a/exercises/diamond/examples/success-standard/src/Diamond.hs
+++ b/exercises/diamond/examples/success-standard/src/Diamond.hs
@@ -2,13 +2,12 @@ module Diamond (diamond) where
 
 import Data.Char (ord, chr)
 
--- replicate blank
-rb :: Int -> String
-rb x = replicate x ' '
+pad :: Int -> String
+pad x = replicate x ' '
 
 helper :: (Char, (Int, Int)) -> String
-helper (c, (0, y)) = rb y ++ [c] ++ rb y
-helper (c, (x, y)) = rb y ++ [c] ++ rb x ++ [c] ++ rb y
+helper (c, (0, y)) = pad y ++ [c] ++ pad y
+helper (c, (x, y)) = pad y ++ [c] ++ pad x ++ [c] ++ pad y
 
 diamond :: Char -> [String]
 diamond = (\x -> x ++ tail (reverse x)) . f . subtract 64 . ord
--- a/exercises/diamond/examples/success-standard/src/Diamond.hs
+++ b/exercises/diamond/examples/success-standard/src/Diamond.hs
@@ -10,7 +10,7 @@ helper (c, (0, y)) = rb y ++ [c] ++ rb y
 helper (c, (x, y)) = rb y ++ [c] ++ rb x ++ [c] ++ rb y
 
 diamond :: Char -> [String]
-diamond = (\x -> x ++ tail (reverse x)) . f . (subtract 64) . ord
+diamond = (\x -> x ++ tail (reverse x)) . f . subtract 64 . ord
   where rows x = zip (0 : take (x-1) [1, 3..]) [x-1, x-2..0]
-        f      = map helper . zip abc . rows
-        abc    = (map chr [65..90])
+        f      = zipWith (curry helper) abc . rows
+        abc    = map chr [65..90]
--- a/exercises/diamond/examples/success-standard/src/Diamond.hs
+++ b/exercises/diamond/examples/success-standard/src/Diamond.hs
@@ -5,12 +5,12 @@ import Data.Char (ord, chr)
 pad :: Int -> String
 pad x = replicate x ' '
 
-helper :: Char -> (Int, Int) -> String
-helper c (0, y) = pad y ++ [c] ++ pad y
-helper c (x, y) = pad y ++ [c] ++ pad x ++ [c] ++ pad y
+oneRow :: Char -> (Int, Int) -> String
+oneRow c (0, y) = pad y ++ [c] ++ pad y
+oneRow c (x, y) = pad y ++ [c] ++ pad x ++ [c] ++ pad y
 
 diamond :: Char -> [String]
 diamond = (\x -> x ++ tail (reverse x)) . f . subtract 64 . ord
   where rows x = zip (0 : take (x-1) [1, 3..]) [x-1, x-2..0]
-        f      = zipWith helper abc . rows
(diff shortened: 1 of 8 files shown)


petertseng:
The diamond exercise is interesting. According to  https://github.com/exercism/problem-specifications/blob/master/exercises/diamond/canonical-data.json#L10-L19 it is possible to use property-based testing. According to https://github.com/exercism/problem-specifications/pull/191#issuecomment-254589087 it is not only possible, it is preferred.

It is certainly better to have this exercise (in any form) than to not have it at all. It is probably better to have it with property-based tests than value-based tests. With this in mind, your options are:

* To use property-based tests, using whatever means you deem appropriate.
  * Note that the https://github.com/exercism/haskell/blob/master/exercises/octal/test/Tests.hs and https://github.com/exercism/haskell/blob/master/exercises/trinary/test/Tests.hs and https://github.com/exercism/haskell/blob/master/exercises/simple-linked-list/test/Tests.hs use QuickCheck.
  * Because this is an option, I delay my review of tests because I wouldn't want to review something that ends up getting rewritten.
* To indicate that you are unable or unwilling to do so.
  * With this option, I will review the tests as-is.

In neither case is it necessary for me to delay reviewing the other files, so they will happen when able (as should go without saying). Able is not yet defined, but next two days are unable.

petertseng:
> To indicate that you are unable or unwilling to do so.

With this option, an issue will be opened asking any other contributor to replace the value-based tests with property-based tests.

petertseng:
> With this option, an issue will be opened asking any other contributor to replace the value-based tests with property-based tests.

an issue will be opened after the PR is merged

Average-user:
@petertseng Ok, I'll try, But let me know what you think about the other files, when you are able to.

Average-user:
@petertseng mhh, I'm not achieving to implement  property-based-tests, and in this  case doesn't look so bad to use value-based-tests. So if you are ok, I would merge like it is, and after that, if you want, I can open the issue myself.

petertseng:
> if you want, I can open the issue myself.

thank you, I would appreciate if you would open it.

exercises/diamond/test/Tests.hs
```diff
@@ -0,0 +1,109 @@
...
+        assertion = diamond input `shouldBe` expected
+
+
+data Case = Case { description :: String
+                 , input       :: Char
+                 , expected    :: [String]
+                 }
+
+cases :: [Case]
+cases = [ Case { description = "Diamonf of 'A'"
```
petertseng: diamonf

exercises/diamond/examples/success-standard/src/Diamond.hs
```diff
@@ -0,0 +1,13 @@
+module Diamond (diamond) where
+
+import Data.Char (ord, chr)
+
+diamondH :: (Int -> String) -> (Char, (Int, Int)) -> String
+diamondH rb (c, (0, y)) = rb y ++ [c] ++ rb y
```
petertseng: I am having a hard time understanding what the H in diamondH stands for. I would prefer a name that I didn't have to ask someone to explain. This makes a single row, right? Maybe something about that.

petertseng: I am having a hard time understanding what rb stands for. I would prefer a name that I didn't have to ask someone to explain.

exercises/diamond/examples/success-standard/src/Diamond.hs
```diff
@@ -0,0 +1,13 @@
...
+diamondH rb (c, (x, y)) = rb y ++ [c] ++ rb x ++ [c] ++ rb y
+
+diamond :: Char -> [String]
+diamond = (\x -> x ++ tail (reverse x)) . f . abs . (64-) . ord
+  where rows x = zip (0 : take (x-1) [1,3..]) [x-1,x-2..0]
+        f      = map (diamondH rb) . zip (map chr [65..90]) . rows
```
petertseng: I am considering whether it is better to write `'A'..'Z' here rather than `map chr [65..90]` but I guess I would leave it to make clear the relationship with the 64 above.

exercises/diamond/examples/success-standard/src/Diamond.hs
```diff
@@ -0,0 +1,13 @@
+module Diamond (diamond) where
+
+import Data.Char (ord, chr)
+
+diamondH :: (Int -> String) -> (Char, (Int, Int)) -> String
+diamondH rb (c, (0, y)) = rb y ++ [c] ++ rb y
+diamondH rb (c, (x, y)) = rb y ++ [c] ++ rb x ++ [c] ++ rb y
+
+diamond :: Char -> [String]
+diamond = (\x -> x ++ tail (reverse x)) . f . abs . (64-) . ord
```
petertseng: the `abs` seems clumsy here, since you want to subtract 64 from result of ord. Do you prefer `(subtract 64) . ord` or `abs . (64-) . ord` ? I think the former, but unsure.

exercises/diamond/examples/success-standard/src/Diamond.hs
Average-user: replicate blanks.  And it's defined on line 13.

petertseng:
I can't understand the history with all the merge commits (https://github.com/exercism/haskell/pull/596/commits/50ba4f473c18f2d555d5576e2f47a5dce01c3da7, https://github.com/exercism/haskell/pull/596/commits/6d7a7c1af8625b64b20be9dded7783a56d37fe52, https://github.com/exercism/haskell/pull/596/commits/3efac137ab522114053e2e91aa6f3ae33c1c4f21) , next time I need you to rebase on exercism/master, not merge. Otherwise, too many unrelated commits in the commit list and I can't tell what you're trying to get merged in this PR. I'm rebasing now.

petertseng:
I hope you don't mind me renaming all your functions.

Also note that you had a `(curry helper)` but that is the only place where `helper` is being used. In that case, you might as well make `helper` a curried function and change `(curry helper)` to `helper`. See https://github.com/exercism/haskell/pull/596/commits/8f58618dabb42e2f9d32390d59286fa9ca82a3d4. I should have let you do this one by yourself, sorry.

petertseng:
Thank you for implementing this exercise, your third, I believe.

Average-user:
Yep. Thanks fot the correction

PR #597, opened 2017-09-23
maintainers: Set rbasso as alumnus

### LINKED ISSUE #589: maintainers: rbasso has left the building
I know that I'm not being very active here since May, and I don't know when/if I'll find the time to return to my duties as a maintainer.

While I would love to spend some time with Haskell and Exercism, the truth is that my current, non-IT-related, full-time job is killing my hobbies.

I may be able to take a look at the repository from time to time, but please don't count on me for anything. Feel free to remove me from the maintainers team when you think it is best.

Sorry for leaving you alone without notice, @petertseng, but I'm sure this track is in good hands!

Thank you for the opportunity of becoming a maintainer, @kytrinyx. It was a really great experience!

See you!
😄

rbasso:
Closes #589

### DIFF
--- a/config/maintainers.json
+++ b/config/maintainers.json
@@ -22,8 +22,8 @@
     },
     {
       "github_username": "rbasso",
-      "show_on_website": true,
-      "alumnus": false,
+      "show_on_website": false,
+      "alumnus": true,
       "name": null,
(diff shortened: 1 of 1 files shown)


rbasso:
> you didn't have to remove your bio, but if it is what you want

Fixed. 😄

PR #598, opened 2017-10-04
READMEs: Import problem-specifications commits

petertseng:


petertseng:
If you ask, the import was done by hand. I'll think about automating it later.

petertseng:
It is intended that all commit messages link to the original problem-specifications PR, and that the author of each commit is the author of the problem-specifications commit. Let me know if I got any of that wrong.

petertseng:
I made a mistake and a "their" -> "the" got into the "bowling: wrap paragraphs" commit, but it's not so important that I will need to retroactively change.

PR #601, opened 2017-10-04
Implementing: complex-numbers

Average-user:


Average-user:
How can I check the Travis build before upload?

petertseng:
> How can I check the Travis build before upload?

You may see the exact commands that the travis runs by looking at https://github.com/exercism/haskell/blob/master/.travis.yml#L23-L49 . You may run any or all of those commands on your own computer.

Average-user:
@petertseng Also, there is a way to don't **have** to follow the HLints suggestions? Sometimes the extra parentheses can help to read something, I believe.

petertseng:
> don't have to follow the HLints suggestions

https://github.com/exercism/haskell/blob/master/README.md#running-hlint

Average-user:
@petertseng I want your feedback about if it is Float the right type for this, maybe, I don't really know. Or maybe is it better tu use mixed types

petertseng:
I'm not sure I understood the question correctly. I think my answer is "as long as it passes the tests, it doesn't matter what the internal implementation details are", right?

config.json
```diff
@@ -115,6 +115,17 @@
       ]
     },
     {
+      "uuid": "40d4bb98-2e4a-4579-810e-g1230bc158e9",
+      "slug": "complex-numbers",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 2,
+      "topics": [
+        "Mathematics",
```
petertseng: I think I will feel the urge to remove this. I will feel bad having this exercise with "mathematics" topic, and none of the other exercises that should have "mathematics" topic have it. It isn't one that would affect a placement decision (the topics added in https://github.com/exercism/haskell/pull/476 ). 

I guess I would accept a PR that adds the mathematics topic to all the exercises that should have it.

config.json
```diff
@@ -115,6 +115,17 @@
...
+        "Number-theory"
```
petertseng: will any other exercise have this topic? if so, I'd like to see it added to them as well. If not, I'm not finding it particularly useful

exercises/complex-numbers/src/ComplexNumbers.hs
```diff
@@ -0,0 +1,41 @@
...
+ conjugate,
+ abs,
+ real,
+ imaginary,
+ mul,
+ add,
+ sub,
+ div) where
+
+import Prelude hiding (div, abs) -- you have to redifine this two functions
```
petertseng: I don't think the comment is necessary since we did not use it in any other exercise that has `hiding`, but if it is, it would have to be spelled "redefine these two functions"

exercises/complex-numbers/src/ComplexNumbers.hs
```diff
@@ -0,0 +1,41 @@
...
+
+-- Data definition -------------------------------------------------------------
+data Complex = Complex Float Float deriving (Eq, Show)
```
petertseng: I would not put this here, because students might think they are not allowed to change it. I would like to let the students have their creativity, to choose what they like. So, this line would probably have to be `data Complex = Dummy deriving (Eq, Show)`. Maybe even `data Complex = Dummy` if we can get away with it.

exercises/complex-numbers/test/Tests.hs
```diff
@@ -0,0 +1,191 @@
...
+          describe "div"       $ for_ divCases       $ testA div
+          describe "add"       $ for_ addCases       $ testA add
+          describe "sub"       $ for_ subCases       $ testA sub
+ where
+  testA f CaseA{..} = it descriptionA $ f number1A number2A `shouldBe` expectedA
+  testB f CaseB{..} = it descriptionB $ f number1B `shouldBe` expectedB
+  testC f CaseC{..} = it descriptionC $ f number1C `shouldBe` expectedC
+
+
+data CaseA = CaseA { descriptionA :: String
```
petertseng: Possible to have more descriptive names than A, B, C?

BinOp, UnOp, Scalar?

exercises/complex-numbers/test/Tests.hs
```diff
@@ -0,0 +1,191 @@
...
+conjugateCases =
+    [ CaseB { descriptionB = "Conjugate a purely real number"
+            , number1B     = Complex 5 0
+            , expectedB    = Complex 5 0
+            }
+    , CaseB { descriptionB = "Conjugate a purely imaginary number"
+            , number1B     = Complex 0 5
+            , expectedB    = Complex 0 (-5)
+            }
+    , CaseB { descriptionB = "Conjugate a number with real and  imaginary part"
```
petertseng: extra space between "and" and "imaginary"

exercises/complex-numbers/README.md
```diff
@@ -0,0 +1 @@
+# Pls regenerate this
```
petertseng: a reminder to myself to regenerate it before merging.

config.json
Average-user: Maybe I'll do that.

exercises/complex-numbers/test/Tests.hs
Average-user: Could, be. Don't know if it is worth with something this simple.

exercises/complex-numbers/src/ComplexNumbers.hs
Average-user: But what will happen if someone uses a definition like ```data Complex = Complex double double``` I think that would be ok, but the tests wouldn't work.

petertseng: That's true. You could look at https://github.com/exercism/haskell/blob/master/exercises/space-age/test/Tests.hs if you want help with that, I suppose. There are example https://github.com/exercism/haskell/blob/master/exercises/space-age/examples/success-double/src/SpaceAge.hs which uses doubles and https://github.com/exercism/haskell/blob/master/exercises/space-age/examples/success-rational/src/SpaceAge.hs which uses rational - both compile with those tests. You could possibly do something similar.

Average-user:
Don't really know why it fails. When I run ```bin/test-all-examples exercises/complex-numbers/``` works perfectly, but I'm guessing that has to do with the ```Complex = Dummy``` definition.

petertseng:
> Don't really know why it fails

Why not? Will you explain what part of https://travis-ci.org/exercism/haskell/jobs/286216812 you did not understand?

> When I run `bin/test-all-examples exercises/complex-numbers/` works perfectly

So that means the example is good.

> but I'm guessing that has to do with the `Complex = Dummy` definition.

Then you should read what https://travis-ci.org/exercism/haskell/jobs/286216812 says, and/or run `stack test` to test the stub, and/or run https://github.com/exercism/haskell/blob/master/bin/test-stub.

As you can see, Travis says

```
/home/travis/build/exercism/haskell/build/complex-numbers/stub/test/Tests.hs:48:30: error:
    Data constructor not in scope:
      Complex :: Integer -> Integer -> Complex
```

Indeed, there is no such data constructor, since it's `Complex = Dummy`.

So, how about a `Integer -> Integer -> Complex` function (maybe called `complex`) that the student has to implement? Then the test can use that function instead of using `Complex` (and thus forcing `Complex` to take `Integer -> Integer` as well). This is continuing the trend of allowing the internal representation to be changed at will.

exercises/complex-numbers/test/Tests.hs
```diff
@@ -0,0 +1,191 @@
...
+          describe "div"       $ for_ divCases       $ testA div
+          describe "add"       $ for_ addCases       $ testA add
+          describe "sub"       $ for_ subCases       $ testA sub
+ where
+  testA f CaseA{..} = it descriptionA $ f number1A number2A `shouldBe` expectedA
+  testB f CaseB{..} = it descriptionB $ f number1B `shouldBe` expectedB
+  testC f CaseC{..} = it descriptionC $ f number1C `shouldBe` fromInteger expectedC
+
+
+data CaseA = CaseA { descriptionA :: String
```
petertseng: I'll think about the names

petertseng:
Also, you're probably wondering why I didn't just tell you to do this at the same time as changing to `Complex = Dummy`.

It's because I forgot about it. Sorry for my bad memory.

Average-user:
@petertseng Maybe it was fine as it was before. Cause you say that a function  ```Integer -> Integer -> Complex``` Will force to take ```Complex``` with ```Integer -> Integer```.

But on this section of code:
```Haskell
real :: Complex -> Float
real = error "You need to implement this function"

imaginary :: Complex -> Float
imaginary = error "You need to implement this function"
```
It was already kind of implicit the ```Complex``` with ```Float -> Float```. So, taking that in count it might still be a good idea to just define ```data Complex = Complex Float Float```

I believe.

petertseng:
an interesting question will then be whether it makes sense to let it be a parameterized type, such as https://hackage.haskell.org/package/base/docs/Data-Complex.html has `Complex a`, and then it has `realPart :: Complex a -> a` and `imagPart :: Complex a -> a`. That could be considered.

Average-user:
You mean ```data Complex a = Complex a a``` ? If thats the case, what would be the better thing to put on the ```src/ComplexNumbers.hs``` definitions?

petertseng:
> You mean `data Complex a = Complex a a`?

You can choose what goes in the example. What you said is a possibility.

> what would be the better thing to put on the `src/ComplexNumbers.hs` definitions? Should be `data Complex a = Complex a a` given?

I still think it should be `data Complex a = Dummy`, see all instances of this in this repo

```
$ git grep Dummy            (10-10 23:13)
exercises/bank-account/src/BankAccount.hs:data BankAccount = Dummy
exercises/binary-search-tree/src/BST.hs:data BST a = Dummy deriving (Eq, Show)
exercises/clock/src/Clock.hs:data Clock = Dummy
exercises/custom-set/src/CustomSet.hs:data CustomSet a = Dummy deriving (Eq, Show)
exercises/forth/src/Forth.hs:data ForthState = Dummy
exercises/grade-school/src/School.hs:data School = Dummy
exercises/linked-list/src/Deque.hs:data Deque a = Dummy
exercises/matrix/src/Matrix.hs:data Matrix a = Dummy deriving (Eq, Show)
exercises/robot-name/src/Robot.hs:data Robot = Dummy
exercises/robot-simulator/src/Robot.hs:data Robot = Dummy
exercises/simple-linked-list/src/LinkedList.hs:data LinkedList a = Dummy deriving (Eq, Show)
exercises/zipper/src/Zipper.hs:data Zipper a = Dummy deriving (Eq, Show)
```

Average-user:
But if I leave the ```Complex = Dummy``` the previous error will remain.

petertseng:
> But if I leave the `Complex = Dummy` the previous error will remain.

So have a function that creates a `Complex a` from two `a`, and make the tests use it.

exercises/complex-numbers/examples/success-standard/src/ComplexNumbers.hs
```diff
@@ -5,34 +5,46 @@ module ComplexNumbers
...
 conjugate (Complex a b) = Complex a (-b)
 
-abs :: Complex -> Float
+abs :: Floating a => Complex a -> a
 abs (Complex a b) = sqrt (exp2 a + exp2 b)
 
 -- binary operators ------------------------------------------------------------
-mul :: Complex -> Complex -> Complex
+complex :: (a, a) -> Complex a
+complex (a, b) = Complex a b
```
petertseng: What about `complex = curry Complex` ?

petertseng: Oh sorry, that needs to be `complex = uncurry Complex`, get them mixed up too often.

exercises/complex-numbers/src/ComplexNumbers.hs
```diff
@@ -7,35 +7,39 @@ module ComplexNumbers
...
+real :: Num a => Complex a -> a
 real = error "You need to implement this function"
 
-imaginary :: Complex -> Float
+imaginary :: Num a => Complex a -> a
 imaginary = error "You need to implement this function"
 
 -- binary operators ------------------------------------------------------------
-mul :: Complex -> Complex -> Complex
+complex :: (a, a) -> Complex a
```
petertseng: I wouldn't put this in binary operators section. Binary operator implies `Complex a -> Complex a -> something`. This `complex` creates a complex number.

exercises/complex-numbers/examples/success-standard/src/ComplexNumbers.hs
Average-user: Forgot about that. I don't use neither of them too often.

exercises/complex-numbers/src/ComplexNumbers.hs
Average-user: Binary operator implies two arguments. But I see what you mean.

petertseng:
I think it is good now after I regenerate the README. I may need some days to do it.

Average-user:
Untill them, then.

petertseng:
All right, thanks for your fourth exercise contribution!

Average-user:
@petertseng 

> The files shouldn't be executable, but they all were. Let's not do that for the next PR.

Don't really understand what you mean with that. Also your last commit appears to have zero changes.

petertseng:
> Also your last commit appears to have zero changes.

I don't see any commit with no changes. I see my last commit is https://github.com/exercism/haskell/pull/601/commits/f44dd15dfbdf9ca84b0a55d5f7e7748f35e09050 but there are changes, all modes changed from 100755 to 100644.

Average-user:
@petertseng 
> I don't see any commit with no changes. I see my last commit is f44dd15 but there are changes, all modes changed from 100755 to 100644.

Right didn't see that. My bad.

petertseng:
> 100755 to 100644

(meaning I made them no longer executable)

config.json
```diff
@@ -115,6 +115,15 @@
       ]
     },
     {
+      "uuid": "40d4bb98-2e4a-4579-810e-g1230bc158e9",
```
petertseng: note: check whether g is OK for uuid

NobbZ: AFAIK UUID is usually 16 byte written in hexadecimal notation and separated into 5 groups. So it doesn't look like a valid UUID to me.

NobbZ: Has this UUID been generated by configlet?

NobbZ: I've opened a bug in the configlet repo.

Wether generated or not, at least the linter should have complained:

exercism/configlet#100

Average-user: @NobbZ It wasn't generated by configlet, my bad there. Also you might want to specify in the issue that it has already been fixed the UUID of this repo.

PR #604, opened 2017-10-13
implementing rotational-cipher

Average-user:


config.json
```diff
@@ -78,6 +78,15 @@
       ]
     },
     {
+      "uuid": "6pol4yme-6665-4ri6-sl21-4651u11cb4a4",
```
petertseng: I can't help but notice that this has letters outside of a-f, not sure that's acceptable for a UUID. Also need to check https://github.com/exercism/haskell/pull/601#discussion_r144734644.

exercises/rotational-cipher/README.md
```diff
@@ -0,0 +1 @@
+# Regenerate this pls!!!
```
petertseng: I remind myself to regenerate before merging.

exercises/rotational-cipher/examples/success-standard/src/RotationalCipher.hs
```diff
@@ -0,0 +1,9 @@
+module RotationalCipher (rotate) where
+
+import Data.List (elemIndex)
+
+rotate :: Int -> String -> String
+rotate r = map (f r ['a'..'z'] . f r ['A'..'Z'])
+  where f x abc e = case elemIndex e abc of
+                      Nothing  -> e
+                      (Just a) -> abc!!mod (a+x) 26
```
petertseng: huh, I didn't think the parentheses around `Just a` are required, but I don't know

config.json
Average-user: Do I make another PR for the complex-numbers UUID, or I include it in here?

petertseng: As long as it's in a separate commit, I would take it either way, but I suggest a different PR since I can merge that one quickly and can't merge this one quickly yet.

petertseng:
generated README, rebased on master, I think that's all we have to do

Average-user:
@petertseng I think so.

petertseng:
Thank you for your fifth exercise.

PR #609, opened 2017-10-21
Number Theory topic

Average-user:
These are the exercise that I think that should have "Number Theory" as a topic.

### DIFF
--- a/config.json
+++ b/config.json
@@ -59,6 +59,16 @@
       ]
     },
     {
+      "uuid": "6eba4eac-8395-4ad6-ac21-3651a11ab4b4",
+      "slug": "collatz-conjecture",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "Maybe"
+      ]
+    },
+    {
       "uuid": "85d77b8e-9b87-4d02-9fba-81f843bd66f1",
       "slug": "rna-transcription",
       "core": false,
--- a/config.json
+++ b/config.json
@@ -65,12 +65,22 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-        "Maybe"
+        "Maybe",
+        "Number Theory"
       ]
     },
     {
       "uuid": "6eba4eac-6665-4ad6-ac21-3651a11ab4b4",
       "slug": "isogram",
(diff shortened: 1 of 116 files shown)


petertseng:
Too hard to read the diff and determine which exercises got the topic. Easier to `ruby -rjson -e 'puts JSON.parse(File.read("config.json"))["exercises"].select { |e| e["topics"]&.include?("Number Theory") }.map { |e| e["slug"] }' | sort`
collatz-conjecture
difference-of-squares
largest-series-product
nth-prime
pascals-triangle
prime-factors
pythagorean-triplet
sieve
sum-of-multiples
triangle

petertseng:
I think I agree with all of these. I'm unsure about largest-series-product since it's just multiplying numbers and comparing them, but whatever.

I think complex-numbers should have.

Maybe grains should have, but I'm not sure.

Average-user:
@petertseng Should I do a squash?

petertseng:
> Should I do a squash?

Optional, because if you don't, I will. But that also means you are welcome to do it.

Average-user:
@petertseng How do you squash? ``` reset --soft```?

petertseng:
> How do you squash?

I interpreted this question as asking, how do **I** do it? I use `git rebase -i master`.

You can use whatever you want that achieves the same effect. I haven't used `reset --soft` before but if it works, it works.

Average-user:
@petertseng  Ok thanks.

>but if it works, it works.

Looks to be your style jeje.

petertseng:
collatz-conjecture
complex-numbers
difference-of-squares
grains
nth-prime
pascals-triangle
prime-factors
pythagorean-triplet
sieve
sum-of-multiples
triangle

petertseng:
I would think either both grains and largest-series-product have the topic, or neither do. I don't know which is right since I don't know what number theory is. My guess is no.

Weren't you going to squash??

Average-user:
@petertseng I think you were right about taking off largest-series-product, and adding grains. I know that to solve grains you don't really do much, but it still a classic problem, I just didn't realize that grains is the name for it.

petertseng:
Oh, so you think this list is good as-is. OK then.

PR #610, opened 2017-10-23
Normalize the config file to minimize noise later

kytrinyx:
We will be scripting a number of changes to the track configuration.
This runs configlet fmt, which ensures that any further changes can be minimized.

See https://github.com/exercism/meta/issues/95

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,836 +1,793 @@
 {
-  "language": "Haskell",
   "active": true,
-  "solution_pattern": "example.*[.]hs",
-  "ignore_pattern": "example",
   "exercises": [
     {
-      "uuid": "c97d9dda-c7ee-4595-b447-693559d03ba5",
+      "core": false,
+      "difficulty": 1,
       "slug": "hello-world",
-      "core": false,
+      "topics": null,
       "unlocked_by": null,
-      "difficulty": 1,
-      "topics": [
-      ]
+      "uuid": "c97d9dda-c7ee-4595-b447-693559d03ba5"
     },
     {
-      "uuid": "0ad7d756-1cbd-44f4-a64f-0dc7e9eb40f4",
+      "core": false,
+      "difficulty": 1,
       "slug": "leap",
-      "core": false,
+      "topics": null,
       "unlocked_by": null,
-      "difficulty": 1,
-      "topics": [
-      ]
+      "uuid": "0ad7d756-1cbd-44f4-a64f-0dc7e9eb40f4"
     },
     {
-      "uuid": "6cbb7841-9d96-4528-88e3-6e98a450fd7c",
+      "core": false,
+      "difficulty": 1,
       "slug": "space-age",
-      "core": false,
+      "topics": null,
       "unlocked_by": null,
-      "difficulty": 1,
-      "topics": [
-      ]
+      "uuid": "6cbb7841-9d96-4528-88e3-6e98a450fd7c"
     },
     {
-      "uuid": "e570c730-12e5-4b9e-adb0-eae85360e1b5",
+      "core": false,
+      "difficulty": 1,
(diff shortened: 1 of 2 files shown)


config.json
```diff
@@ -1,836 +1,793 @@
 {
-  "language": "Haskell",
```
petertseng: This seems too important to shunt to the bottom of the file. I have evidence in https://github.com/exercism/bash/commit/6fabd6f26c76e799335ee43434826f9fac4ea3a9 and https://github.com/exercism/bash/pull/91 that this caused a mistaken assumption that the `language` attribute was absent. If this should indeed move, I need to hear a way to prevent others from making the same mistake.

config.json
```diff
@@ -1,836 +1,793 @@
...
     }
   ],
-  "foregone": [
-
-  ]
-}
+  "foregone": [],
+  "ignore_pattern": "example",
+  "language": "Haskell",
+  "solution_pattern": "example.*[.]hs"
```
petertseng: should it not be the case that ignore_pattern and solution_pattern are next to each other, instead of having slug interposed in between?

config.json
```diff
@@ -1,836 +1,793 @@
...
-      "uuid": "c97d9dda-c7ee-4595-b447-693559d03ba5",
-      "slug": "hello-world",
       "core": false,
-      "unlocked_by": null,
       "difficulty": 1,
-      "topics": [
-      ]
+      "slug": "hello-world",
+      "topics": null,
+      "unlocked_by": null,
```
petertseng: shouldn't it be the case that unlocked_by and core are next to each other? They seem too closely related to separate in this manner.

config.json
```diff
@@ -1,836 +1,793 @@
...
   "exercises": [
     {
-      "uuid": "c97d9dda-c7ee-4595-b447-693559d03ba5",
-      "slug": "hello-world",
       "core": false,
-      "unlocked_by": null,
       "difficulty": 1,
-      "topics": [
-      ]
+      "slug": "hello-world",
```
petertseng: slug seems important. it seems like it should come first or second (behind uuid)

config.json
petertseng: should it not be the case that ignore_pattern and solution_pattern are next to each other, instead of having language interposed in between?

config.json
Insti: slug first please.

petertseng:
I get it, I should not expect in JSON to have any sort of ordering among my keys, and so "might as well pick alphabetical it's as good as any".

I counter that the JSON has been serialised to a file and presumably this file is at least partially for human consumption: we ask humans to make changes to config.json when adding a new exercise, for example, and do not provide any tooling to do so!

As a result, it is my obligation to defend the humans and say that, as represented in the JSON files in the repository, we should not accept a change that impairs the human's ability to understand.

The claim of this PR is that it will make scripted changes easier. Well, what I'm going to do in such an instance is the equivalent of:

* run `configlet fmt`, record the diff.
* make the desired scripted change
* revert the diff from `configlet fmt`, leaving only the scripted change in place.

I accept the cost of doing this myself for the tracks that I maintain, instead of reformatting config files.

kytrinyx:
Would you please weigh with these observations in exercism/meta#95? They are useful to the broader discussion.

kytrinyx:
There's a discussion happening in exercism/meta#95 about how this normalization degrades the readability of the configs.

That discussion has some good suggestions for improvements to the serializer.
I'd appreciate it if you would weigh in on that discussion with any suggestions for improving how we handle configs, consistency, and normalization.

Meanwhile, I'm going to close this and wait for that discussion to play out.

I'm sorry about the noise, and any extra work that I've caused you.

PR #612, opened 2017-10-23
Implementing: isbn-verifier

Average-user:


config.json
```diff
@@ -23,6 +23,16 @@
       ]
     },
     {
+      "uuid": "e5a0bed3-90ca-4caa-a73f-2124940f84c9",
+      "slug": "isbn-verifier",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "Strings"
```
petertseng: probably make a PR adding strings to all exercises that have it

petertseng:
I generally think this is good, but I don't like some of the problem-specifications cases so I'm going to get that sorted before merging it here.

Average-user:
@What do you mean with sorted? How is that change the fact of you disliking it?

petertseng:
3-598-2X507-0 should be 3-598-2X507-9, otherwise it doesn't check that X is only valid as a check digit

3-598-21507-XA should be 3-598-21507-XX, otherwise it doesn't check too long ISBN

Average-user:
@petertseng But how is that sorting exactly?

petertseng:
Well, I'm getting it fixed, am I not?

Average-user:
Ok. Should someone do a PR in the originak tests cases?

config.json
Average-user: It is ok if I make that after this get merged? If that ever happens.

petertseng: Yes of course, since doing it after is one of the two possible options:

1. remove string from this PR, make string PR after this PR is merged
2. make string PR and it is merged before this PR is merged, in which case keeping string here is OK

petertseng:
Well it looks like I can merge https://github.com/exercism/problem-specifications/pull/993 soon.

petertseng:
OK https://github.com/exercism/problem-specifications/pull/993 is merged, you can update the test cases now

config.json
Average-user: I was thinking on doing the first when I removed Strings.

Average-user:
@petertseng Ok, in a moment

exercises/isbn-verifier/package.yaml
```diff
@@ -0,0 +1,20 @@
+name: isbn-verifier
+version: 1.0.0.1
```
petertseng: make it 1.1.0.1

Average-user: Thats because the update ?

petertseng: indeed, because that's what https://github.com/exercism/problem-specifications/blob/master/exercises/isbn-verifier/canonical-data.json#L3 says.

petertseng:
thank you for implementing your seventh exercise!

Average-user:
Thanks to you too. Another one some time?

petertseng:
Well, whenever you or anyone else sees fit to implement them.

PR #615, opened 2017-10-29
Implementing exercise binary-search

samosaara:
On the works!

### DIFF
--- a/exercises/binary-search/README.md
+++ b/exercises/binary-search/README.md
@@ -1,0 +1,79 @@
+# Binary Search 
+
+Implement a binary search algorithm.
+
+Searching a sorted collection is a common task. A dictionary is a sorted list of word definitions. Given a word, one can find its definition. A telephone book is a sorted list of people's names, addresses, and telephone numbers. Knowing someone's name allows one to quickly find their telephone number and address.
+
+If the list to be searched contains more than a few items (a dozen, say) a binary search will require far fewer comparisons than a linear search, but it imposes the requirement that the list be sorted.
+
+In computer science, a binary search or half-interval search algorithm finds the position of a specified input value (the search "key") within an array sorted by key value.
+
+In each step, the algorithm compares the search key value with the key value of the middle element of the array.
+
+If the keys match, then a matching element has been found and its index, is returned.
+
+Otherwise, if the search key is less than the middle element's key, then the algorithm repeats its action on the sub-array to the left of the middle element or, if the search key is greater, on the sub-array to the right.
+
+If the remaining array to be searched is empty, then the key cannot be found in the array and Nothing is returned.
+
+A binary search halves the number of items to check with each iteration, so locating an item (or determining its absence) takes logarithmic time. A binary search is a dichotomic divide and conquer search algorithm.
+
+
+## Getting Started
+
+For installation and learning resources, refer to the
+[exercism help page](http://exercism.io/languages/haskell).
+
+## Running the tests
+
+To run the test suite, execute the following command:
+
+```bash
+stack test
+```
+
+#### If you get an error message like this...
+
+```
+No .cabal file found in directory
+```
+
+You are probably running an old stack version and need
+to upgrade it.
+
+#### Otherwise, if you get an error message like this...
+
+```
+No compiler found, expected minor version match with...
+Try running "stack setup" to install the correct GHC...
+```
+
+Just do as it says and it will download and install
+the correct compiler version:
+
+```bash
+stack setup
+```
+
+## Running *GHCi*
+
+If you want to play with your solution in GHCi, just run the command:
+
+```bash
+stack ghci
+```
+
+## Feedback, Issues, Pull Requests
+
+The [exercism/haskell](https://github.com/exercism/haskell) repository on
+GitHub is the home for all of the Haskell exercises.
+
+If you have feedback about an exercise, or want to help implementing a new
+one, head over there and create an issue.  We'll do our best to help you!
+
+## Source
+
+Wikipedia [https://en.wikipedia.org/wiki/Isogram](https://en.wikipedia.org/wiki/Isogram)
+
+## Submitting Incomplete Solutions
+It's possible to submit an incomplete solution so you can see how others have completed the exercise.
--- a/exercises/binary-search/stack.yaml
+++ b/exercises/binary-search/stack.yaml
@@ -1,0 +1,1 @@
+resolver: lts-8.21
--- a/exercises/binary-search/stack.yaml
+++ b/exercises/binary-search/stack.yaml
@@ -1,1 +1,1 @@
-resolver: lts-8.21
+resolver: lts-9.11
--- a/exercises/binary-search/test/Tests.hs
+++ b/exercises/binary-search/test/Tests.hs
@@ -1,0 +1,72 @@
+{-# LANGUAGE RecordWildCards #-}
+
+import Data.Foldable     (for_)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
+import BinarySearch (binarySearch)
+
+main :: IO ()
+main = hspecWith defaultConfig {configFastFail = True} specs
+
+specs :: Spec
+specs = describe "foundElement" $ for_ cases test
+  where
+    test Case{..} = it description assertion
+      where
+        assertion = binarySearch input value `shouldBe` expected
+
+
+data Case a = Case { description :: String
+                 , input       :: [a]
+                 , value       :: a
+                 , expected    :: Maybe Int
+                 }
+
+cases :: [Case Int]
+cases = [ Case { description = "finds a value in an array with one element"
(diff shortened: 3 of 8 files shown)


NobbZ:
What kind of Container will you use?

If it is a non continious memory store (eg. a `List`), the binary search is nonsense (binary search gets even worse than a linear search due to subslicing).

samosaara:
Does it even matter bro? Aren't you overthinking it? it's not the point to make the better possible way to search an element in an array, rather then figuring out how to implement the algorithm, even if I knew what the heck were you talking about and indeed it were a lot slower, I don't think it would matter anyway, but that's just IMHO, what do you even mean? It's going to be slow?

NobbZ:
Yes, it does matter.

The README sells the binary search as superior while it isn't for lists, so either implement the exercise in a way that it is not using a List but some kind of Array, or don't do it at all!

At least that's my opinion, but I am not a maintainer here, so at the end @exercism/haskell descides.

Average-user:
@samosaara This is supposed to be an example, therefor, the implementation matters.

NobbZ:
PS: As far as I can remember the best possible implementations of binary search in a linked List ar O(n log n), but most are near to O(n²).

In worst case (element is found at the very last possible step) you'll have copied parts of the list log n times, you have to iterate each copied slice 1 to 1.5 times on naiv implementations (first half time for the element in the middle, another half time when you need left subslice, or a full time when needing the right subslice).

Average-user:
@NobbZ [Sequences](http://hackage.haskell.org/package/containers-0.2.0.1/docs/Data-Sequence.html) can do the work here right?

NobbZ:
Yes, they can perform better than Lists when using their `splitAt`, `drop`, and `take` functions.

samosaara:
Oh so the whole discussion was to have the best ever, swag, 1337 h@x0r, code golf, solution, well, ok whatever I'll use sequences. Guess I missed the point, thought you guys were complaining about the exercise itself.

Average-user:
@samosaara  It isn't about having the best ever, thats impossible. The idea is to show when is useful to actually use a binary search algorithm, and apparently with lists it isn't.

NobbZ:
I'm fine with the exercise, as long as the binary search is used with a correct datatype to not suggest bad practices and sell them as a good one.

petertseng:
I thought https://hackage.haskell.org/package/array-0.5.2.0/docs/Data-Array.html would have been sufficient. Wouldn't O(1) access be more suited to showing binary search than `O(log(min(i, n-i)))` that I see in https://hackage.haskell.org/package/containers-0.5.10.2/docs/Data-Sequence.html#v:index ? Or is even that sufficient? I guess since there would be at most `O(log n)` bisections one would call the runtime O(log^2 n)`, so you are saying that is good enough, right? Even if it is good enough, would it still be better to use an Array? If I miss something, please educate me, thank you.

Average-user:
@petertseng I just said sequences because it is what I've used. Probably arrays work too, but I can't find where it says that the access is constant.

NobbZ:
If Arrays are O(1) in random reading and also can be easily sliced, then of course they are perfect for this algorithm, but I can't find a slicing accessor in the docs…

petertseng:
> I can't find where it says that the access is constant.
> If Arrays are O(1) in random reading

Okay, y'all are right, https://hackage.haskell.org/package/array-0.5.2.0/docs/Data-Array.html only specifies that "a programmer may reasonably expect rapid access to the components" without specifically explaining how rapid is the rapid referred to.

> and also can be easily sliced

Shrug. Can just keep track of the left/right index of the subarray you're searching, and thereby avoid a need to slice, just keep the same array throughout.

samosaara:
Okay, so, I guess I'm not that proficient in Haskell to implement with the kinda of quality you guys are requesting, for the example, so I guess I would need help (accepting PR on my fork XD) or at least time to implement this :sweat_smile:

Average-user:
@samosaara I think that you have time. And the only "challenge" here, it is to get used to Data.Array.

samosaara:
Oh wait but just to be sure, I should do change the signature of the exercise itself or just convert back and forth in the example implementation?

Average-user:
@samosaara  Probably the first, so that the users that are going to implement this, will be forced to use a container in which this is efficient.

exercises/binary-search/stack.yaml
```diff
@@ -0,0 +1 @@
+resolver: lts-8.21
```
petertseng: before I forget: I moved to lts-9.11 now so this should change too

samosaara:
Well, there is a problem, I cant implement one of the tests because there is no way to represent an empty Data.Array, Do I ignore it? What should I do?

petertseng:
I was able to find http://mail.haskell.org/pipermail/haskell-cafe/2004-October/007098.html by asking my preferred search engine about empty arrays, but I can't say whether this information is still up to date since it is ten years old. Is it in fact out of date?

exercises/binary-search/test/Tests.hs
```diff
@@ -18,55 +19,56 @@ specs = describe "foundElement" $ for_ cases test
...
                }
         , Case { description = "a value larger than the array's largest value is not included"
-               , input       = [1, 3, 4, 6, 8, 9, 11]
+               , input       = array1
                , value       = 13
                , expected    = Nothing
                }
         , Case { description = "nothing is included in an empty array"
-               , input       = []
+               , input       = listArray (1, 0) [] -- There is no way to represent an empty array
```
samosaara: Ignore the dumb comment, there is. it is working

samosaara:
Actually I have doubts if I got that right, I didn't find a way to slit the array itself, and I think the way I'm doing it there is not much of a point, because I'm just making it into lists so I can split them... Which was the whole problem of this discussion, that doing that was slow and stuff. Did I miss a method or did I actually done it?

NobbZ:
In general (so regardless of the language) these algorithms avoid to copy the array. The best way to do this is to pass it around **as is**, and provide 2 additional parameters which determine upper and lowerbound of the active subslice.

```
find xs e = go (lowerIdx xs) (upperIdx xs) xs
  where
    go l u xs | xs[(l + u) / 2] == e = True
    go …
```

exercises/binary-search/test/Tests.hs
```diff
@@ -0,0 +1,74 @@
...
+specs :: Spec
+specs = describe "foundElement" $ for_ cases test
+  where
+    test Case{..} = it description assertion
+      where
+        assertion = binarySearch input value `shouldBe` expected
+
+
+data Case a = Case { description :: String
+                 , input       :: Array Int a
```
petertseng: I would indent this line two spaces further so that the comma lines up with the open bracket on the previous row.

exercises/binary-search/test/Tests.hs
```diff
@@ -0,0 +1,74 @@
...
+                 , value       :: a
```
petertseng: > I would indent this line two spaces further so that the comma lines up with the open bracket on the previous row.

same

exercises/binary-search/test/Tests.hs
```diff
@@ -0,0 +1,74 @@
...
+                 , expected    :: Maybe Int
```
petertseng: > I would indent this line two spaces further so that the comma lines up with the open bracket on the previous row.

same

exercises/binary-search/test/Tests.hs
```diff
@@ -0,0 +1,74 @@
...
+                 }
```
petertseng: > I would indent this line two spaces further so that the comma lines up with the open bracket on the previous row.

same except now I'm talking about the closing bracket

exercises/binary-search/test/Tests.hs
```diff
@@ -0,0 +1,74 @@
...
+               , value       = 13
+               , expected    = Nothing
+               }
+        , Case { description = "nothing is included in an empty array"
+               , input       = listArray (1, 0) []
+               , value       = 1
+               , expected    = Nothing
+               }
+        ]
+        where array1 = listArray (0, 6) [1, 3, 4, 6, 8, 9, 11]
```
petertseng: I'm considering whether this should go above the cases so that I can see it before I see the cases it's going to be used in. I'm not sure, so let me know what you think about it.

exercises/binary-search/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+The [exercism/haskell](https://github.com/exercism/haskell) repository on
+GitHub is the home for all of the Haskell exercises.
+
+If you have feedback about an exercise, or want to help implementing a new
+one, head over there and create an issue.  We'll do our best to help you!
+
+## Source
+
+Wikipedia [https://en.wikipedia.org/wiki/Isogram](https://en.wikipedia.org/wiki/Isogram)
```
petertseng: why isogram?

sshine:
Hello @samosaara! I went and improved on this solution in #807, since it was so close to being finished.

A few complexities have been added to the track since you started this PR, so I've addressed those.

Thanks for doing all the groundwork.

PR #619, opened 2017-11-05
stack.yaml: lts-8.21 -> lts 9.11

petertseng:
Last update:
https://github.com/exercism/haskell/pull/571
Four months ago. Seems like it's time to make sure everything still
works.

https://www.stackage.org/diff/lts-8.21/lts-9.11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ cache:
     - $HOME/.foldercache # Per exercise `.stack-work` cache.
 
 env:
- - RESOLVER="lts-8.21" CURRENT="YES" # Equal to each stack.yaml.
+ - RESOLVER="lts-9.11" CURRENT="YES" # Equal to each stack.yaml.
  - RESOLVER="nightly"                # Latest nightly snapshot.
 
 matrix:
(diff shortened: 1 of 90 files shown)


AtelesPaniscus:
I have a question regarding this exercise.  I apologise if this is not the right place to ask.

From the problem description it seems clear to me that dashes must appear at certain offsets in an ISBN-10 number (or not at all) and that X, but not x, may appear as the final character.

There are no test cases to cover either possibility (none that fail and none that pass).  One submission after another appears to simply assume whatever is most convenient for the approach the student has taken to solving the problem.

In other words, students focus not on the problem but on passing the tests.

My question ?  Which test cases are missing ?

petertseng:
> I have a question regarding this exercise.

Sorry, I don't know which exercise you're referring to.

> From the problem description it seems clear to me that dashes must appear at certain offsets in an ISBN-10 number (or not at all) and that X, but not x, may appear as the final character.

> There are no test cases to cover either possibility (none that fail and none that pass).

Then they should get added to https://github.com/exercism/problem-specifications/blob/master/exercises/isbn-verifier/canonical-data.json

PR #627, opened 2017-11-29
Implementing rail-fence-cipher

Average-user:
I'm not sure about the types on ```src/RailFenceCipher.hs```

Should I change them to ``` f :: Int -> String -> String``` or change the ones from the tests?

### DIFF
--- a/exercises/rail-fence-cipher/README.md
+++ b/exercises/rail-fence-cipher/README.md
@@ -1,1 +1,119 @@
-## pls regenerate this!
+# Rail Fence Cipher
+
+Implement encoding and decoding for the rail fence cipher.
+
+The Rail Fence cipher is a form of transposition cipher that gets its name from
+the way in which it's encoded. It was already used by the ancient Greeks.
+
+In the Rail Fence cipher, the message is written downwards on successive "rails"
+of an imaginary fence, then moving up when we get to the bottom (like a zig-zag).
+Finally the message is then read off in rows.
+
+For example, using three "rails" and the message "WE ARE DISCOVERED FLEE AT ONCE",
+the cipherer writes out:
+
+```text
+W . . . E . . . C . . . R . . . L . . . T . . . E
+. E . R . D . S . O . E . E . F . E . A . O . C .
+. . A . . . I . . . V . . . D . . . E . . . N . .
+```
+
+Then reads off:
+
+```text
+WECRLTEERDSOEEFEAOCAIVDEN
+```
+
+To decrypt a message you take the zig-zag shape and fill the ciphertext along the rows.
+
+```text
+? . . . ? . . . ? . . . ? . . . ? . . . ? . . . ?
+. ? . ? . ? . ? . ? . ? . ? . ? . ? . ? . ? . ? .
+. . ? . . . ? . . . ? . . . ? . . . ? . . . ? . .
+```
+
+The first row has seven spots that can be filled with "WECRLTE".
+
+```text
+W . . . E . . . C . . . R . . . L . . . T . . . E
+. ? . ? . ? . ? . ? . ? . ? . ? . ? . ? . ? . ? .
+. . ? . . . ? . . . ? . . . ? . . . ? . . . ? . .
+```
+
+Now the 2nd row takes "ERDSOEEFEAOC".
+
+```text
+W . . . E . . . C . . . R . . . L . . . T . . . E
+. E . R . D . S . O . E . E . F . E . A . O . C .
+. . ? . . . ? . . . ? . . . ? . . . ? . . . ? . .
+```
+
+Leaving "AIVDEN" for the last row.
+
+```text
+W . . . E . . . C . . . R . . . L . . . T . . . E
+. E . R . D . S . O . E . E . F . E . A . O . C .
+. . A . . . I . . . V . . . D . . . E . . . N . .
+```
+
+If you now read along the zig-zag shape you can read the original message.
+
+
+## Getting Started
+
+For installation and learning resources, refer to the
+[exercism help page](http://exercism.io/languages/haskell).
+
+## Running the tests
+
+To run the test suite, execute the following command:
+
+```bash
+stack test
+```
+
+#### If you get an error message like this...
+
+```
+No .cabal file found in directory
+```
+
+You are probably running an old stack version and need
+to upgrade it.
+
+#### Otherwise, if you get an error message like this...
+
+```
+No compiler found, expected minor version match with...
+Try running "stack setup" to install the correct GHC...
+```
+
+Just do as it says and it will download and install
+the correct compiler version:
+
+```bash
+stack setup
+```
+
+## Running *GHCi*
+
+If you want to play with your solution in GHCi, just run the command:
+
+```bash
+stack ghci
+```
+
+## Feedback, Issues, Pull Requests
+
+The [exercism/haskell](https://github.com/exercism/haskell) repository on
+GitHub is the home for all of the Haskell exercises.
+
+If you have feedback about an exercise, or want to help implementing a new
+one, head over there and create an issue.  We'll do our best to help you!
+
+## Source
+
+Wikipedia [https://en.wikipedia.org/wiki/Transposition_cipher#Rail_Fence_cipher](https://en.wikipedia.org/wiki/Transposition_cipher#Rail_Fence_cipher)
+
+## Submitting Incomplete Solutions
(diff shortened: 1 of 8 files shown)


Average-user:
Or maybe change the example to:
```Haskell
encode' :: (Ord a) => Int -> [a] -> [a]
encode' n xs = concat [[a | (a, b) <- zip xs c, b == i] | i <- [1..n]]
  where c = cycle $ [1..n] ++ [n-1,n-2..2]

encode :: Int -> String -> String
encode = encode'

decode :: Int -> String -> String
decode n xs = map snd $ sort $ zip (encode' n [0..length xs - 1]) xs
```

exercises/rail-fence-cipher/examples/success-standard/src/RailFenceCipher.hs
```diff
@@ -0,0 +1,10 @@
+module RailFenceCipher (encode, decode) where
+
+import Data.List (sort)
+
+encode :: (Ord a) => Int -> [a] -> [a]
```
petertseng: Will you explain to me why `(Ord a) =>` necessary? I am not seeing why, but I may have just missed it.

exercises/rail-fence-cipher/examples/success-standard/src/RailFenceCipher.hs
```diff
@@ -0,0 +1,10 @@
...
+encode n xs = concat [[a | (a, b) <- zip xs c, b == i] | i <- [1..n]]
+  where c = cycle $ [1..n] ++ [n-1,n-2..2]
+
+decode :: (Ord a) => Int -> [a] -> [a]
```
petertseng: > Will you explain to me why `(Ord a) =>` necessary? I am not seeing why, but I may have just missed it.

same question

exercises/rail-fence-cipher/src/RailFenceCipher.hs
```diff
@@ -0,0 +1,7 @@
+module RailFenceCipher (encode, decode) where
+
+encode :: Int -> [a] -> [a]
```
petertseng: I see you ask whether to change this. I would use `String` here simply because that is all tests will use, and it's fine if students realise it can be generalised and therefore change it back to `[a]`. I would therefore also not change the example, minus the comment I make on it.

exercises/rail-fence-cipher/src/RailFenceCipher.hs
```diff
@@ -0,0 +1,7 @@
...
+encode = "You need to implement this function!"
+
+decode :: Int -> [a] -> [a]
```
petertseng: > I see you ask whether to change this. I would use `String` here simply because that is all tests will use, and it's fine if students realise it can be generalised and therefore change it back to `[a]`. I would therefore also not change the example, minus the comment I make on it.

same comment

exercises/rail-fence-cipher/README.md
```diff
@@ -0,0 +1 @@
+## pls regenerate this!
```
petertseng: I remind myself to regenerate.

exercises/rail-fence-cipher/src/RailFenceCipher.hs
```diff
@@ -0,0 +1,7 @@
+module RailFenceCipher (encode, decode) where
+
+encode :: Int -> [a] -> [a]
+encode = "You need to implement this function!"
```
petertseng: don't forget `error`!

exercises/rail-fence-cipher/src/RailFenceCipher.hs
```diff
@@ -0,0 +1,7 @@
...
+
+decode :: Int -> [a] -> [a]
+decode = "You need to implement this function!"
```
petertseng: > don't forget `error`!

same

exercises/rail-fence-cipher/src/RailFenceCipher.hs
Average-user: But why make it less general when it works fine with other types?

exercises/rail-fence-cipher/examples/success-standard/src/RailFenceCipher.hs
Average-user: Seems like it isn't, I think it was from a previous version

exercises/rail-fence-cipher/examples/success-standard/src/RailFenceCipher.hs
Average-user: same answer

exercises/rail-fence-cipher/src/RailFenceCipher.hs
Average-user: Ahh ok, dint read this before

exercises/rail-fence-cipher/src/RailFenceCipher.hs
petertseng: One thing I see is that https://github.com/exercism/problem-specifications/blob/master/exercises/rail-fence-cipher/description.md only talks about working with strings, thus that is the extent of the problem being specified. Therefore, any function that I write that adheres to this specification **need** only operate with strings. It is of course great if it works with other types. I imagine most solutions will. The possible complaint with putting `[a]` here is that "why is this stub asking for more than the problem specification is asking for?" the possible complaint with putting `String` is "this stub is unnecessarily restricting this function to only work with the specification, but it could do more"

It's not a perfect analogue, but there are text like https://github.com/exercism/haskell/blob/master/exercises/leap/HINTS.md or https://github.com/exercism/haskell/blob/master/exercises/series/HINTS.md that we can add.

I don't remember right now whether this track has a policy or other examples of either choice ("give minimal signature that satisfies the specification and let students expand" or "give the most general signature that a reasonable solution should be able to achieve") and I can't go looking for examples now, so I make my recommendation based on what I can see now, but I would change if there is an example of something we have done in the past.

exercises/rail-fence-cipher/src/RailFenceCipher.hs
Average-user: whoops!

exercises/rail-fence-cipher/src/RailFenceCipher.hs
Average-user: Well, I thing with ```Int -> String -> String``` would be just fine, afterall

exercises/rail-fence-cipher/src/RailFenceCipher.hs
Average-user: > whoops!

same :(

exercises/rail-fence-cipher/examples/success-standard/src/RailFenceCipher.hs
Average-user: Wow my bad there, seems like in the sort is needed, cause it takes on count the second argument.

Average-user: Should I leave it as it is? Or use sortBy?

Average-user: to sum up, I think that there are 2 options:
```Haskell
decode :: Int -> [a] -> [a]
decode n xs = map snd $ sortBy (compare `on` fst) zippedL
  where zippedL = zip (encode n [0..length xs - 1]) xs
```
or what we already have.

petertseng: I see, because you are using `sort` in `decode`.

In contrast to the stub, where I argued that we only need show the student what was specified by the problem description, here in the example I would in fact prefer not to add constraint that isn't necessary. Because the example is doing well by expanding from `String` to some other type. But if it's stopping at `(Ord a)` then it's only going halfway! It should go all the way!

petertseng: Thus, let us use `sortBy` here!

Average-user:
> let's do it.

Do what?

Average-user:
ohh, I see.

exercises/rail-fence-cipher/package.yaml
```diff
@@ -0,0 +1,20 @@
+name: rail-fence-cipher
+version: 1.1.0.1
```
petertseng: my mistake. it was 1.0.1 not 1.1.0. Fixed in #634

PR #629, opened 2017-12-04
travis: Ensure READMEs are updated 

petertseng:
I don't feel like checking this manually.

Lifted straight from the Rust track.
https://github.com/exercism/rust/pull/386


petertseng:
Note that although this will cause CI to fail if new exercises contain just a README that says "regenerate this README", the offer is still on the table that I can do this for others if desired. It just makes it easy for me to check whether I regenerated correctly.

Average-user:
>Note that although this will cause CI to fail if new exercises contain just a README that says "regenerate this README", the offer is still on the table that I can do this for others if desired. It just makes it easy for me to check whether I regenerated correctly.

Couldn't you specify ```regenerate this``` as valid README?

petertseng:
> Couldn't you specify `regenerate this` as valid README?

I wish there were a way to make it so that it would let Travis succeeds BUT that I still get a reminder not to merge it unless I've actually regenerated. I don't have a way to do this yet, I don't think, so the danger if I let `regenerate this` pass Travis is that I'll then forget to regenerate.

petertseng:
The disadvantage is of course that it is hard for someone who submits "regenerate this" to tell whether the rest of the exercise has been completed correctly without errors because then they **must** read the test output. I acknowledge this is true and that I don't have a best solution for this yet, other than the two options of "regenerate the README" or "read the test output to see whether README is the only problem". This may mean I have to regenerate READMEs sooner in the review process, I can do that.

Also, since a previous reason I discouraged contributors generating READMEs is "even if you do it, I have to do it a second time just to make sure you did it right", now since CI is checking, it will be OK for contributors to generate READMEs since I will not have to do it a second time. So that's an option.

PR #644, opened 2018-01-15
config.json: Reorder exercises

### LINKED ISSUE #639: Exercise ordering: isbn-verifier and rail-fence-cipher
`isbn-verifier` and `rail-fence-cipher` are way harder than `space-age`.

This seems to be causing a lot of people to quit early in the track. Take a look at [the numbers](http://www.exercism.io/stats/haskell).

Also, **it would be expected to see a big spike in the new exercises**, because of old-time users finishing just the new ones, but what we see the opposite there, so I guess we have a problem with the ordering.

In the past months, a lot of exercises where added, most of them with difficult *1* or *2*. Maybe it is time check the ordering to make the progression smoother.

I'm sorry but I'm just reporting the problem here. I'll probably not have the time to propose a new ordering soon.

petertseng:
Closes #639

* (collatz-conjecture stays where it is, 1)
* perfect-numbers 2
* isogram 2
* diamond 2
* (twelve-days stays where it is, 3)
* isbn-verifier 3
* run-length-encoding 3
  * this is also helpful because it places two exercises between
    twelve-days and beer-song, avoiding boredom from repeated
    text-generation exercfise
* complex-numbers 3
  * instructions are given, but there are many of them
* rotational-cipher 5
* rail-fence-cipher 6

I don't have justifications for many of these numbers; they're just
arbitrary feelings.

### DIFF
--- a/config.json
+++ b/config.json
@@ -23,24 +23,6 @@
       ]
     },
     {
-      "uuid": "e5a0bed3-90ca-4caa-a73f-2124940f84c9",
-      "slug": "isbn-verifier",
-      "core": false,
-      "unlocked_by": null,
-      "difficulty": 1,
-      "topics": [
-      ]
-    },
-    {
-      "uuid": "5bf7612c-427e-484d-bc67-8553fac4f64d",
-      "slug": "rail-fence-cipher",
-      "core": false,
-      "unlocked_by": null,
-      "difficulty": 1,
-      "topics": [
-      ]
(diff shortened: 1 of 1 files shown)


petertseng:
Not perfect, but I've also historically not been good at making a curriculum and I don't really want to put too much thought into it. It's iportant to move rail-fence-cipher ASAP, students are affected.

PR #647, opened 2018-01-22
rna-transcription: Add test case for invalid input

MazeChaZer:
The return type of `Maybe String` already anticipates error handling, but I've seen many solutions on exercism.io that would just throw an exception on invalid input. This test should point people in the right direction.

### DIFF
--- a/exercises/rna-transcription/package.yaml
+++ b/exercises/rna-transcription/package.yaml
@@ -1,5 +1,5 @@
 name: rna-transcription
-version: 1.2.0.6
+version: 1.2.0.7
 
 dependencies:
   - base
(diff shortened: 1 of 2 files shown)


petertseng:
In https://github.com/exercism/problem-specifications/pull/1027 three error cases were removed by the result of a coin flip. Now it is left up to the decision of individual tracks whether to add them back. That is really unfortunate because I really don't care to make such a decision, but I'll resolve the problem by simply saying if someone will do the work I'll merge it.

If the Haskell track is wanting to test invalid invalid inputs, I would suggest using those exact three case that were removed in https://github.com/exercism/problem-specifications/pull/1027 and the justification for keeping them can just be that the justification for removing them was not strong enough.

MazeChaZer:
I updated the file with the prior examples from the specification.

petertseng:
I'm going to use something like this

Revert "rna 1.1.0.5: remove Nothing tests"

This reverts commit 07c74263c85b35db3040d3356889498f1a4718ba.

In https://github.com/exercism/problem-specifications/pull/1027 three
error cases were removed by the result of a coin flip. Since that is
insufficient reason to remove them from the Haskell track, the Haskell
chooses to keep them, and needs no further justification to do so.

PR #651, opened 2018-02-06
Remove space from test string.

jwg4:
It might seem fair enough to have a space here. But the parsing is quite strict, for example expecting "" to parse to a matrix with 0 rows and 0 columns. So there really shouldn't be an extra space in a space-separated list of values.


jwg4:
@petertseng Thanks for fixing the version and the commit message! I will include these changes next time.

PR #652, opened 2018-02-07
Inline the exercise README insert

kytrinyx:
Now that we're operating with a track-wide template for generating exercise
READMEs we no longer need a separate markdown document to use for hints
and instructions that apply to all the exercises in a track. These instructions
can be defined directly in the template.

Ref: https://github.com/exercism/meta/issues/94


petertseng:
A reminder to maintainers: To override a README. place a `.meta/description.md`. Feel free to ask me for ideas on how to manage this with `.patch` files instead of maintaining full description.md files if the resulting file is only a few lines different from that in problem-specifications. I think about it from time to time but have not implemented anything yet as of this writing.

PR #659, opened 2018-02-28
Use Either instead of Maybe for collatz-conjecture

clementi:


petertseng:
Thanks for this PR. Will you explain the benefits of using Either instead of Maybe? Since there is no additional information gained from the Left (since all Left are the same here) I will want the help to understand the benefits.

clementi:
@petertseng I thought it better to express the case of `n <= 0` as producing an error, alerting the user that `n <=0` is not valid; that is, such values are outside of the domain of this function. It would be similar to throwing an `IllegalArgumentException` in Java. The Rust and Go versions of this exercise, which I recently completed, have `Result<u64, &'static str>` and `int, error` as return types, respectively, which clearly express that values of `n <= 0` are invalid inputs rather than inputs that just give you `Nothing`. But I will admit that this is mostly an aesthetic concern, and if anyone has major objections to it, I won't push the matter.

clementi:
I closed this because I deleted my fork.

petertseng:
I see, so the concern is how we express to the caller of the function how to rectify the mistake. Whereas I thought that:

* if there is only reason for the function to fail then it should use Maybe and the reason for failure can be noted in documentation of the function
* If there is more than one reason to fail it should use Either and a type that can be inspected by the caller (not a String) for information about what went wrong

However, I was unable to find any sources that back up my point of view, which makes me worry that my point of view is wrong. If anyone finds a source that says anything (right or wrong), please let me know so I can learn!

> I closed this because I deleted my fork.

Understood. You have reminded me that deleting the branch (`git push origin :branchname`) would also cause this to happen.

petertseng:
I tried to look for sources that might give a recommendation. I find some: http://blog.ezyang.com/2011/08/8-ways-to-report-errors-in-haskell-revisited/

> Maybe is a convenient, universal mechanism for reporting failure when there is only one possible failure mode and it is something that a user probably will want to handle in pure code.

It is a true statement that there is only one possible failure mode.

Also https://wiki.haskell.org/Handling_errors_in_Haskell

> It means you don't care why there isn't a value, or you already know.

Whether we "already know" is debatable, since this ties into questions of to what extent callers of a function are expected to, say, read its documentation. Or, if they are expected to intuit the reason why there is Nothing, and if so what is reasonable to expect someone to intuit.

I'll keep looking for any advice, including any contrary advice: any advice that says we should use Either even when there is only one cause of error. Anyone please let me know upon finding any so that we can learn.

PR #660, opened 2018-03-04
Implemented transpose exercise

tuxagon:
I implemented the transpose exercise for haskell.

### DIFF
--- a/exercises/transpose/examples/success-standard/src/Transpose.hs
+++ b/exercises/transpose/examples/success-standard/src/Transpose.hs
@@ -1,9 +1,7 @@
-{-# LANGUAGE LambdaCase #-}
-
 module Transpose (transpose) where
 
 import Data.List (dropWhileEnd)
-import Data.Maybe (isNothing)
+import Data.Maybe (fromMaybe, isNothing)
 
 transpose :: [String] -> [String]
 transpose [] = []
@@ -15,5 +13,4 @@ transpose rows =
   where
     longestLength = maximum $ map length rows
     padRight line = line ++ replicate (longestLength - length line) Nothing
-    toString line = map (\case Nothing -> ' '
-                               Just c -> c) $ dropWhileEnd isNothing line
+    toString line = map (fromMaybe ' ') $ dropWhileEnd isNothing line
--- a/exercises/transpose/examples/success-standard/src/Transpose.hs
+++ b/exercises/transpose/examples/success-standard/src/Transpose.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE LambdaCase #-}
+
 module Transpose (transpose) where
 
 import Data.Maybe (fromMaybe, isNothing)
@@ -12,8 +14,8 @@ transpose lines =
   where 
     longestLength = maximum $ map length lines
     padRight line = line ++ replicate (longestLength - length line) Nothing
-    toString line = concatMap (\x -> case x of Nothing -> " "
-                                               Just c -> c:"") $ dropRight line
+    toString line = concatMap (\case Nothing -> " "
+                                     Just c -> c:"") $ dropRight line
     dropRight line = snd $ foldr 
           (\x (ignore, xs) -> if ignore && isNothing x 
                                 then (ignore, xs) 
--- a/exercises/transpose/examples/success-standard/src/Transpose.hs
+++ b/exercises/transpose/examples/success-standard/src/Transpose.hs
@@ -1,0 +1,20 @@
+module Transpose (transpose) where
+
+import Data.Maybe (fromMaybe, isNothing)
+
+transpose :: [String] -> [String]
+transpose [] = []
+transpose lines =
+  map toString $ foldr 
+    (zipWith (:) . padRight . map Just) 
+    (replicate longestLength []) 
+    lines
+  where 
+    longestLength = maximum $ map length lines
+    padRight line = line ++ replicate (longestLength - length line) Nothing
+    toString line = concatMap (\x -> case x of Nothing -> " "
+                                               Just c -> c:"") $ dropRight line
+    dropRight line = snd $ foldr 
+          (\x (ignore, xs) -> if ignore && isNothing x 
+                                then (ignore, xs) 
+                                else (False, x:xs)) (True,[]) line
--- a/exercises/transpose/examples/success-standard/src/Transpose.hs
+++ b/exercises/transpose/examples/success-standard/src/Transpose.hs
@@ -8,11 +8,11 @@ import Data.Maybe (isNothing)
 transpose :: [String] -> [String]
 transpose [] = []
 transpose rows =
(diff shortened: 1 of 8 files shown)


exercises/transpose/examples/success-standard/src/Transpose.hs
```diff
@@ -0,0 +1,22 @@
...
+    rows
+  where 
+    longestLength = maximum $ map length rows
+    padRight line = line ++ replicate (longestLength - length line) Nothing
+    toString line = concatMap (\case Nothing -> " "
+                                     Just c -> c:"") $ dropRight line
+    dropRight line = snd $ foldr 
+          (\x (ignore, xs) -> if ignore && isNothing x 
+                                then (ignore, xs) 
+                                else (False, x:xs)) (True,[]) line
```
petertseng: In the `(True,[])` there is no space after the comma, but there is a space after all other commas in this file. I think for the purpose of consistency it would be good to have a comma after the space too.

exercises/transpose/package.yaml
```diff
@@ -0,0 +1,20 @@
+name: transpose
+version: 1.0.0.1
```
petertseng: Because https://github.com/exercism/problem-specifications/blob/master/exercises/transpose/canonical-data.json#L3 says 1.1.0, we should say 1.1.0.1 here.

exercises/transpose/test/Tests.hs
```diff
@@ -0,0 +1,181 @@
...
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
+import Transpose (transpose)
+
+main :: IO ()
+main = hspecWith defaultConfig {configFastFail = True} specs
+
+specs :: Spec
+specs = describe "responseFor" $ for_ cases test
```
petertseng: `describe` should describe what it's testing, but I think we are not testing `responseFor` here, so let's change this

exercises/transpose/examples/success-standard/src/Transpose.hs
```diff
@@ -0,0 +1,22 @@
...
+transpose [] = []
+transpose rows =
+  map toString $ foldr 
+    (zipWith (:) . padRight . map Just) 
+    (replicate longestLength []) 
+    rows
+  where 
+    longestLength = maximum $ map length rows
+    padRight line = line ++ replicate (longestLength - length line) Nothing
+    toString line = concatMap (\case Nothing -> " "
```
petertseng: It looks like the lambda always results in a single-character string), in which case I think `map` would make work just as well as `concatMap` and be clearer. Let me know if `map` can work, OR if I misunderstood the code and it's actually not possible.

petertseng:
EBWODP: 

> there are no trailing newlines in the submitted files?

I meant trailing **spaces** is what I don't want. Trailing newline at end of file should still be present.

exercises/transpose/examples/success-standard/src/Transpose.hs
```diff
@@ -0,0 +1,22 @@
...
+                                     Just c -> c:"") $ dropRight line
+    dropRight line = snd $ foldr
```
petertseng: correct me if I'm wrong - is this https://hackage.haskell.org/package/base-4.10.1.0/docs/Data-List.html#v:dropWhileEnd ? If so, I think that should be used, for clarity.

exercises/transpose/examples/success-standard/src/Transpose.hs
tuxagon: Good call!

exercises/transpose/package.yaml
tuxagon: Ah, nice find. I forgot to check that.

exercises/transpose/examples/success-standard/src/Transpose.hs
tuxagon: @petertseng this piece of the code confused me too when i was writing it because with just `map`, it ends up being `[[[Char]]]`. I used `concatMap` because I needed to use `concat` and `map` to remove that nested layer.

petertseng: Hold on, I don't think I understand this explanation and where `[[[Char]]]` comes from. Can you show me some code and explain what you mean and what compiler error occurs?

What I am saying is that these two pices of code are equivalent:

```haskell
xs = [1, 2, 3]
exprA = concatMap (\x -> [x + 1]) xs
exprB = map (\x -> x + 1) xs
```

I think the code as written in this PR is equivalent to exprA, and I think it should be exprB.

exercises/transpose/examples/success-standard/src/Transpose.hs
tuxagon: You are correct. I did not know about that function. Thanks for that!

exercises/transpose/examples/success-standard/src/Transpose.hs
tuxagon: Actually, due to `dropWhileEnd`, this entirely goes away.

exercises/transpose/examples/success-standard/src/Transpose.hs
```diff
@@ -0,0 +1,19 @@
...
+transpose [] = []
+transpose rows =
+  map toString $ foldr
+    (zipWith (:) . padRight . map Just)
+    (replicate longestLength [])
+    rows
+  where
+    longestLength = maximum $ map length rows
+    padRight line = line ++ replicate (longestLength - length line) Nothing
+    toString line = concatMap (\case Nothing -> " "
```
petertseng: As evidenced by the `Just c -> c:""` arm (since only a `Char` can be `:` to a `[Char]`), the lambda is `Maybe Char -> [Char]`, but each resulting `[Char]` has exactly one element (either a single space or the one character contained in the `Just`), so instead this should be a `map` using a `Maybe Char -> Char` lambda.

```haskell
xs = [1, 2, 3]
exprA = concatMap (\x -> [x + 1]) xs
exprB = map (\x -> x + 1) xs
```

tuxagon: Ah, I see. Mapping out the types is helpful in these cases.

tuxagon: Since this is the case, I can probably just use `fromMaybe` and get rid of the lambda entirely.

petertseng: Yes, `fromMaybe` would be good too!

petertseng:
Fantastic, thank you!

tuxagon:
Thanks for reviewing @petertseng! This was a fun exercise and the suggestions were helpful. The allowed-jaggedness actually added the most fun to this exercise, as well.

PR #663, opened 2018-03-08
Add poker

Average-user:
This is sort of a translation of the python implementation. In case that matters in some way.

### DIFF
--- a/exercises/poker/examples/success-standard/src/Poker.hs
+++ b/exercises/poker/examples/success-standard/src/Poker.hs
@@ -1,0 +1,51 @@
+module Poker (bestHands) where
+
+import Data.Maybe --(fromJust)
+import Data.List  --(isPrefixOf, nub, elemIndex, sortBy)
+
+validHand :: String -> Bool
+validHand h = and [ length (words h) == 5
+                  , all (`elem` "A23456789TJQK") (head <$> h')
+                  , all (`elem` "HSDC") (last <$> h')
+                  , all ((==2) . length) h']
+  where
+    h' = parseHand h
+
+parseHand :: String -> [String]
+parseHand = map f . words
+  where
+    f x | "10" `isPrefixOf` x = 'T' : [last x]
+        | otherwise           = x
+
+rankHand :: String -> (Int, [Int])
+rankHand hand | counts == [5]       = (9, ranks') 
+              | straight && flush   = (8, ranks') 
+              | counts == [4,1]     = (7, ranks') 
+              | counts == [3,2]     = (6, ranks') 
+              | flush               = (5, ranks') 
+              | straight            = (4, ranks') 
+              | counts == [3,1,1]   = (3, ranks') 
+              | counts == [2,2,1]   = (2, ranks') 
+              | counts == [2,1,1,1] = (1, ranks') 
+              | otherwise           = (0, ranks') 
+  where
+    h        = parseHand hand
+    r        = fromJust . flip elemIndex "..23456789TJQKA" . head <$> h
+    groups   = let x = nub r in sortBy (flip compare) (zip (times r <$> x) x)
+    counts   = fst <$> groups
+    ranks    = snd <$> groups
+    ranks'   = if ranks == [14,5,4,3,2] then [5,4,3,2,1] else ranks
+    straight = length counts == 5 && (maximum ranks' - minimum ranks') == 4
+    flush    = length (nub (last <$> words hand)) == 1
+    times xs x = length $ filter (==x) xs
+
+bestHands :: [String] -> Maybe [String]
+bestHands hands | not (all validHand hands) = Nothing
+                | otherwise                 = Just $ f (0,[]) [] hands
+  where
+    f _ r []                          = r
+    f m r (x:xs) | null r || rank > m = f rank [x] xs
+                 | rank == m          = f m (x : r) xs
+                 | otherwise          = f m r xs
+      where
+        rank = rankHand x
--- a/exercises/poker/examples/success-standard/src/Poker.hs
+++ b/exercises/poker/examples/success-standard/src/Poker.hs
@@ -3,13 +3,11 @@ module Poker (bestHands) where
 import Data.Maybe (fromJust)
 import Data.List  (nub, elemIndex, sortBy)
 
-validHand :: String -> Bool
-validHand h = and [ length (words h) == 5
-                  , all (`elem` "A23456789TJQK") (head <$> h')
-                  , all (`elem` "HSDC") (last <$> h')
-                  , all ((==2) . length) h']
-  where
-    h' = parseHand h
+validHand :: [String] -> Bool
+validHand h = and [ length h == 5
+                  , all (`elem` "A23456789TJQK") (head <$> h)
+                  , all (`elem` "HSDC") (last <$> h)
+                  , all ((==2) . length) h]
 
 parseHand :: String -> [String]
 parseHand = map f . words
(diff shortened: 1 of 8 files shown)


Average-user:
What difficulty should this have?

exercises/poker/examples/success-standard/src/Poker.hs
```diff
@@ -0,0 +1,51 @@
+module Poker (bestHands) where
+
+import Data.Maybe --(fromJust)
```
petertseng: I'd like to be able to keep track of where identifiers are coming from, so I'd say to uncomment the lists!

exercises/poker/package.yaml
```diff
@@ -0,0 +1,20 @@
+name: poker
+version: 1.1.1.0
```
petertseng: https://github.com/exercism/problem-specifications/blob/master/exercises/poker/canonical-data.json says 1.1.0 so thi should be 1.1.0.1

exercises/poker/test/Tests.hs
```diff
@@ -0,0 +1,170 @@
...
+                               , "3H 4H 5C 6C JD"]
+               , expected    = Just [ "3S 4S 5D 6H JH"
+                                    , "3H 4H 5C 6C JD"]
+               }
+        , Case { description =  "multiple hands with the same high cards, tie compares next highest ranked, down to last card"
+               , input       = [ "3S 5H 6S 8D 7H"
+                               , "2S 5D 6D 8C 7S"]
+               , expected    = Just ["3S 5H 6S 8D 7H"]
+               }          
+        , Case { description = "one pair beats high car"
```
petertseng: not high car

exercises/poker/examples/success-standard/src/Poker.hs
Average-user: Forgot about that

exercises/poker/package.yaml
Average-user: I placed the 1 on the other side, that was dumb

exercises/poker/test/Tests.hs
Average-user: should be this "multiple hands with the same high cards, tie compares next highest ranked, down to last card" right?

Average-user:
> Shrug. As long as it's understandable (it is).

It looks a little bit too imperative, but I don't know, maybe is just me being paranoid. To be honest I just used imperative/OO programming with python like for 1 month. So I don't really know anything about OOP, just FP and Prolog. If you think that something can be improved let me/us (if someone joins to this) about it.

exercises/poker/test/Tests.hs
petertseng: no. it should be "one pair beats high card"

Average-user: Ohh, apparently I'm blind

Average-user:
Weird. Hlint makes this suggestion
```Bash
exercises/poker/test/Tests.hs:19:21: Warning: Functor law
Found:
  sort <$> (bestHands input) `shouldBe` sort <$> expected
Why not:
  sort . ((bestHands input) `shouldBe` sort) <$> expected
```
but, when I do the changes, the program no longer works.

petertseng:
Maybe there is some reason why that would work, but it doesn't look like there is a way for that to work. You thus may consider filing an issue for hlint.

In the meantime, I have two potential alternatives for you to consider:

```hlint
(sort <$> bestHands input) `shouldBe` (sort <$> expected)
fmap sort (bestHands input) `shouldBe` fmap sort expected
```

From what I saw, HLint has no suggestions to make for either line.

exercises/poker/examples/success-standard/src/Poker.hs
```diff
@@ -0,0 +1,51 @@
...
+                  , all (`elem` "A23456789TJQK") (head <$> h')
+                  , all (`elem` "HSDC") (last <$> h')
+                  , all ((==2) . length) h']
+  where
+    h' = parseHand h
+
+parseHand :: String -> [String]
+parseHand = map f . words
+  where
+    f x | "10" `isPrefixOf` x = 'T' : [last x]
```
petertseng: I know there's no test for this currently, but **if** a test were added with a card "10XXXXXH" that would be an invalid card but this code would treat it as valid. In case such a test case were added in the future, then this could would have to be changed. I would like to avoid this by dealing with this now instead of later. Might I suggest something like?

```haskell
    f ('1':'0':x) = 'T' : x 
    f x           = x 
```

Or anything else similar that does the same.

Average-user: But thats, why I created the function ```validHand```
```Haskell
validHand :: String -> Bool
validHand h = and [ length (words h) == 5
                  , all (`elem` "A23456789TJQK") (head <$> h')
                  , all (`elem` "HSDC") (last <$> h')
                  , all ((==2) . length) h']
  where
    h' = parseHand h
```
Or maybe I'm not getting the point here.

petertseng: `parseHand` discards the XXXX portion of `10XXXXH`, doesn't it? So invalid `10XXXXH` will be treated as invalid.

petertseng: parseHand discards the `XXXX` portion of `10XXXXH`, doesn't it? So invalid `10XXXXH` will be incorrectly treated as valid, instead of invalid.

Average-user: Nope, because of ...
```Haskell
length (words h) == 5
````

Comment (petertseng) [reply exercises/poker/examples/success-standard/src/Poker.hs]: please fix the failing test.

Comment (petertseng) [reply exercises/poker/examples/success-standard/src/Poker.hs]: `length (words h) == 5` has nothing to do with `10XXXXH`. `10XXXXH` is ONE word.

Comment (Average-user) [reply exercises/poker/examples/success-standard/src/Poker.hs]: Does this suits your requirements?
 ```Haskell
validHand :: String -> Bool
validHand h = and [ length (words h) == 5
                  , all (`elem` "A23456789TJQK") (head <$> h')
                  , all (`elem` "HSDC") (last <$> h')
                  , all ((==2) . length) h'
                  , all f (words h)]
  where
    h'  = parseHand h
    f x = length x == 2 || (length x == 3 && take 2 x == "10")

parseHand :: String -> [String]
parseHand = map f . words
  where
    f ('1':'0':x) = 'T' : x 
    f x           = x 
```

Average-user: Well, It could be just one of the changes, don't know which one is better

petertseng: You're right that it could just be one of the two changes.

I think the change to `parseHand` (`f ('1':'0':x) = 'T' : x`) is easier to reason about, because it makes fewer assumptions. I can read `parseHand` and say "Okay, this will split the hand into its cards, and replace any 10 with T. Validity can be checked by only examining the output of `parseHand`". In fact, we can cement this if we say that `validHand` should be `[String] -> Bool` instead of `String -> Bool`.

If the change happens in `validHand` (to use `f x = length x == 2 || (length x == 3 && take 2 x == "10")`), then I have to look at both `parseHand` and `validHand` to verify that they together are doing the right thing. I have to say "Okay, `parseHand` split the hand into cards, and deletes any letters after 10, so that means `validHand` needs to look at both the input and output of `parseHand` to be able to do its work." I can't properly understand one without the other.

Average-user: let me think about it a while

Average-user:
@petertseng  [Here](https://github.com/ndmitchell/hlint/issues/454) Is the HLint issue in case you are interested.

Average-user:
Don't really know. Might be a little confusing  to call it earlier, but call it twice is not ideal either...

petertseng:
Let's talk about what made it confusing

Average-user:
```Haskell
module Poker (bestHands) where

import Data.Maybe (fromJust)
import Data.List  (nub, elemIndex, sortBy)

validHand :: [String] -> Bool
validHand h = and [ length h == 5
                  , all (`elem` "A23456789TJQK") (head <$> h)
                  , all (`elem` "HSDC") (last <$> h)
                  , all ((==2) . length) h]

parseHand :: String -> [String]
parseHand = map f . words
  where
    f ('1':'0':xs) = 'T' : xs
    f xs           = xs

rankHand :: [String] -> (Int, [Int])
rankHand h | counts == [5]       = (9, ranks')
           | straight && flush   = (8, ranks')
           | counts == [4,1]     = (7, ranks')
           | counts == [3,2]     = (6, ranks')
           | flush               = (5, ranks')
           | straight            = (4, ranks')
           | counts == [3,1,1]   = (3, ranks')
           | counts == [2,2,1]   = (2, ranks')
           | counts == [2,1,1,1] = (1, ranks')
           | otherwise           = (0, ranks')
  where
    r        = fromJust . flip elemIndex "..23456789TJQKA" . head <$> h
    groups   = let x = nub r in sortBy (flip compare) (zip (times r <$> x) x)
    counts   = fst <$> groups
    ranks    = snd <$> groups
    ranks'   = if ranks == [14,5,4,3,2] then [5,4,3,2,1] else ranks
    straight = length counts == 5 && (maximum ranks' - minimum ranks') == 4
    flush    = length (nub (last <$> h)) == 1
    times xs x = length $ filter (==x) xs

bestHands :: [String] -> Maybe [String]
bestHands hands | not (all (validHand . snd) hands') = Nothing
                | otherwise                          = Just $ f (0,[]) [] hands'
  where
    hands' = zip hands (parseHand <$> hands)
    f _ r []                              = r
    f m r ((x,y):xs) | null r || rank > m = f rank [x] xs
                     | rank == m          = f m (x : r) xs
                     | otherwise          = f m r xs
      where
        rank   = rankHand y
```
This is not that confusing after all

petertseng:
I see, so it looks better because `validHand` only needs to worry about the parsed version of the hand, which makes sense. But `bestHands` is still obligated to give back the unparsed version, so it needs to `zip hands (parseHand <$> hands)`. I would accept your proposal.

I might suggest that it might be best to zip *only* in the `Just` case. `hands'` can just be `parseHand <$> hands` and you can `zip hands hands'` when needed. All that does is allow `(validHand . snd)` to become just `validHand`

Average-user:
like this
```Haskell
bestHands :: [String] -> Maybe [String]
bestHands hands | not (all validHand hands') = Nothing
                | otherwise                  = Just $ f (0,[]) [] hands''
  where
    hands'  = parseHand <$> hands
    hands'' = zip hands hands'
    f _ r []                              = r
    f m r ((x,y):xs) | null r || rank > m = f rank [x] xs
                     | rank == m          = f m (x : r) xs
                     | otherwise          = f m r xs
      where
        rank   = rankHand y
```
?

petertseng:
Sure, though I would just write `zip hands hands'` inline, without `hands''`.

petertseng:
Cooool. Thanks!

PR #664, opened 2018-03-09
complex-numbers: add exp test cases

### LINKED ISSUE #616: complex-numbers: `exp` property from canonical data is missing
Since I had trouble to implement said function for my own track, I wanted to sneak peak here, but its not implemented here…

I finally got it fixed for erlang (in fact my tests were wrong, not the function ;)).

cmccandless:
Closes #616

### DIFF
--- a/exercises/complex-numbers/examples/success-standard/src/ComplexNumbers.hs
+++ b/exercises/complex-numbers/examples/success-standard/src/ComplexNumbers.hs
@@ -35,6 +35,9 @@ conjugate (Complex a b) = Complex a (-b)
 abs :: Floating a => Complex a -> a
 abs (Complex a b) = sqrt (exp2 a + exp2 b)
 
+exp :: Num a => Complex a -> Complex a
+exp (Complex a b) = Complex ((cos b) * (exp a)) ((sin b) * (exp a))
+
 -- binary operators ------------------------------------------------------------
 mul :: Num a => Complex a -> Complex a -> Complex a
 mul (Complex a b) (Complex c d) = Complex (a*c - b*d) (a*d + b*c)
--- a/exercises/complex-numbers/test/Tests.hs
+++ b/exercises/complex-numbers/test/Tests.hs
@@ -192,3 +192,23 @@ subCases =
             , expectedA    = (-2, -2)
             }
     ]
+
+expCases :: [CaseB]
+expCases = 
+    [ CaseB { descriptionB = "Euler's identity/formula"
+            , number1B     = (0, pi)
+            , expectedB    = (-1, 0)
+            }
+    , CaseB { descriptionB = "Exponential of 0"
+            , number1B     = (0, 0)
+            , expectedB    = (1, 0)
+            }
+    , CaseB { descriptionB = "Exponential of a purely real number"
+            , number1B     = (1, 0)
+            , expectedB    = (exp 1, 0)
+            }
+    , CaseB { descriptionB = "Exponential of a purely real number"
+            , number1B     = (log 2, pi)
+            , expectedB    = (-2, 0)
+            }
+    ]
--- a/exercises/complex-numbers/src/ComplexNumbers.hs
+++ b/exercises/complex-numbers/src/ComplexNumbers.hs
@@ -31,6 +31,9 @@ real = error "You need to implement this function"
 imaginary :: Num a => Complex a -> a
 imaginary = error "You need to implement this function"
 
+exp :: Num a => Complex a -> Complex a
+exp = error "You need to implement this function"
+
 -- binary operators ------------------------------------------------------------
 mul :: Num a => Complex a -> Complex a -> Complex a
 mul = error "You need to implement this function"


exercises/complex-numbers/examples/success-standard/src/ComplexNumbers.hs
```diff
@@ -35,6 +36,9 @@ conjugate (Complex a b) = Complex a (-b)
 abs :: Floating a => Complex a -> a
 abs (Complex a b) = sqrt (exp2 a + exp2 b)
 
+expComplex :: Num a => Complex a -> Complex a
```
petertseng: just as `abs` and `div` use their respective `Prelude` names, I think `exp` should as well, instead of getting a `Complex` suffix. Take a look at the `import Prelude hiding (div, abs)` and adjust accordingly. Or, explain why `exp` should have a different treatment than `abs` and `div`.

cmccandless: I agree. However, I'm not experienced enough with Haskell to know how to hide Prelude.exp and still use it (needed for e = `exp 1`). If someone else knows how to do this, I'd love to know and would be happy to remove the suffix.

Average-user:
```Haskell
import Prelude hiding (whatEverFunctionYouWantToHide)
```

petertseng: > and still use it

Good point. Try `Prelude.exp`, which should still work even if `hiding` `exp`.

petertseng: I think that should work, ` gchi` session which demonstrates my belief

```
$ ghci                                                           (03-10 10:10)
GHCi, version 8.2.2: http://www.haskell.org/ghc/  :? for help
Prelude> import Prelude hiding (exp)
Prelude> exp

<interactive>:2:1: error: Variable not in scope: exp
Prelude> Prelude.exp 1
2.718281828459045
```

cmccandless:
Apologies on the late update on this... The notification somehow fell between the cracks.

I've applied the requested changes. Assuming CI passes, this PR should be ready for merge; I recommend squashing.

exercises/complex-numbers/test/Tests.hs
```diff
@@ -12,14 +12,15 @@ main = hspecWith defaultConfig {configFastFail = True} specs
...
 specs = do
-          describe "real"      $ for_ realCases      $ testC real
-          describe "imaginary" $ for_ imaginaryCases $ testC imaginary
-          describe "conjugate" $ for_ conjugateCases $ testB conjugate
-          describe "abs"       $ for_ absCases       $ testC abs
-          describe "mul"       $ for_ mulCases       $ testA mul
-          describe "div"       $ for_ divCases       $ testA div
-          describe "add"       $ for_ addCases       $ testA add
-          describe "sub"       $ for_ subCases       $ testA sub
+          describe "real"       $ for_ realCases       $ testC real
```
petertseng: since it's no longer necessary to add an extra space to all these lines, I'd say to remove them

petertseng:
I may need to do more investigation on that `Prelude.exp` thing. I was convinced it would work, but maybe not.

exercises/complex-numbers/test/Tests.hs
cmccandless: Done.

petertseng:
I wonder if (as suggested by https://wiki.haskell.org/Import ) this'll need an additional `import qualified Prelude as P` and using `P.exp`, but I haven't investigated whether that'll do.

I admit I didn't test the `Prelude.exp` thing which is why I seem to have given a bad suggestion

petertseng:
The `Prelude as P` thing is good. I need to take some time to think about what to do about the `Floating a` thing.

```
    • Could not deduce (Floating a) arising from a use of ‘cos’
      from the context: Num a
        bound by the type signature for:
                   exp :: forall a. Num a => Complex a -> Complex a
        at src/ComplexNumbers.hs:40:1-38
      Possible fix:
        add (Floating a) to the context of
          the type signature for:
            exp :: forall a. Num a => Complex a -> Complex a
    • In the first argument of ‘(*)’, namely ‘cos b’
      In the first argument of ‘Complex’, namely ‘(cos b * P.exp a)’
      In the expression: Complex (cos b * P.exp a) (sin b * P.exp a)
   |
41 | exp (Complex a b) = Complex (cos b * P.exp a) (sin b * P.exp a)
   |                              ^^^^^
```

petertseng:
Well, that exposes floating point error. Actually the division tests are susceptible to floating point error too. That's unfortunate.

cmccandless:
The [python track](https://github.com/exercism/python/blob/master/exercises/complex-numbers/example.py#L42-L45) example solution uses rounding to avoid floating point error. Would that be acceptable here?

petertseng:
Something like that. I was thinking it may have to use rounding to check only for approximate equality rather than exact equality. Something like what was done in https://github.com/exercism/haskell/blob/master/exercises/space-age/test/Tests.hs#L27-L28 . I do not yet have a concrete proposal, but that is my general idea.

cmccandless:
@petertseng @sshine 
I have merged upstream/master into this PR and restored a change that was lost in the merge. The tests all pass, are there any other changes that need to be made?

cmccandless:
If this is ready to merge, I highly suggest squash-merge.

sshine:
@petertseng: It says that you are requesting changes, but has this request expired?

petertseng:
Oh yeah I think that's just for the tests passing. They certainly pass now so time to dismiss

PR #675, opened 2018-04-18
sieve : Hint that trial division might be used

bartavelle:
This hopefully should make it clear when trial division is used instead of the proper algorithm.

For further readings, [this](https://www.cs.hmc.edu/~oneill/papers/Sieve-JFP.pdf) paper :)

### DIFF
--- a/exercises/sieve/examples/success-standard/src/Sieve.hs
+++ b/exercises/sieve/examples/success-standard/src/Sieve.hs
@@ -4,6 +4,8 @@ import Control.Monad (when)
 import qualified Data.Vector as V
 import qualified Data.Vector.Mutable as MV
 
+import Prelude hiding (div, mod, divMod, rem, quotRem, quot, (/))
+
 primesUpTo :: Int -> [Int]
 primesUpTo upperBound = V.ifoldr f [] $ V.create $ do
   mv <- MV.replicate (1 + upperBound) (1 :: Int)
--- a/exercises/sieve/src/Sieve.hs
+++ b/exercises/sieve/src/Sieve.hs
@@ -1,4 +1,8 @@
 module Sieve (primesUpTo) where
 
+-- You should not use any of the division operations when implementing
+-- the sieve of Eratosthenes.
+import Prelude hiding (div, mod, divMod, rem, quotRem, quot, (/))
+
 primesUpTo :: Integer -> [Integer]
 primesUpTo n = error "You need to implement this function."


petertseng:
As you can see by looking at the Travis CI result, this is going to need something like https://github.com/ndmitchell/hlint#ignoring-hints . One may ask "is it wise to have this HLint annotation in something we give to the student" but I think it is helpful in case the student also chooses to run HLint. So it makes sense to add the annotation. Ask if you need help figuring out what it should look like (I don't know off the top of my head either).

bartavelle:
I added an annotation ... isn't it a bit scary?

petertseng:
Are you perhaps thinking that a student will be unsure what the annotation is about? Or something else? Let's talk about it. For example, is it possible to alleviate any confusion by adding a comment above the annotation explaining what it is for?

If instead the better solution is to not have the annotation in the file served to the student, then we can brainstorm ways to achieve that.

bartavelle:
Well, this is a problem indeed. It seems you can only have a single `.travis.yaml` file, which would be global. The in-file annotation would be the less intrusive solution ..

petertseng:
Yeah that's interesting.

There'd be a disadvantage of having the `.hlint.yaml` in repo root, because whoever maintains this repo may forget to look there, so makes the behaviour harder to understand than if there were an `.hlint.yaml` in the exercise directory.

Putting it in the file will make it available to the student, which can be convenient if the student wishes to run `hlint`. Otherwise, they would get the same suggestion as well!

If Travis CI https://github.com/exercism/haskell/blob/master/.travis.yml#L57 were to change to invoke hlint on each exercise directory separately instead of on the entire repository at once, there would be a possibility to use per-exercise `.hlint.yaml`. The potential disadvantage would be if `hlint` takes significant time to start up. Would have to experiment a little to see.

bartavelle:
Where do you see this going? Do you want to experiment with travis?

petertseng:
I would say that having a global `.hlint.yaml` would be too surprising; I would forget to look there. So I wouldn't want that.

Between the other two potions being considered, I have no particular preference between a per-exercise `.hlint.yaml` versus putting the annotation in the file directly. I'm happy to leave it to your judgment which of those two you would like. But as may be apparent from my not having managed to do so in the past 4 days, it seems that I do not have the time to experiment with Travis.

bartavelle:
I will try to experiment with travis next month.

petertseng:
:+1:

petertseng:
Thanks for this! I'm glad to see it finally get merged.

bartavelle:
And I did not have to do anything more! I did not forget about this ... at all ;)

PR #678, opened 2018-04-19
sieve : Hint that trial division might be used [2]

bartavelle:
This hopefully should make it clear when trial division is used instead
of the proper algorithm.

For further readings,
[this](https://www.cs.hmc.edu/~oneill/papers/Sieve-JFP.pdf) paper :)

This is a followup to #675 , as I do not know how to edit this PR to add a file.

### DIFF
--- a/exercises/sieve/src/.hlint.yaml
+++ b/exercises/sieve/src/.hlint.yaml
@@ -1,0 +1,1 @@
+- ignore: {name: Unnecessary hiding}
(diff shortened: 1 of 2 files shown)


petertseng:
You should instead push the commit to the same branch.

How did I arrive at that conclusion? At https://github.com/exercism/haskell/pull/675, at the bottom there are instructions that say "Add more commits by pushing to the [patch-2](https://github.com/bartavelle/haskell/tree/patch-2) branch on [bartavelle/haskell](https://github.com/bartavelle/haskell)." Therefore, that is how you would update the existing PR. You can either push a new commit to the branch, OR amend and force push, either is fine.

I need to ask that you update the original PR instead of opening a new one. The reason I must ask this is so that all the discussion on the proposed change is in the same place. It will be too difficult to follow the discussion if it is spread across two places.

bartavelle:
This seemed to work.

PR #685, opened 2018-06-08
Reformat config files

kytrinyx:
This runs the configlet fmt command, which normalizes the
contents and ordering of keys in the track config and
maintainers config.

This will let us script changes to the config files without
having unnecessary noise in the diffs when submitting pull
requests.

See exercism/meta#95

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,9 +1,9 @@
 {
   "language": "Haskell",
   "active": true,
+  "blurb": "",
   "ignore_pattern": "example",
   "solution_pattern": "example.*[.]hs",
-  "foregone": [],
   "exercises": [
     {
       "slug": "hello-world",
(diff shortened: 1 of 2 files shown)


config.json
```diff
@@ -852,27 +852,47 @@
     {
       "slug": "binary",
       "uuid": "48ca1682-ca1b-4a2c-be55-624620f336a7",
+      "core": false,
```
petertseng: that's unfortunate.

kytrinyx: Yes, so sorry I missed this in the previous round of fixes.

kytrinyx:
I'm going to go ahead and merge this so we can move forward, but will come through and fix the issue with deprecated exercises as soon as I can get configlet fixed.

PR #689, opened 2018-06-15
ABOUT: Reflow sentences for one sentence per line

jackhughesweb:
I've reformatted the about section and removed the code sample for the new formatting guidelines in v2 [issue #686]

### DIFF
--- a/docs/ABOUT.md
+++ b/docs/ABOUT.md
@@ -1,24 +1,16 @@
-
-Haskell is a purely functional, lazy, statically-typed programming
-language with type inference.
+Haskell is a purely functional, lazy, statically-typed programming language with type inference.
 
 **Functional** means that functions are first-class data types.
 
 **Purely Functional** means (roughly) that there are no side effects.
-Every function will always return the same value for a given argument
-will do nothing else.
+Every function will always return the same value for a given argument will do nothing else.
 
-**Lazy** (a.k.a "non-strict") means that the compiler will put off
-evaluating a thing until absolutely neccessary.  This lets you safely
-do weird stuff like operating on an infinite list--the language will
-only create it up to the last value you actually use.
+**Lazy** (a.k.a "non-strict") means that the compiler will put off evaluating a thing until absolutely neccessary.
+This lets you safely do weird stuff like operating on an infinite list--the language will only create it up to the last value you actually use.
 
-**Statically-typed** means that identifiers have a type set at compile
-time--like those in Java, C++ or C#--instead of holding data of any
-type like those in Python, Ruby or JavaScript.
+**Statically-typed** means that identifiers have a type set at compile time--like those in Java, C++ or C#--instead of holding data of any type like those in Python, Ruby or JavaScript.
 
-**Type inference** means that the compiler will often figure out the
-type of an identifier by itself so you don't have to specify it.
+**Type inference** means that the compiler will often figure out the type of an identifier by itself so you don't have to specify it.
 Scala and later versions of C# both do this.
 
 Haskell syntax is beautiful and minimal.  For example, here is an
@@ -34,8 +26,5 @@ quicksort (x:xs) =
 ```
 [(source)](http://learnyouahaskell.com/recursion#quick-sort)
 
-In addition, Haskell is standardized and has multiple high-quality
-implementations, some of which produce standalone native binaries.
-There is also a collection of free third-party libraries available and
-a package manager ("cabal") to automatically fetch them for you.
-
+In addition, Haskell is standardized and has multiple high-quality implementations, some of which produce standalone native binaries.
+There is also a collection of free third-party libraries available and a package manager ("cabal") to automatically fetch them for you.


docs/ABOUT.md
```diff
@@ -1,41 +1,18 @@
-
-Haskell is a purely functional, lazy, statically-typed programming
-language with type inference.
+Haskell is a purely functional, lazy, statically-typed programming language with type inference. 
+It has a beautiful and minmal syntax.
```
petertseng: Note for reviewers: This sentence was added.

jackhughesweb: I've added that as it was part of the sentence next to the code section (which has been removed as part of the new formatting guidelines). Feedback is welcome on the changes 😄

petertseng: Previously, *immediately* after the list of traits describing Haskell (purely functional, lazy, etc.), there comes a description of each trait. 

Now, there is this sentence in between. Does this interposition interfere with the reading and understanding of this text?

Alternative: Leaving this sentence about beautiful and minimal syntax where it originally was. Is that better or worse, without having a code example?

(I have not decided what my own answer to these two questions is yet, so I'll leave them for others to think about for now)

petertseng: Applicable if this sentence is chosen to remain:

spelling: minmal -> minimal

petertseng: Applicable if this sentence is chosen to remain:

spelling: minmal -> minimal

jackhughesweb: I've updated the file to fix the typo

docs/ABOUT.md
```diff
@@ -1,41 +1,18 @@
-
-Haskell is a purely functional, lazy, statically-typed programming
-language with type inference.
+Haskell is a purely functional, lazy, statically-typed programming language with type inference. 
+It has a beautiful and minimal syntax.
```
petertseng: I am copying this old comment so that it remains visible on the changed line. It is unchanged from before, so no additional response is necessary! It is only a reminder!

(I have not myself given additional thought to this issue, but if nobody else has any ideas I'm going to suggest the sentence move back to where it was before so it doesn't come in between the list of adjectives and the definitions of each adjective)

---



Previously, immediately after the list of traits describing Haskell (purely functional, lazy, etc.), there comes a description of each trait.

Now, there is this sentence in between. Does this interposition interfere with the reading and understanding of this text?

Alternative: Leaving this sentence about beautiful and minimal syntax where it originally was. Is that better or worse, without having a code example?

(I have not decided what my own answer to these two questions is yet, so I'll leave them for others to think about for now)

jackhughesweb: Any thoughts from the other @exercism/haskell @exercism/haskell-contributors members?

docs/ABOUT.md
```diff
@@ -1,41 +1,18 @@
...
-**Type inference** means that the compiler will often figure out the
-type of an identifier by itself so you don't have to specify it.
+**Type inference** means that the compiler will often figure out the type of an identifier by itself so you don't have to specify it.
 Scala and later versions of C# both do this.
 
-Haskell syntax is beautiful and minimal.  For example, here is an
-implemention of quicksort in 6 lines:
-
-```haskell
-quicksort :: (Ord a) => [a] -> [a]
```
petertseng: Should this become the https://github.com/exercism/haskell/blob/master/docs/SNIPPET.txt ? Is the code within the size requirements?

jackhughesweb: It's too wide for the box (42 characters max per line)

docs/ABOUT.md
petertseng: Looks like none of them are interested in responding yet. In that case, I will give my advice, with appropriate time for anyone to respond to this advice. Stand by.

petertseng:
Nobody responded to my questions. In the interest of forward progress, I'm inclined to ignore my previous concerns about whether the sentence reads well with the interposition of the "beautiful and minimal syntax' bit. It's probably fine.

Let me instead unfortunately bring forth new ones.

I would treat this PR as a necessity if the website were broken without this PR. However, my untrained eye does not see any obvious defects in https://exercism.io/tracks/haskell . In particular:

* the code block is formatted with monospace text as it should be (though one thing we do miss is syntax highlighting, but we don't even have that for SNIPPET)
* the sentences seem to be line-broken in reasonable places, rather than in the middle of the sentence

So based on this understanding, that means I see that the two components of this PR have the following effects:

* Sentence reflow: Cosmetic change, doesn't matter whether we have it
* Removing the code snippet: Removes value from this page; we lose an example of code.

Given these effects, I can't in good conscience merge this.

For it to make sense, I'd need to understand any negative effects of not merging this PR. The guidelines exist. It certainly is the case that they exist for a reason; I was just unaware of the reason and need to be educated. Or, my untrained eye failed to spot some defects that exist currently, in which case I just need the proper training to know what to look for.

My proposed plan for this PR; these things can happen in *either order* and don't necessarily have to be done by the same person:

* Determine the reason why each sentence should be on a single line.
  * If so, one PR with only the sentence reflow (leave out the code deletion), and it can be merged forthwith
  * If not, skip this portion and do not do it
* Determine the reason why the code should be removed
  * If so, one PR with only the removal. Replace the code with a reasonable code in SNIPPET.txt; the SNIPPET.txt does not currently contain something interesting enough (it is hello world)
  * If not, skip this portion and do not do it

NobbZ:
> Determine the reason why each sentence should be on a single line.

Basically because its less conflict-prone and produces easier to read diffs.

> Determine the reason why the code should be removed

Basically because there is now a code example determined from the SNIPPET. Also the ABOUT.md should be a short excerpt of the languages features, not a showcase of its beatifulness ;)

PR #690, opened 2018-06-15
Added blurb

jackhughesweb:
I've added a blurb to `config.json` for issue #686

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,7 +1,7 @@
 {
   "language": "Haskell",
   "active": true,
-  "blurb": "",
+  "blurb": "Haskell is a functional programming language which is pure and statically-typed. It's known for lazy evaluation, where evaluation is deferred until necessary, and its purity, where monads are used for working with side-effects.",
   "ignore_pattern": "example",
   "solution_pattern": "example.*[.]hs",
   "exercises": [


config.json
```diff
@@ -1,7 +1,7 @@
 {
   "language": "Haskell",
   "active": true,
-  "blurb": "",
+  "blurb": "Haskell is a functional programming language, which is pure and statically-typed. It's known for lazy evaluation, where evaluation is deferred until necessary, and it has support for list comprehensions.",
```
petertseng: > and it has support for list comprehensions.

Can you convince reviewers that this particular feature is important enough to mention here? I do indeed think that everything else (pure, static typing, lazy evaluation) defines the language such that it is important to mention here, but list comprehensions doesn't seem to belong in the same class as it is just one feature rather than defining the language.

To get an idea of how people summarise Haskell, I have now been looking at https://en.wikipedia.org/wiki/Haskell_(programming_language) and https://www.haskell.org/ . I guess I see that list comprehensions are listed in https://en.wikipedia.org/wiki/Haskell_(programming_language)#Features , but not in the introduction. 

So I guess what I am trying to say here is:

This proposed blurb has a good number of things that affect the whole language. It additionally has one individual feature, and the fact that there is only one makes it seem out of place. I think zero or three would be better numbers of features to list.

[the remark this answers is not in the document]
jackhughesweb: I've removed the comma for readability

jackhughesweb: I've removed the part about list comprehensions, and added a part about purity/monads (using zero or three features would make the blurb too long/short in comparison to blurbs for other tracks). Let me know what you think/any changes or suggestions you have

config.json
```diff
@@ -1,7 +1,7 @@
 {
   "language": "Haskell",
   "active": true,
-  "blurb": "",
+  "blurb": "Haskell is a functional programming language which is pure and statically-typed. It's known for lazy evaluation, where evaluation is deferred until necessary, and it's purity, where monads are used for working with side-effects.",
```
petertseng: Seems good. `it's purity` should be `its purity`, right?

I think with that change, that is the only change I have remaining to ask to make.

jackhughesweb: I have corrected this mistake

PR #691, opened 2018-06-15
Updated exercise data in config.json

jackhughesweb:
- Added core exercises
- Added missing topics
- Added unlocked_by for non-core
- Set four exercises as non-core but unlocked by default (bonus exercises)
- Added `auto_approve` for the first exercise (hello-world)

[Issue #686]

PR #692, opened 2018-06-19
Add auto_approve to hello-world exercise

jackhughesweb:
This adds the auto_approve property to the hello-world 
exercise as part of preparing tracks for the v2 launch.

The current code on the server will automatically approve
hello-world exercises anyway, but that is a temporary fix
and will be removed in the future.

See exercism/haskell#686

### DIFF
--- a/config.json
+++ b/config.json
@@ -9,6 +9,7 @@
       "slug": "hello-world",
       "uuid": "c97d9dda-c7ee-4595-b447-693559d03ba5",
       "core": false,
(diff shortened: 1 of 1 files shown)


jackhughesweb:
This is already being worked on in #691 - closing this PR

petertseng:
I think this deserves to be its own commit, not mashed together with all other changes. Therefore I'm going to reopen it.

PR #697, opened 2018-07-31
stack.yaml: lts-11.19 -> lts-12.4

petertseng:
https://www.stackage.org/diff/lts-11.19/lts-12.4

Last update 4 months ago
https://github.com/exercism/haskell/pull/668

I think it would be interesting to get on GHC 8.4.x, so LTS 12.x is the
ticket.


petertseng:
Turns out, a `Multiset a` might as well just be a `Map a Int`.

PR #698, opened 2018-07-31
various exercises: Use problem-specifications versions

petertseng:
I think there's about 3.5 months' worth of changes to catch up on


petertseng:
holding on isbn-verifier 2.7.0, current solution fails it

petertseng:
same w/ forth 1.6.0

PR #700, opened 2018-08-01
forth 1.6.0.8: add tests for word redefinition

petertseng:
Most submitted implementations did not implement these behaviours:

* Words using other previously-defined words
* Words using the then-current version of those previously-defined words
* Recursive definition (word using previous version of itself)

These are now added to check these behaviours.

These were confirmed to match multiple Forth implementations:

* https://repl.it/languages/forth
* https://tio.run/#forth-gforth

https://github.com/exercism/problem-specifications/pull/1243


exercises/forth/test/Tests.hs
```diff
@@ -119,6 +119,17 @@ specs = do
         runTexts [ ": + * ;"
                  , "3 4 +"   ] `shouldBe` Right [12]
 
+      it "can use different words with the same name" $
+        runTexts [ ": foo 5 ;"
+                 , ": bar foo ;"
+                 , ": foo 6 ;"
+                 , "bar foo"     ] `shouldBe` Right [5, 6]
+
+      it "can use different words with the same name" $
```
petertseng: wrong!

https://github.com/exercism/haskell/pull/718

PR #706, opened 2018-08-12
TESTS: clarify running tests in the exercise directory; Add instructions for test with file watch

bennypowers:


petertseng:
Thanks for the tips!

PR #711, opened 2018-08-28
difference-of-squares: Rename squareOfSums to squareOfSum

Scientifica96:
In the exercise "difference-of-squares", I find the name `squareOfSums` weird, as it returns the square of a single sum. Maybe the name `squareOfSum` is more appropriate.


petertseng:
Very good, thank you for your first contribution!

Scientifica96:
Thank you for your valuable remarks. Gonna take them into consideration.

You're welcome, glad I helped :).

PR #713, opened 2018-09-06
Bob's conversational partner is a purist!

sshine:
In the [Bob exercise](https://github.com/exercism/haskell/tree/master/exercises/bob) it says:

> Bob's conversational partner is a purist when it comes to written
> communication and always follows normal rules regarding sentence
> punctuation in English.

But in the unit test the following input is fed to `responseFor`:

    "\nDoes this cryogenic chamber make me look fat?\nno"

This is a rhetorical question, and by the expected value, this is deemed
to not be an actual question, which is very reasonable. But if Bob's
conversational partner is a punctuation purist, it should at least be
"no." with the period.

I found this by wanting to grab the last character that `isPunctuation`,
but had to go with `not . isSpace` (like the reference solution). But
this really sounds wrong when I think about it: I actually want the last
punctuation character and am allowed to assume "normal rules".

Making this change will not render `not . isSpace` solutions wrong.

### DIFF
--- a/exercises/bob/examples/success-standard/src/Bob.hs
+++ b/exercises/bob/examples/success-standard/src/Bob.hs
@@ -1,5 +1,6 @@
 module Bob (responseFor) where
-import Data.Char (isSpace, isUpper, isAlpha)
+import Data.Char (isSpace, isUpper, isAlpha, isPunctuation)
+import Safe (lastMay)
 
 data Prompt = Silence | YellQuestion | Yell | Question | Other
 
@@ -10,7 +11,7 @@ classify s | all isSpace s = Silence
            | question = Question
            | otherwise = Other
            where yell = any isAlpha s && all isUpper (filter isAlpha s)
-                 question = '?' == last (filter (not . isSpace) s)
+                 question = Just '?' == lastMay (filter isPunctuation s)
 
 response :: Prompt -> String
 response Silence = "Fine. Be that way!"
--- a/exercises/bob/package.yaml
+++ b/exercises/bob/package.yaml
@@ -1,5 +1,5 @@
 name: bob
-version: 1.2.0.5
+version: 1.4.0.7
 
 dependencies:
   - base
--- a/exercises/bob/examples/success-standard/package.yaml
+++ b/exercises/bob/examples/success-standard/package.yaml
@@ -2,6 +2,7 @@ name: bob
 
 dependencies:
   - base
+  - safe
 
 library:
   exposed-modules: Bob
--- a/exercises/bob/test/Tests.hs
+++ b/exercises/bob/test/Tests.hs
@@ -101,7 +101,7 @@ cases = [ Case { description = "stating something"
                , expected    = "Fine. Be that way!"
                }
         , Case { description = "multiple line question"
-               , input       = "\nDoes this cryogenic chamber make me look fat?\nno"
+               , input       = "\nDoes this cryogenic chamber make me look fat?\nNo."
                , expected    = "Whatever."
                }
         , Case { description = "starting with whitespace"


exercises/bob/package.yaml
```diff
@@ -3,6 +3,7 @@ version: 1.2.0.5
 
 dependencies:
   - base
+  - safe
```
petertseng: `safe` should not be included in the list of dependencies distributed to students, since not all student solutions will use `safe`

exercises/bob/examples/success-standard/src/Bob.hs
```diff
@@ -1,5 +1,6 @@
 module Bob (responseFor) where
-import Data.Char (isSpace, isUpper, isAlpha)
+import Data.Char (isSpace, isUpper, isAlpha, isPunctuation)
+import Safe
```
petertseng: don't forget to add `safe` to ` exercises/bob/examples/success-standard/package.yaml`.

petertseng: since only `lastMay` is being used from `Safe`, I'd suggest making that clear by listing it. Otherwise it may not be clear when/if it is safe to remove `import Safe` if code ever changes in the future.

petertseng:
Indeed, you will want to remove that commit from this PR, because each PR should only be focused on a single task.

You can do so by removing it from the branch that this PR is based off of, which is the branch listed after "from" in the portion of the page that reads "merge 4 commits into `exercism:master` from `sshine:master`"

Don't hesitate to ask if you need help figuring out how to do so.

sshine:
I removed the rogue commit and squashed the remaining ones related to this issue.

(That sounded violent.)

sshine:
I'll go with option 2 tomorrow.

petertseng:
> I think if you merge #717 and then this one, versions should add up.

That is true, though neither Git nor GitHub are smart enough to understand what to do with the two commits changing the same line. Nevertheless, easy enough to tell it what we want, and it was to be expected anyway.

(As in the last PR, don't panic about my adding a merge commit in, repository requirements, it will be squashed)

petertseng:
Finally, you had a question.

> Will the tests for all language tracks be rebuilt based on this, or will the language track maintainers be notified of the change somehow?

To date, there has not been a good system for this, neither automatic rebuilds nor automatic notifications. The relevant project is https://github.com/exercism/meta/issues/99

Until it is completed, language track maintainers are kind of on their own. Some might use scripts similar to https://github.com/petertseng/exercism-problem-specifications/tree/up-to-date/up-to-date to compare their track's versions vs that of problem-specifications. There may be other solutions to this problem floating around out there.

In certain cases, a concerned individual decides the change is so important that it warrants opening an issue in all tracks. Certainly nothing preventing such a concerned individual from doing so, and there is even https://github.com/exercism/blazon that assists in such a matter. Whether it actually gets done is usually decided case-by-case depending on how concerned the individual is.

sshine:
> So there are three choices to proceed:
>
> * Combine the test changes into this PR, having one PR move `bob` from 1.2.0 to 1.4.0
> * Put 1.2.0 to 1.3.0 in a separate PR and this PR only deals with 1.3.0 to 1.4.0
> * Do nothing and make me do the 1.2.0 to 1.3.0 change, and this PR will only deal with 1.3.0 to 1.4.0. This is an option because I wouldn't want to hold up 1.4.0 and make you do more work just because this track is not up to date.
>
> I have no particular preference between these three options right now.

@petertseng: Do you have a preference between these options today?

In that case, I'd like to clarify the formulation in #852.

petertseng:
When I consider my preference today, I keep in mind that my preference may affect how a contributor is able to contribute to this repository.

Therefore: no preference.

A contributor may decide to only apply the earlier of two version changes, and leave the later one to a different contributor.
Or a contributor may decide to apply all of them in one PR.
Or a contributor may decide to apply all of them, but in one PR each.

Note that I didn't mention what should happen when a contributor only wants to apply the later of two version changes but leave the earlier one for a different contributor. But that raises difficulties on what, if anything, the version should be changed to. But I don't really expect anyone to want to do that.

In all cases, the relevant problem-specifications PRs must be linked. But that is already a stipulation of https://github.com/exercism/haskell/issues/852.

PR #715, opened 2018-09-10
acronym: Add "ROTFLSHTMDCOALM" test case from shared metadata.

sshine:


petertseng:
Do not panic that I accidentally changed `nucleotide-count`'s version instead of `acronym`; I have rectified that mistake

petertseng:
I hope I haven't done so in too unfriendly of a manner previously (for example, if I was too terse and therefore seemed snippy)! If I have, I apologise.

petertseng:
(It was equally likely that it referred to me as a maintainer of any other open source project, so I preferred to overapologise for something I didn't do rather than fail to apologise for something I did)

Thanks for this work!

PR #722, opened 2018-09-12
phone-number 1.5.0.6: add cases to test data that cover invalid use of "1" or "0"

sshine:
This should catch solutions with a validation function that calls itself recursively when the first digit is '1'.

Addresses exercism/problem-specifications#1325

### DIFF
--- a/exercises/phone-number/package.yaml
+++ b/exercises/phone-number/package.yaml
@@ -1,5 +1,5 @@
 name: phone-number
-version: 1.4.0.5
+version: 1.5.0.6
 
 dependencies:
   - base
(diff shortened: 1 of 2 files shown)


sshine:
I'll adopt the new cases.

sshine:
To go to version 1.6.0 of problem-specifications for phone-number seems to require us to use `Either String` instead of `Maybe`: https://github.com/exercism/problem-specifications/commit/7290b84b809ccdacfa7ab674b1cb1237ca55105f -- I'm not sure where the decision for this should rest.

NobbZ:
@sshine Why? How a specific error object from the spec is represented in the actual implementations is totally up to the track.

sshine:
I think we should keep to `Maybe` and think of a good gradient point for using `Either` as error handling.

I agree with your points.

Issue #695, opened 2018-07-24
Clock exercise promotes bad behavior


isovector:
The clock exercise is really bad. The only meaningful semantics of a clock are of an absolute point in time, which aren't closed under addition. Thus `Num` is an exceedingly bad instance to give for them, and the exercise itself suggests ignoring most of the methods.

A much better solution is to have a `addDelta :: Hour -> Minute -> Clock -> Clock`

petertseng:
While whoever wrote this exercise probably thought it was handy that you can use the `+` and `-` operators, I believe @isovector is correct here; it doesn't actually make sense to use them. The operation `08:20 + 09:01` (where both operands denote a time) is not sensible (the answer is NOT 17:21) because logically I can't add two times together. I can add *nine hours and one minute* to *eight twenty*, but I can't add *one past nine* to *eight twenty*.

I can add a duration to a time and get another time.
I can subtract a duration from a time and get another time.
I can subtract two times and get a duration. (but this is not covered in this exercise)
The typeclass `Num` does not express these desired semantics.
As we can see in http://hackage.haskell.org/package/base-4.11.1.0/docs/Prelude.html#t:Num , `Num` implies that you can do `a + a` and get an `a`, and as discussed above that is not sensible for times.
Therefore, as it currently stands, the `clock` exercise is not well-typed.

The proposed `addDelta :: Hour -> Minute -> Clock -> Clock` is a reasonable proposal that deals with the problems.

Assuming I haven't missed any files, these are the files that should be changed to resolve this issue:

* https://github.com/exercism/haskell/blob/master/exercises/clock/.meta/hints.md (remove mention of `Num`)
* https://github.com/exercism/haskell/blob/master/exercises/clock/README.md (regenerate using https://github.com/exercism/configlet or any compatible tool after editing hints.md, OR just make the exact same change as made to hints.md)
* https://github.com/exercism/haskell/blob/master/exercises/clock/examples/success-standard/src/Clock.hs
* https://github.com/exercism/haskell/blob/master/exercises/clock/src/Clock.hs
* https://github.com/exercism/haskell/blob/master/exercises/clock/test/Tests.hs

tiniuclx:
I think the proposed changes make sense, I'll see what I can do about this. I'll have a go at solving the exercise as-is, then I'll write up a pull request for this issue.

PR #723, opened 2018-09-20
Do not use a Num instance in Clock

tiniuclx:
Addresses #695 .


exercises/clock/examples/success-standard/src/Clock.hs
```diff
@@ -1,18 +1,16 @@
-module Clock (fromHourMin, clockHour, clockMin, toString) where
+module Clock (clockHour, clockMin, fromHourMin, toString, addDelta) where
```
petertseng: I think I see that the benefit of moving fromHourMin was to keep this in alphabetical order. In that case, `addDelta` should be the first element, should it not?

exercises/clock/examples/success-standard/src/Clock.hs
```diff
@@ -21,4 +19,4 @@ clockMin :: Clock -> Int
 clockMin = (`mod` 60) . unClock
 
 toString :: Clock -> String
-toString c = printf "%02d:%02d" (clockHour c) (clockMin c)
+toString c = printf "%02d:%02d" (clockHour c) (clockMin c)
```
petertseng: I think the whitespace change, being unrelated to removing Num, should be amended away from this PR

(also I think the trailing newline should stay, but whether I think that is not really relevant to this PR)

exercises/clock/src/Clock.hs
```diff
@@ -1,4 +1,4 @@
-module Clock (clockHour, clockMin, fromHourMin, toString) where
+module Clock (clockHour, clockMin, fromHourMin, toString, addDelta) where
```
petertseng: for alphabetical order, I think addDelta should go first. Unless there's no actual reason to use a particular ordering

exercises/clock/test/Tests.hs
```diff
@@ -4,33 +4,19 @@ import Data.Foldable     (for_)
 import Test.Hspec        (Spec, describe, it, shouldBe)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
 
-import Clock (fromHourMin, toString)
+import Clock (fromHourMin, toString, addDelta)
```
petertseng: probably `addDelta` first, for alphabetical order

petertseng:
Do not panic about the merge commit just added, administrative requirement of the repo that will be squashed away anyway.

petertseng:
I changed my mind and decided there is a reasonable sequence of commits such that travis CI passes at each step and the changes are logically separate at each step.

tiniuclx:
Thank you very much for merging this! I shall pay more attention to such details if I end up contributing to this track in the future.

petertseng:
Thanks for doing that! Another thing I can suggest is to pay attention to the list of words in https://help.github.com/articles/closing-issues-using-keywords/ that will auto-close an issue (if a PR will do that)

PR #724, opened 2018-09-24
Create Series-Alt.hs

jrp2014:
An alternative solution.  Offers a couple of alternative windowing functions and a slightly simpler / more intuitive way of multiplying through the Eithers.


petertseng:
Your intention is to add another example solution, is that right? If so, you'll want to add it under another directory in https://github.com/exercism/haskell/tree/master/exercises/largest-series-product/examples. For an example of an exercise with multiple examples, see https://github.com/exercism/haskell/tree/master/exercises/anagram/examples .

I will evaluate whether this example solution should be added to this repository later; you should give a reason why so. 

I really think historically we only add alternate solution so when we want to make sure that an implementation that uses different type signature still works (see the anagram examples how one uses `[String]` and another uses `SetText`). Because of the time taken in Travis CI, it may not make sense to add alternative implementations for reasons that are not related to checking alternate type signatures. 

If it is applicable to your goals, you may consider replacing the current solution in the track with your proposed solution, or adding the solution to https://github.com/exercism/mentors/tree/master/tracks/haskell/exercises if it is relevant to mentors who may wish to comment on the Series solutions

jrp2014:
Thanks, Peter. Yes, it’s an alternative. As I said in the comments, I think that it is simpler (more idiomatic) avoiding the need for Maybes, and fewer exotic functions. Let me know whether it is worth adding / replacing the existing model solution. Incidentally, the window function used in this example is similar to another exercise, the name of which I can’t immediately recall. Best, John

petertseng:
Reviewers who want to compare this against the current example solution: You may find the link https://github.com/exercism/haskell/blob/master/exercises/largest-series-product/examples/success-standard/src/Series.hs convenient. If you already have an alternate preferred way of viewing the existing example set up, please excuse my interruption.

jrp2014:
Thanks, Peter. Another variant of this problem that would be even more instructive would allow the use of foldable / traversable. But that would probably need a different api as the Either doesn’t have instances of those classes by default, and I suspect that if you write them there would be no advantage to using them over Maybe. Best, John

jrp2014:
... and here is an even simpler alternative answer:

```
module Series
  ( Error(..)
  , largestProduct
  )
where

import           Data.Char
import           Data.List                      ( tails )

data Error
  = InvalidSpan
  | InvalidDigit Char
  deriving (Show, Eq)

largestProduct :: Int -> String -> Either Error Integer
largestProduct size digits = if size > length digits || size < 0
  then Left InvalidSpan
  else do
    digits' <- traverse parseDigits digits
    return $ maximum $ product <$> window' size digits'

-- Two alternative sliding window functions
window :: Int -> [a] -> [[a]]
window size xs | size <= 0        = [[]]
               | length xs < size = []
               | otherwise        = take size xs : window size (tail xs)

window' :: Int -> [a] -> [[a]]
window' size xs = take (length xs - (size - 1)) $ map (take size) (tails xs)

parseDigits :: Char -> Either Error Integer
parseDigits c = if isDigit c
  then Right $ fromIntegral $ digitToInt c
  else Left $ InvalidDigit c
  ```

**Comment by petertseng** [comment]
Since nobody else reviewed, now it's my turn.

Even though `window'` is shorter, I contend it's a bit harder to reason through the fact that `length xs - (size - 1)` is the correct number of things to take. I think `window` should be kept. Unless there actually are concerns of efficiency with `window`, but I don't know enough to say.

I propose that this proposed code should replace the existing solution in success-standard/src/Series.hs - no Series-alt.hs.

The reasoning behind replacing the existing solution is that I'm happy to reduce dependencies such as that on `safe`. Please remember to remove the dependency https://github.com/exercism/haskell/blob/master/exercises/largest-series-product/examples/success-standard/package.yaml#L10 in that case.

The reasoning behind not keeping multiple solutions is that the purpose of the examples in track repositories isn't to serve as a collection of various alternative solutions for students to consider (I may have to conveniently ignore the fact that https://github.com/exercism/haskell/blob/master/exercises/accumulate/examples/success-standard/src/Accumulate.hs#L11-L13 exists, for this discussion); their foremost purpose is to prove that there exists a solution that can pass the tests (and therefore that the tests are not completely unsolvable). With build times as high as they currently are I think it only makes sense to maintain multiple solutions for an exercise when we need to make sure multiple type signatures work with the tests.

Finally, please remember to make the solution pass CI (as you see it does not right now).

**Comment by sshine** [comment]
@jrp2014: This is pretty close to being merged.

Would you like to do this?

**Comment by jrp2014** [comment]
Thanks.  I suggest that you just take the Oct 2 version above and replace the current version, if you have write permission.

**Comment by sshine** [comment]
@jrp2014: I would rather have that the product of your efforts is linked to the documentation within this pull request. I can only make that happen by opening another pull request and linking to it, or by sending a pull request to your feature branch, `patch-1`.

If you could `--amend` your changes to `patch-1` and `push -f` them, I'll gladly rebase and merge.

**Comment by jrp2014** [comment]
OK.  I'll have a go.  But before I do, here is another, more terse solution
```
module Series (Error(..), largestProduct) where

import Control.Monad    ((>=>))
import Data.Char        (digitToInt, isDigit)
import Data.List        (tails)
import Data.Maybe       (mapMaybe)
import Safe             (maximumMay)
import Safe.Exact       (takeExactMay)

data Error = InvalidSpan | InvalidDigit Char deriving (Show, Eq)

rightOr :: a -> Maybe b -> Either a b
rightOr a Nothing = Left a
rightOr _ (Just x) = Right x

largestProduct :: (Integral a, Num b, Ord b) => a -> String -> Either Error b
largestProduct n = traverse charToNum >=> rightOr InvalidSpan . maximumMay . products
  where
    products = mapMaybe (fmap product . takeExactMay (fromIntegral n)) . tails
    charToNum x
        | isDigit x = Right . fromIntegral . digitToInt $ x
        | otherwise = Left (InvalidDigit x)
```

**Comment by jrp2014** [comment]
Well I think that I managed to include the Oct 2 code, but I cannot interpret the failure messages, which don't seem to be related to my change.

**Comment by petertseng** [comment]
Don't forget to remove the line https://github.com/exercism/haskell/blob/master/exercises/largest-series-product/examples/success-standard/package.yaml#L10 , since example doesn't depend on Safe anymore.

We wouldn't want to forget to take the step that actually lets us reap the benefits of this PR.

**Comment by jrp2014** [comment]
OK.  Done that.  I have not changed the version number in the main package.yaml for the exercise.

**Comment by petertseng** [comment]
> OK. Done that. 

As you can see, you also need to delete `dependencies:` since there are now no dependencies (https://github.com/exercism/haskell/issues/672)

> I have not changed the version number in the main package.yaml for the exercise.

yup, it should not be changed. Nothing that is delivered to the student is changing in this PR

File: exercises/largest-series-product/examples/success-standard/src/Series.hs [hunk 1]
```diff
@@ -1,22 +1,34 @@
...
+  deriving (Show, Eq)
 
-rightOr :: a -> Maybe b -> Either a b
-rightOr a Nothing = Left a
-rightOr _ (Just x) = Right x
+largestProduct :: Int -> String -> Either Error Integer
+largestProduct size digits = if size > length digits || size < 0
+  then Left InvalidSpan
+  else do
+    digits' <- traverse parseDigits digits
```
sshine: Excellent idea to handle errors monadically. I might have used `mapM` instead of `traverse` now that we're in a do-block, but since they're equivalent, I don't mind. I suppose I like the notion of `mapM` because of the `...M` suffix.

If, however, you put this in a separate function, e.g. `digits :: String -> Either Error [Int]`, you can compose it more easily.

exercises/largest-series-product/examples/success-standard/src/Series.hs
```diff
@@ -1,22 +1,34 @@
...
+    return $ maximum $ product <$> window size digits'
```
sshine: Use `.` instead of `$` here: If you need to chain multiple `$`s, what you're effectively doing is composition.

exercises/largest-series-product/examples/success-standard/src/Series.hs
```diff
@@ -1,22 +1,34 @@
...
+               | length xs < size = []
+               | otherwise        = take size xs : window size (tail xs)
+
+-- window' :: Int -> [a] -> [[a]]
+-- window' size xs = take (length xs - (size - 1)) $ map (take size) (tails xs)
+
+parseDigits :: Char -> Either Error Integer
+parseDigits c = if isDigit c
+  then Right $ fromIntegral $ digitToInt c
+  else Left $ InvalidDigit c
```
sshine: Use guards and `.` here, too.

I'd call it `toDigit` and leave the error handling implied by the type.

sshine:
I would have liked to put the purple "Merged" icon on your PR rather than my own, but unfortunately I can't squash, amend and rebase your branch, nor make updates to this PR, nor push directly to master at exercism/haskell, which is probably a good thing.

jrp2014:
Thanks.

PR #726, opened 2018-09-25
Add topics to config.json

sshine:
Perhaps I don't understand the policy of [TOPICS.txt](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt); some of the topics there would not make sense to add to Haskell exercises; e.g. recursion, pattern matching and polymorphism would appear excessively. But perhaps on the first exercise that aims to teach those it would make sense.

With that in mind I've added two topics.

### DIFF
--- a/config.json
+++ b/config.json
@@ -22,7 +22,9 @@
       "core": true,
       "unlocked_by": null,
       "difficulty": 1,
-      "topics": []
+      "topics": [
+        "booleans"
+      ]
     },
     {
       "slug": "space-age",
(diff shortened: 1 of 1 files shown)


sshine:
If anyone cares to say that this is the right direction, I can continue to add topics to exercises.

My intent is to migrate this labelling to the Ocaml and SML tracks when appropriate.

petertseng:
Excuse my having to rebase. Needed that for the administrative (branch must be up-to-date) requirement of this repo. The commits made sense to keep separate and I'm not willing to use *Rebase and merge* functionality of GitHub so I have to rebase.

PR #727, opened 2018-09-26
Changing robot-name to use StateT

zhenengxie:
Altered the type signatures in the exercise, the testing suite and the example solution to use the StateT monad transformer.

### DIFF
--- a/exercises/robot-name/examples/success-standard/src/Robot.hs
+++ b/exercises/robot-name/examples/success-standard/src/Robot.hs
@@ -1,22 +1,44 @@
-module Robot (Robot, mkRobot, robotName, resetName) where
+module Robot (Robot, initialState, mkRobot, resetName, robotName) where
 
-import Control.Concurrent (MVar, readMVar, swapMVar, newMVar)
-import Control.Monad (void)
-import System.Random (randomRIO)
+import           Control.Concurrent.MVar (MVar, newMVar, putMVar, readMVar,
+                                          takeMVar)
+import           Control.Monad.State     (StateT, get, modify)
+import           Control.Monad.Trans     (lift)
+import           Data.Set                (Set)
+import qualified Data.Set                as S
+import           System.Random           (randomRIO)
 
 newtype Robot = Robot { robotNameVar :: MVar String }
+type RunState = Set String
 
-mkRobot :: IO Robot
-mkRobot = fmap Robot $ generateName >>= newMVar
+initialState :: RunState
+initialState = S.empty
 
-resetName :: Robot -> IO ()
-resetName r = void (generateName >>= swapMVar (robotNameVar r))
+randomName :: IO String
+randomName = mapM randomRIO [letter, letter, digit, digit, digit]
+  where
+    letter = ('A', 'Z')
+    digit = ('0', '9')
+
+randomUniqueName :: StateT RunState IO String
+randomUniqueName = do
+  name <- lift randomName
+  nameSpace <- get
+  if name `S.member` nameSpace
+    then randomUniqueName
+    else do
+      modify (S.insert name)
+      return name
+
+mkRobot :: StateT RunState IO Robot
+mkRobot = fmap Robot $ randomUniqueName >>= lift . newMVar
+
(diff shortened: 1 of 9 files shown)


exercises/robot-name/examples/success-standard/src/Robot.hs
```diff
@@ -1,22 +1,43 @@
...
+    else return name
+
+mkRobot :: StateT RunState IO Robot
+mkRobot = fmap Robot $ randomUniqueName >>= lift . newMVar
+
+resetName :: Robot -> StateT RunState IO ()
+resetName (Robot name) = do
+  oldName <- lift $ takeMVar name
+  newName <- randomUniqueName
+  modify (S.delete oldName . S.insert newName)
```
ajkavanagh: The specs for the exercise seem to be misleading.  I thought that every name had to be unique, even if they were reset (i.e. never delete the old names).  This is obviously a different interpretation.  I don't know if there is a "correct" interpretation?

exercises/robot-name/examples/success-standard/src/Robot.hs
```diff
@@ -1,22 +1,43 @@
...
+    letter = ('A', 'Z')
+    digit = ('0', '9')
+
+randomUniqueName :: StateT RunState IO String
+randomUniqueName = do
+  name <- lift randomName
+  nameSpace <- get
+  if name `S.member` nameSpace
+    then randomUniqueName
+    else return name
```
ajkavanagh: I may be missing something here, and it's quite likely(!), but were does the new name get stored back into the set from initialState?

zhenengxie:
Thanks for pointing that out! I've fixed the code to add the generated name to the namespace whenever it's generated. Also figured out how the READMEs are generated, the Travis checks now pass.

ajkavanagh:
This is good.  I think there might need to be some more explanation in the exercise overview about _why_ the StateT needs to be used, and perhaps a pointer to where the student can learn about it.  Also, this is stacking monads, which is a non-trivial exercise; perhaps a pointer to the chapter in the free [Real World Haskell](https://wiki.haskell.org/Real_World_Haskell) book?

zhenengxie:
Thanks for the feedback. I've added links in the README to articles on the State monad and the article on monad transformers you recommended. I haven't included any explanation in the hints though since this seems to be the style of hints in other exercises; they bring up concepts and trust the user to learn about them on their own rather than explain the concepts.

ajkavanagh:
Yes, that looks really nice now.  I can't land it, that'll be @petertseng :)

zhenengxie:
I think I've addressed most of your concerns now. The new testing for collisions generates 5000 robots which means if the solution generates robot names uniformly at random without collision testing then there is almost certainly going to be a collision. A current problem is that when the test fails all 5000 items get printed to screen by hspec.

exercises/robot-name/examples/success-standard/src/Robot.hs
```diff
@@ -1,22 +1,44 @@
-module Robot (Robot, mkRobot, robotName, resetName) where
+module Robot (Robot, mkRobot, resetName, robotName, initialState) where
 
-import Control.Concurrent (MVar, readMVar, swapMVar, newMVar)
-import Control.Monad (void)
-import System.Random (randomRIO)
+import           Control.Concurrent.MVar (MVar, newMVar, putMVar, readMVar,
```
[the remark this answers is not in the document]
zhenengxie: My emacs config auto formats like this when saving the solution.

petertseng:
> when the test fails all 5000 items get printed to screen by hspec.

that... is undeed unfortunate. Of the following options, how would you like to proceed, given that this problem exists?

1. Advise that this be merged even with this problem, and hope that students do not run into it often. In such a case, I highly encouraging making an issue about this problem if you want it to be fixed someday.
2. Attempt to fix this problem before this be merged. You might consider using `expectationFailure` such as how https://github.com/exercism/haskell/blob/master/exercises/beer-song/test/Tests.hs#L28-L30 has it.

zhenengxie:
Thanks for the tip. I've now fixed the error message to only list the repeated names.

PR #734, opened 2018-10-09
allergies: Use Bounded in example.

sshine:


petertseng:
Let's take this but not change the version, since students don't perceive any change

sshine:
Okay.

PR #735, opened 2018-10-09
allergies: Property-based tests.

sshine:
I was adding the `allergies` exercise to the Ocaml track and wanted to write some property-based tests for it, since that seemed quicker.

I found Ocaml's QCheck library hard to understand, so I decided to write them in Haskell and port them to Ocaml. I don't know if they're really needed here, since the coverage for this exercise is already quite fine. But here they are.

### DIFF
--- a/exercises/allergies/test/Tests.hs
+++ b/exercises/allergies/test/Tests.hs
@@ -1,7 +1,10 @@
 {-# OPTIONS_GHC -fno-warn-type-defaults #-}
 
+import Test.QuickCheck   (Gen, forAll, forAllShrink, elements, sublistOf, suchThat)
 import Test.Hspec        (Spec, describe, it, shouldBe)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+import Data.List         (delete, lookup)
+import Data.Maybe        (mapMaybe)
 
 import Allergies
   ( Allergen ( Cats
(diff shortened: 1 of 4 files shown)


sshine:
From the failed build:

    /home/travis/build/exercism/haskell/build/allergies/stub/test/Tests.hs:29:47: error:
        • No instance for (Show Allergen) arising from a use of ‘forAll’
        • In the expression: forAll allergenWithScore

sshine:
Travis CI passes for this one now.

sshine:
Indeed.

For the Ocaml tests, I did generate negative samples.

sshine:
Negative properties and shrinking added.

sshine:
Negative properties and shrinking added.

sshine:
Negative properties and shrinking added.

exercises/allergies/test/Tests.hs
```diff
@@ -98,3 +129,81 @@ specs = do
...
+
+            it "accepts single allergens" $ forAll allergenWithScore $
+              \(allergen, score) -> allergies score == [allergen]
+
+            -- Property: For an arbitrary `score` and arbitrary `allergen`
+            -- that does not match it, that `allergen` is not an element of
+            -- `allergies score`.
+
+            it "rejects mismatching allergens" $ forAll complementWithScore $
+              \(allergen, score) -> allergen `notElem` allergies score
```
petertseng: I think this particular test (`notElem`) is not necessary, because testing for exact list equality in `allergies` already precludes this possibility. So I would think either "accepts single allergens" needs to change to using `elem` (and this test remains), or this test should be removed

exercises/allergies/test/Tests.hs
```diff
@@ -98,3 +129,81 @@ specs = do
...
+              forAllShrink allergensWithScore shrinkAllergensWithScore $
+                \(allergens, score) -> allergies score == allergens
+
+            -- Property: For an arbitrary `score` and all `allergens` that
+            -- are not part of that score, there is no overlap between
+            -- `allergens` and `allergies score`.
+
+            it "rejects multiple mismatching allergens" $
+              forAllShrink complementsWithScore shrinkComplementsWithScore $
+                \(allergens, score) -> null (allergens `intersect` allergies score)
```
petertseng: for similar reasons as the `notElem` test, I think this test is not necessary, since exact list equality of `allergies` already means no unnecessary allergens can be in the result of `allergies`. So the above "accepts multiple allergens" would have to also use an `intersect`, or this test should be removed

exercises/allergies/test/Tests.hs
sshine: I was quite tired when I read this, so I've waited until now to address this.

I understand the difference and agree that this test is redundant.

exercises/allergies/test/Tests.hs
sshine: I agree to removing it.

sshine:
It doesn't get a lot. I think there has been one since Jeremy reset the non-core queue.

Also, it gets unlocked after `Difference Of Squares` (the next-last core exercise) as the third icon, so it's kind of hidden away.

PR #744, opened 2018-10-26
Added a section on GHC installation.

jeeger:
Shortly described how GHC installation works with stack, and how to
check that GHC is installed correctly. Also added a "cautionary" note
about trying to run the Exercism exercises with a non-stack GHC (i.e.
might work, but stack is easiest).

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -15,7 +15,7 @@ installed correctly.
 ## Installing GHC
 
 Stack will automatically download and install GHC (the Glasgow Haskell
-compiler) when it is required. You can start the installation test it
+compiler) when it is required. You can start the installation and test
 by running the following command in a terminal:
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/INSTALLATION.md
```diff
@@ -12,6 +12,24 @@ stack --version
 If it outputs a Stack version equal to or greater than 1.1.2, then you have it
 installed correctly.
 
+## Installing GHC
+
+Stack will automatically download and install GHC (the Glasgow Haskell
+compiler) when it is required. You can start the installation test it
```
petertseng: in "...start the installation test it by running..." I think the sentence is more grammatically correct if "it" is removed

jeeger:
Thanks for the feedback!

docs/INSTALLATION.md
jeeger: Yes, that's an error ☺

PR #750, opened 2018-10-29
Removed mention of `Show` from README.

### LINKED ISSUE #747: Clock exercise doesn't test `show`
The Clock exercise description calls for implementing an instance of `Show`, but only a function `toString` is tested. Should the description or test be updated?

jeeger:
Since it's not necessary for the exercise or for testing, the `Show`
instance has been removed from the README.

Fixes #747.

### DIFF
--- a/exercises/clock/.meta/hints.md
+++ b/exercises/clock/.meta/hints.md
@@ -1,7 +1,7 @@
 ## Hints
 
 To complete this exercise you need to define the data type `Clock`,
-with `Eq` and `Show` instances, and implement the functions:
+add an `Eq` instance, and implement the functions:
 
 - addDelta
 - fromHourMin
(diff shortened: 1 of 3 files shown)


petertseng:
One small favour to ask of this PR:

https://github.com/exercism/haskell/blob/4a6bf1435a53596f2398eb69cce703085c229849/exercises/clock/examples/success-standard/src/Clock.hs#L4

It will help give more credibility if `Show` is removed from that line as well.

The example is not shown to the students, but is used in Travis CI, and removing `Show` from the example will prove beyond a doubt that `Show` is not necessary.

I have faith that the tests will still pass with that change. And in that case, I will merge it.

jeeger:
Thanks for your great vetting! Done!

petertseng:
Thanks!

PR #760, opened 2018-11-07
bob 1.4.0.8: Make test suite compatible with `Data.Text`

sshine:
The only thing that is necessary for the test suite to support a

    responseFor :: Text -> Text

is to assume that it has the type `IsString s => s -> s`.

We leave the stub as `String -> String`, but make the test suite compatible with the change. This means that mentors can suggest a solution based on `Data.Text` if they like. One case where that might make sense is when they refer to `Data.Text.strip`.

A `Data.Text`-based solution is provided as an example.

Exercise version bumped from 1.4.0.7 to 1.4.0.8.

### DIFF
--- a/exercises/bob/examples/success-text/package.yaml
+++ b/exercises/bob/examples/success-text/package.yaml
@@ -1,15 +1,12 @@
 name: bob
-version: 1.4.0.8
 
 dependencies:
   - base
+  - text
 
 library:
   exposed-modules: Bob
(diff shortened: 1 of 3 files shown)


sshine:
Sure. There aren't any other text-heavy core exercises, but as an extension to #761, we could have that.

Do you think we should switch entirely to `Data.Text` in `Bob`?

PR #703, opened 2018-08-02
Add 'mathematics' topic to relevant exercises in track config

kytrinyx:
The new site lets people filter optional exercises by topic, and this
makes it easy for people to skip past them.

This is important since the math-y exercises are not key to learning
the language, and a lot of people find them intimidating and demotivating.

I'm opening this PR to help stimulate discussion--please discuss in exercism/exercism.io#4110

PR #763, opened 2018-11-08
Add 'math' to 'grains' and 'triangle'

sshine:
All exercises that have the topic 'number_theory' should have the topic 'math', but not vice versa.

The topic 'number_theory' was created by @Average-user in #609.

It does not conform to [TOPICS.txt](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt), but I don't know if the policy is that it ought to.


sshine:
Related: #703. I realize that #4110 decides that it's non-math but that this was debated.

Still, how can it be number-theoretic and not mathy?

petertseng:
I would be inclined to say not - at least if we take from https://github.com/exercism/exercism/issues/4110 . Grains was explicitly called out as an exercise that should not have `math` applied. I don't suppose I can say the same for triangle though.

But, to stay in keeping with the principle, perhaps that means `number_theory` should be removed from them?

sshine:
Are you saying triangles aren't mathy?

petertseng:
I don't think "triangles aren't mathy" is a reasonable stance to take.

I wonder if it was missed in the examination.

If the meaning behind the `math` label is not "math" but instead "math but without a motivating story behind it", that would be unfortunate since clearly that nuance doesn't preserve if config.json simply writes "math".

What of grains?

sshine:
And the fact that it has a non-math theme (a prince and a chess board) should still warn people of there being math in it, since, I suppose, topics aren't based on categorizing themes. Otherwise, it should probably have `chess` in it, too?

PR #768, opened 2018-11-15
Add sshine to maintainers.conf

sshine:


petertseng:
yup.

The wording of your question leads to a follow-up: were you thinking of an alternative course of action when rebasing is necessary?

PR #773, opened 2018-11-16
Add -Wall to all package.yaml

sshine:
I added `ghc-options: -Wall` below `source-dirs: src` 

    library:
      ...
      source-dirs: src
      ghc-options: -Wall

for both example solutions and exercise stubs, using the following command:

    find . -name package.yaml -exec perl -i -pe 's/(source-dirs: src)/$1\n  ghc-options: -Wall/s' {} \;

This causes warnings to show on both `stack build` and `stack test`.

I realize that version numbers should be bumped, too. Oh dear.

### DIFF
--- a/exercises/binary/package.yaml
+++ b/exercises/binary/package.yaml
@@ -1,5 +1,5 @@
 name: binary
-version: 0.9.0.2
+version: 0.9.0.3
 
 dependencies:
   - base
@@ -7,6 +7,7 @@ dependencies:
 library:
   exposed-modules: Binary
   source-dirs: src
+  ghc-options: -Wall
   # dependencies:
   # - foo       # List here the packages you
   # - bar       # want to use in your solution.
--- a/exercises/binary/package.yaml
+++ b/exercises/binary/package.yaml
@@ -1,5 +1,5 @@
 name: binary
-version: 0.9.0.2 # 2016-07-26
+version: 0.9.0.2
 
 dependencies:
   - base
--- a/exercises/zipper/package.yaml
+++ b/exercises/zipper/package.yaml
@@ -1,5 +1,5 @@
 name: zipper
-version: 1.1.0.4
+version: 1.1.0.5
 
 dependencies:
   - base
@@ -7,6 +7,7 @@ dependencies:
 library:
   exposed-modules: Zipper
   source-dirs: src
+  ghc-options: -Wall
   # dependencies:
   # - foo       # List here the packages you
   # - bar       # want to use in your solution.
--- a/exercises/accumulate/package.yaml
+++ b/exercises/accumulate/package.yaml
@@ -1,5 +1,5 @@
 name: accumulate
-version: 0.1.0.2
+version: 0.1.0.3
 
 dependencies:
   - base
@@ -7,6 +7,7 @@ dependencies:
 library:
   exposed-modules: Accumulate
   source-dirs: src
+  ghc-options: -Wall
   # dependencies:
   # - foo       # List here the packages you
   # - bar       # want to use in your solution.
--- a/exercises/acronym/package.yaml
+++ b/exercises/acronym/package.yaml
@@ -1,5 +1,5 @@
 name: acronym
-version: 1.6.0.8
+version: 1.6.0.9
 
 dependencies:
   - base
@@ -7,6 +7,7 @@ dependencies:
 library:
   exposed-modules: Acronym
   source-dirs: src
+  ghc-options: -Wall
   # dependencies:
   # - foo       # List here the packages you
   # - bar       # want to use in your solution.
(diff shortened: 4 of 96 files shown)


sshine:
Version numbers bumped:

    for p in $(git diff-tree --no-commit-id --name-only -r HEAD); do
        perl -i -pe 's/(?<=version: )(.*?)(\d+)$/"$1".($2+1)/e' $p
    done

exercises/binary/package.yaml
```diff
@@ -1,12 +1,13 @@
 name: binary
-version: 0.9.0.2 # 2016-07-26
+version: 0.9.0.3
```
petertseng: careful. probably leave the comment as-is?

(as for why it's 0.9, because nobody ever bothered to update it to 1.x because it was deprecated. the script that shows outdated versions just skips it)

exercises/zipper/package.yaml
```diff
@@ -1,12 +1,13 @@
 name: zipper
-version: 1.1.0.4
+version: 1.1.0.6
```
petertseng: you somehow increased this version by two... which I guess is not a completely terrible thing to do, but which I suppose we should avoid if we can.

exercises/binary/package.yaml
sshine: This comment was added in a85c3dcb883ddccf6463576d061691c529a4075c in 2017-05-06 and suggests that

> Test cases adapted from `exercism/x-common/binary.json` on 2016-07-26

We keep this kind of date information as git metadata.

Since the exercise is deprecated, we no longer update the canonical test data.

Other deprecated exercises don't have comments in the `version` line.

Keeping this comment makes automated version retrieval and bumping harder.

exercises/zipper/package.yaml
sshine: Oops!

exercises/binary/package.yaml
petertseng: Back when this track was following https://github.com/exercism/haskell/issues/522 , that comment would have been required.

The good news is that in choosing not to copy that particular clause into the README, I implicitly decided that 0.9.0.x version numbers are now undefined behaviour, and anything goes. We may do whatever we like with this.

So I leave you this to consider:

The commit message has claimed that a version bump has been performed. I do not perceive "remove a comment" to be within the bounds of the operation "bump version number". I recommend this be remedied by either adding in the commit message a note that the comment was removed, or putting the comment removal in its own commit.

This is actually not so important that I would actually be displeased in any way if things were left as-is, but it's otherwise consistent with how I'd act for changes of greater import

sshine: Thanks for linking to #522.

I've removed `# 2016-07-26` in a separate commit.

I agree, it's not so important. But I'd like if the versions are easily extracted.

petertseng:
Oh, something I just realised - The motivations written for this PR indicate that this is geared toward students.

In that case, you might consider simply not adding the line to examples' package.yaml, since they're already being run with `-Wall -Werror` due to https://github.com/exercism/haskell/blob/master/bin/test-example#L53 .

Then again, it's already been done, so shrug.

sshine:
Good point. But should students also have `-Werror` set?

petertseng:
We can probably base that decision on factors such as:

* how strongly mentors feel about students not having warnings
* how much additional burden it will place on students

I caution that it may cause the process of working on an exercise to be more cumbersome. I imagine that during the course of that work there are a few unused bindings, etc. Sometimes I just want to see whether it compiles, even though I have some unused bindings.

That was of course quite speculative. It may be valuable to try to complete an exercise from scratch with `-Wall -Werror` all the way and see what happens.

petertseng:
Righto righto, let me do the README business and this will be good, there will be one PR and one issue for the README business

sshine:
What's the README business? Would you like to add a note saying something about the possible warnings?

petertseng:
I'd like to try really hard never to merge a `Merge branch 'master' into (other branch)` into master. I really hate having the unnecessary commit in the history.

In PRs that should be merged as one commit, that means Squash and Merge is suitable.

But what to do about PRs that should be merged as multiple commits, such as this one?
We found that Rebase and Merge *does* eliminate the `Merge branch 'master' into (other branch)` commit (https://github.com/exercism/haskell/pull/444#issuecomment-263366365), but the disadvantage of that one is that I couldn't find a way to tell, from the command line, which PR the commits came from; that info only shows up in the UI. So I think the way to do it that meets the desires is to rebase from the command line, then use Create a Merge Commit.

I guess the other solution is to ask for the up-to-date requirement to be turned off. Honestly I thought that https://github.com/exercism/discussions/issues/174 indicates there's a desire to have it on for all repos, but I know of at least two track repos that do not have that option.

sshine:
I think I pushed a button that caused this extra merge from master. I really wanted to push it. But it seems like if I had rebased before pushing to my own branch, then I could have merged without clicking it. How about if we ask committers to rebase from master as the last thing they do, since there is no "rebase from master" button, but only a "merge from master" button?

petertseng:
Yeah, that sounds right.

sshine:
Excellent. Learning the GitHub workflow here. :-)

PR #774, opened 2018-11-19
Use rounding in complex-numbers tests

### LINKED ISSUE #772: complex-numbers: Use rounding in tests
It seems that the PR #664 is stuck because of this.

We perform rounding in [the tests of `space-age`](https://github.com/exercism/haskell/blob/master/exercises/space-age/test/Tests.hs).

It doesn't seem that Hspec has any helper functions for dealing with epsilon tests.

So perhaps we should just copy the behavior in the tests of `space-age`.

tejasbubane:
Closes #772

### DIFF
--- a/exercises/complex-numbers/test/Tests.hs
+++ b/exercises/complex-numbers/test/Tests.hs
@@ -1,12 +1,14 @@
 {-# LANGUAGE RecordWildCards #-}
 
 import Prelude hiding    (abs, div)
+import Data.Function     (on)
 import Data.Foldable     (for_)
 import Test.Hspec        (Spec, describe, it, shouldBe)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
 
 import ComplexNumbers
-  ( conjugate
+  ( Complex
+  , conjugate
   , abs
   , real
   , imaginary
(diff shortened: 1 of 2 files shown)


exercises/complex-numbers/test/Tests.hs
```diff
@@ -32,12 +34,18 @@ specs = do
...
+                                        `shouldBeAround` complex expectedA
   testB f CaseB{..} = it descriptionB $ f (complex number1B)
-                                        `shouldBe` complex expectedB
+                                        `shouldBeAround` complex expectedB
   testC f CaseC{..} = it descriptionC $ f (complex number1C)
                                         `shouldBe` expectedC
+  shouldBeAround = shouldBe `on` roundComplex 2
 
+  roundComplex :: Int -> Complex Float -> Complex Float
+  roundComplex n c = complex (roundTo n (real c), roundTo n (imaginary c))
```
tejasbubane: I am not sure if there is a better way to do this, but we don't have access to `Complex` data constructor here so cannot do this:

```haskell
roundComplex n (Complex x y) = ..
```

sshine: Looks good to me.

sshine:
Thanks. :-)

PR #775, opened 2018-11-21
WIP: hamming: Add generator/shrinker to property-based tests

sshine:
The essential properties are already being tested:

 - Strands with different lengths have no distance.
 - Strands with same lengths and two nucleotides in the
   same position that are different, have one more length.

The generator adds three cases for nucleotides:

 - Same nucleotide => same length.
 - Different nucleotides => +1 length.
 - Missing nucleotide => no length.

`shrink` produces all combinations of removing nucleotides pairwise:

 - Same nucleotide => same length.
 - Different nucleotides => -1 length.

For missing nucleotides, `shrink` relies on `expected = Nothing`.

### DIFF
--- a/exercises/hamming/test/Tests.hs
+++ b/exercises/hamming/test/Tests.hs
@@ -4,26 +4,33 @@
 import Data.Foldable     (for_)
 import Test.Hspec        (Spec, describe, it, shouldBe)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+import Data.List         ((\\))
+import Data.Ord          (comparing)
+import Test.QuickCheck
 
 import Hamming (distance)
+
+instance Arbitrary Case where
+  arbitrary = sized hammingGen
+  shrink = hammingShrink
 
 main :: IO ()
 main = hspecWith defaultConfig {configFastFail = True} specs
 
 specs :: Spec
-specs = describe "distance" $ for_ cases test
+specs = describe "distance" $ do
+  for_ cases test
+  it "works in general" (property test')
   where
-
-    test Case{..} = it description assertion
-      where
-        assertion  = expression `shouldBe` fromIntegral <$> expected
-        expression = distance strand1 strand2
+    test Case{..} = it description (distance strand1 strand2 `shouldBe` (fromIntegral <$> expected))
+    test' Case{..} = distance strand1 strand2 === (fromIntegral <$> expected)
 
 data Case = Case { description :: String
                  , strand1     :: String
                  , strand2     :: String
                  , expected    :: Maybe Integer
                  }
+            deriving (Show, Eq)
 
 cases :: [Case]
 cases = [ Case { description = "empty strands"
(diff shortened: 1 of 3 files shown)


sshine:
I'm not sure why it fails to find Test.QuickCheck.

petertseng:
So, how directly do you want me to give you the answer, as opposed to the process of discovery?

for example, does https://github.com/exercism/haskell/blob/master/bin/test-example#L25-L29 help?

petertseng:
Or if the process of discover exceeds your patience, then rot13(lbh nyfb arrq gb nqq dhvpxpurpx gb uggcf://tvguho.pbz/rkrepvfz/unfxryy/oybo/znfgre/rkrepvfrf/unzzvat/rknzcyrf/fhpprff-fgnaqneq/cnpxntr.lnzy)

exercises/hamming/test/Tests.hs
```diff
@@ -1,29 +1,37 @@
 {-# OPTIONS_GHC -fno-warn-type-defaults #-}
 {-# LANGUAGE RecordWildCards #-}
 
+import Data.Ord          (comparing)
+import Data.List         ((\\))
+import Control.Arrow     (second)
```
petertseng: optional: consider alphabetising these

exercises/hamming/test/Tests.hs
```diff
@@ -102,3 +110,65 @@ cases = [ Case { description = "empty strands"
...
+    desc' = "Strands of " ++ length' ++ " with " ++ dist' ++ " distance"
+    dist' = maybe "invalid" show expected
+    length' = case comparing length strand1 strand2 of
+      EQ -> "lengths " ++ show (length strand1)
+      _  -> "different lengths " ++ show (length strand1) ++ " and " ++
+            show (length strand2)
+
+nucleotidePairGen :: Gen (String, String, Maybe Integer)
+nucleotidePairGen = frequency
+  [ (20, zeroDistance)
```
petertseng: any particular reason these particular relative frequencies were chosen? Arbitrary, or was there some driving principle?

In either case, I'd suggest a code comment saying it, so that a next reader of the code will better understand if it is okay to change it to other values

exercises/hamming/test/Tests.hs
```diff
@@ -102,3 +110,65 @@ cases = [ Case { description = "empty strands"
...
+          }
+
+hammingDesc :: Case -> Case
+hammingDesc case'@Case{..} = case' { description = desc' }
+  where
+    desc' = "Strands of " ++ length' ++ " with " ++ dist' ++ " distance"
+    dist' = maybe "invalid" show expected
+    length' = case comparing length strand1 strand2 of
+      EQ -> "lengths " ++ show (length strand1)
+      _  -> "different lengths " ++ show (length strand1) ++ " and " ++
```
petertseng: optional: I guess this is not yet at the point where I would start suggesting `printf`, but I was considering it

exercises/hamming/test/Tests.hs
```diff
@@ -102,3 +110,65 @@ cases = [ Case { description = "empty strands"
...
+    oneDistance = do
+      a <- elements "ACGT"
+      b <- elements ("ACGT" \\ [a])
+      return ([a], [b], Just 1)
+
+    invalidDistance = do
+      a <- elements "ACGT"
+      return ([a], [], Nothing)
+
+without :: [a] -> [(a, [a])]
```
petertseng: note for me, in case I forget when coming back to it later:

`without [1, 2, 3] = [(1,[2,3]),(2,[1,3]),(3,[1,2])]`

exercises/hamming/test/Tests.hs
```diff
@@ -102,3 +110,65 @@ cases = [ Case { description = "empty strands"
...
+without [] = []
+without (x:xs) = (x, xs) : map (second (x:)) (without xs)
+
+hammingShrink :: Case -> [Case]
+hammingShrink case'@Case{..} = map hammingDesc $
+  zipWith shrink' (without strand1) (without strand2)
+  where
+    shrink' (a, strand1') (b, strand2')
+      | a == b = case' { strand1 = strand1', strand2 = strand2' }
```
petertseng: optional: I'm sure it'd be possible to write the `strand1 = strand1', strand2 = strand2'` once instead of twice as it it is now, but I'm not convinced that's necessary.

exercises/hamming/test/Tests.hs
sshine: Will do.

exercises/hamming/test/Tests.hs
sshine: They were chosen "experimentally" such that the probability of producing a negative input is somewhat low. But at this point I've decided that I should rather decide from the top whether it is a positive or a negative case; the current approach felt easy and interesting, but the probabilities are not clear or ideal. (For example, in increasing the size parameter, negative cases will dominate.)

exercises/hamming/test/Tests.hs
sshine: I'll add a comment and name it `select` from Prolog.

exercises/hamming/test/Tests.hs
sshine: There may be a neater way to express this.

exercises/hamming/test/Tests.hs
sshine: I will look into this in some weeks.

sshine:
I will close this PR and focus on property-based tests for exercises that don't have adequate unit tests.

PR #782, opened 2018-11-26
dnd-character: add new exercise

sshine:
This exercise was created in exercism/problem-specifications#1397.

It is a basic exercise in random generators, so the idea I had for the
Haskell track was to practice making QuickCheck generators rather than
simply use the standard random generator.

For this reason, QuickCheck is added as a root dependency.

### DIFF
--- a/exercises/dnd-character/README.md
+++ b/exercises/dnd-character/README.md
@@ -1,0 +1,93 @@
+# Dnd Character
+
+For a game of [Dungeons & Dragons][DND], each player starts by generating a
+character they can play with. This character has, among other things, six
+abilities; strength, dexterity, constitution, intelligence, wisdom and
+charisma. These six abilities have scores that are determined randomly. You
+do this by rolling four 6-sided dice and record the sum of the largest three
+dice. You do this six times, once for each ability.
+
+Your character's initial hitpoints are 10 + your character's constitution
+modifier. You find your character's constitution modifier by subtracting 10
+from your character's constitution, divide by 2 and round down.
+
+Write a random character generator that follows the rules above.
+
+For example, the six throws of four dice may look like:
+
+* 5, 3, 1, 6: You discard the 1 and sum 5 + 3 + 6 = 14, which you assign to strength.
+* 3, 2, 5, 3: You discard the 2 and sum 3 + 5 + 3 = 11, which you assign to dexterity.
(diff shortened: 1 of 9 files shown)


sshine:
@petertseng: How do I run the checks of Travis CI locally before submitting a PR?

It appears that it's the following lines from .travis.yml that triggers the failure:

    for exercise in ${TRAVIS_BUILD_DIR}/exercises/*/ ; do
        time bin/test-stub $exercise
        time bin/test-all-examples $exercise
    done

Perhaps this could be delegated to a script in `bin/` that is compatible with both Travis and running it locally?

My current workflow implies a lot of forth-and-back between the PR and Travis before succeeding.

exercises/dnd-character/test/Tests.hs
```diff
@@ -0,0 +1,72 @@
...
+    property (forAll ability abilityWithinRange)
+
+  describe "character" $ it "generates valid characters" $
+    property (forAll character characterIsValid)
+  where
+    test Case{..} = it description assertion
+      where
+        description = "computes the modifier for " ++ showInput input ++ " to be " ++ showMod expected
+        assertion = modifier input `shouldBe` expected
+        showInput n
```
petertseng: optional: if you so desire, it would be possible to achieve these six lines with a printf

exercises/dnd-character/README.md
```diff
@@ -0,0 +1,93 @@
+# Dnd Character
```
petertseng: oh right, I forgot. You should add a `title` to https://github.com/exercism/problem-specifications/blob/master/exercises/dnd-character/metadata.yml if you want the first word to show up as something other than Dnd. Perhaps you want DnD or DND.

sshine:
Thanks! I'll address your comments on Monday.

sshine:
Since I haven't advanced this one yet, note to self:

- [ ] Make a `.meta/hints.md` for this exercise.

exercises/dnd-character/test/Tests.hs
sshine: Done.

exercises/dnd-character/README.md
sshine: Thanks, I just did that in exercism/problem-specifications#1417.

sshine:
I've added some links in the hints section, changed the title, and switched to `RecordWildCards` in the example.

Issue #783, opened 2018-11-26
Make `fail-*` examples not show up in red in Travis logs


sshine:
When inspecting Travis CI logs, such as [this one](https://travis-ci.org/exercism/haskell/jobs/459782563#L560), the first red text that pops up relates to `exercises/accumulate/examples/fail-reverse`, which is a "fail example".

It's supposed to fail, but the red text in the Travis logs makes it pop up as something went wrong when it failed successfully.

We'd like to get rid of these red log messages so that scrolling the Travis log for actual errors doesn't flash these false negatives.

You can discuss any potential solution in this issue before submitting a PR if you're unsure that the approach is an example of good practice. (Exactly what is good practice here is uncertain to me; it could be that negative examples should be executed differently, or it could just be a shellscript hack that remaps ANSI color codes.)

- [x] Document how these negative examples are used for something. See [this section](https://github.com/exercism/haskell#example-solution).
- [ ] Suppress the red color and/or move negative examples to some other form.

**Edit:** Updated to clarify the remaining scope of this issue.

petertseng:
Excellent!

I have kicked off the first item with a PR.

sshine:
This problem no longer applies since #930 migrated to GitHub Actions which does not color its log.

PR #784, opened 2018-11-26
README: Document the reasons/uses of each type of example

petertseng:
Although the README listed the three types that were proposed and
accepted in https://github.com/exercism/haskell/issues/397, it failed to
explain when we would want to use each type.

We rectify this by explaining each type.

Checks the first box (consider/document/discuss) of
https://github.com/exercism/haskell/issues/783

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,9 +156,17 @@ Examples are named `<type>-<name>`.
 There are three possible types of examples:
 
 * success: The example is expected to pass the tests.
-  There should be at least one of these per exercise.
+  * There _must_ be at least `success` example per exercise, in order to confirm that it is possible to solve the tests.
+  * There _may_ be more than one `success` example for a given exercise, but these are intended for use when we want to confirm that multiple type signatures for a given solution will compile and pass the tests.
+  * We do not intend to use multiple `success` examples just to showcase a wide variety of possible solutions, since that is not in the goals of this repository.
 * fail: The example is expected to build, but fail the tests.
+  * These are intended for use when we want to make sure that the track tests have coverage: Whether the tests find certain classes of incorrect or inefficient solutions.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -156,9 +156,17 @@ Examples are named `<type>-<name>`.
...
+  * There _may_ be more than one `success` example for a given exercise, but these are intended for use when we want to confirm that multiple type signatures for a given solution will compile and pass the tests.
+  * We do not intend to use multiple `success` examples just to showcase a wide variety of possible solutions, since that is not in the goals of this repository.
 * fail: The example is expected to build, but fail the tests.
+  * These are intended for use when we want to make sure that the track tests have coverage: Whether the tests find certain classes of incorrect or inefficient solutions.
+  * It's suggested that these only be used for tests that are specific to the track. This is under the assumption that tests sourced from problem-specifications have already been judged to have appropriate coverage by the reviewers of the problem-specifications repository.
 * error: The example is expected to fail to build.
+  * There is only one intended use of this so far, and that is a single check that a solution without a type signature will fail to build (because CI builds with `--pedantic`).
```
petertseng: Aside: This is of questionable usefulness as actionable information for us, as explained as follows:

If the `error-nosig` example ever successfully builds, that means there is a problem with `stack --pedantic`. What would we do differently in that situation?

Wait for `stack` to solve this problem, potentially reporting it if nobody has before?

We would have to be on our guard that other instances of `--pedantic` are not being respected, I suppose? So, the single `error` example sort of serves as a canary in the ~cole~ coal mine?

So I guess it's useful info, but given that the most likely mode of failure is out of our control, having that cause all further CI to fail would be unfortunate. My ideal would be keeping the information available but not having it break CI.

sshine: I don't mind that we have it now that we do.

sshine:
Should we merge this now?

Issue #780, opened 2018-11-25
Use `Data.Text` when appropriate


sshine:
- [x] List currently implemented exercises that rely heavily on `String`.
- [x] ~~Consider extending their test suites so they're compatible with `Data.Text`.~~
- [x] ~~Consider moving them so they're follow-up exercises to Bob.~~
- [x] ~~Consider if any of them may be good as a first pure `Data.Text` exercise.~~

**Updated:**
- [x] List currently implemented exercises where the test suite can support both `String` and `Text`

sshine:
Exercises with `strings` in topics:

- `hello-world` (core), `bob` (core), `pangram` (core), `acronym`, `isogram`, `diamond`, `etl`, `raindrops`, `twelve-days`, `run-length-encoding`, `anagram`, `series`, `atbash-cipher`, `word-count`, `luhn`, `ocr-numbers`, `rotational-cipher`, `parallel-letters-frequency`, `pig-latin`, `rail-fence-cipher`.

sshine:
Links that may be useful as hints for working with `Data.Text`:

- https://mmhaskell.com/blog/2017/5/15/untangling-haskells-strings (bit poor styling)
- https://hackage.haskell.org/package/text-1.2.3.1/docs/Data-Text.html (bit technical)
- http://blog.ezyang.com/2010/08/strings-in-haskell/ (bit old)
- https://haskell-lang.org/tutorial/string-types (nice! excellent in fact!)

sshine:
I think we should either

1. `bob`: Add a comment in README that encourages `Text`.
2. `bob`: Switch entirely to `Text` just like #785 does to `pangram`.

The choice comes down to whether we should teach that there is a difference by first letting the student use the default, sub-optimal choice, and then progress to use a better, but more involved choice (since it depends on qualified imports and `OverloadedStrings`). Or if we should make that choice for them.

For some reason I find this decision very difficult. @reactormonk thinks we should kill `String`.

I agree that this is the end goal. But while Prelude uses it, there will be confusion.

reactormonk:
I'd argue that we should add a disclaimer that `Prelude` is old, and never build for production use, but rather teaching purposes. And maybe suggest one of our own, we could have a small list of curated ones.

sshine:
Since we're a teaching site is then Prelude should be the right choice. But since we also want to teach of choices beyond Prelude, perhaps custom preludes can be a subject later in the core track?

[...]

PR #785, opened 2018-11-27
WIP: Make Pangram based on `Data.Text`

sshine:
Pangram is the second string-heavy core exercise with Bob being the first.

As part of the `Data.Text` initiative started in #760 and tracked in #780, pangram will be the first exercise where the stub assumes the use of `Data.Text` rather than `String`.

Because the Haskell ecosystem is still adapting, and the Prelude and popular teaching material doesn't cover this extensively yet, we'll have to provide some insight into why this is desirable.

The test suite is still compatible with `String`-based solutions, but the stub does not encourage them.

### DIFF
--- a/exercises/pangram/examples/success-text/src/Pangram.hs
+++ b/exercises/pangram/examples/success-text/src/Pangram.hs
@@ -1,0 +1,13 @@
+module Pangram (isPangram) where
+
+import qualified Data.Set as S
+import           Data.Set (Set)
+
+import qualified Data.Text as T
+import           Data.Text (Text)
+
+isPangram :: Text -> Bool
+isPangram = S.null . T.foldl' (flip S.delete) alphabet . T.toLower
+  where
+    alphabet :: Set Char
+    alphabet = S.fromList ['a'..'z']
--- a/exercises/pangram/README.md
+++ b/exercises/pangram/README.md
@@ -8,6 +8,12 @@ The best known English pangram is:
 The alphabet used consists of ASCII letters `a` to `z`, inclusive, and is case
 insensitive. Input will not contain non-ASCII symbols.
 
+## Hints
+
+Haskell has many string types with `String` being the default. This exercise,
+however, asks that you explore the type `Text` instead. You can read more about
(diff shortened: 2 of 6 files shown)


exercises/pangram/examples/success-text/src/Pangram.hs
```diff
@@ -0,0 +1,11 @@
...
+module Pangram (isPangram) where
+
+import qualified Data.Text as T
+import           Data.Text (Text)
+
+isPangram :: Text -> Bool
+isPangram s = T.all (`member` T.toLower s) alphabet
+  where
+    alphabet = "abcdefghijklmnopqrstuvwxyz"
+    member = T.any . (==)
```
sshine: I don't like this definition particularly much, but it seems that `Data.Text` doesn't have a less general membership function, and since `Text` isn't `Foldable`, I can't use `elem`.

sshine:
How do I add the extra bit to README in such a way that `bin/ensure-readmes-are-updated.sh` accepts it?

exercises/pangram/examples/success-text/src/Pangram.hs
```diff
@@ -0,0 +1,13 @@
+module Pangram (isPangram) where
+
+import qualified Data.Set as S
+import           Data.Set (Set)
+
+import qualified Data.Text as T
+import           Data.Text (Text)
+
+isPangram :: Text -> Bool
+isPangram = S.null . T.foldl' (flip S.delete) alphabet . T.toLower
```
sshine: I've gone with a strict fold instead of lazy termination; while a list-based approach might terminate early by using `S.any null . L.scanl (flip S.delete)`, lazy `Text` is [generally discouraged](https://haskell-lang.org/tutorial/string-types) in favor of *conduits* and *pipes*.

sshine:
I have come to disagree with Pangram, with its current position on the core track, being a good exercise to force `Data.Text` if we're aiming for a gradual introduction to `Data.Text`. The `Data.Text`-based solution is almost equivalent, and the main difference seems to be that lazy solutions (ones that terminate once an input has been determined as pangram before its end) are made more difficult.

Issue #761, opened 2018-11-08
Replace `Difference Of Squares` and `Grains` on the core track


sshine:
After a discussion on #track-haskell @ Slack, I'm proposing here that we move both `Difference Of Squares` and `Grains` away from the core track and replace these two last spots with other exercises.

A short argument is that they don't provide a lot of opportunities to teach Haskell idioms or data types. Finding explicit formula instead of recursive formula is more of a mathy exercise. While I like `Grains`, I don't think it deserves to be the last core exercise.

In coming up with alternatives, not a lot of exercises have been suggested: `Bracket Push`, `Nucleotide Count`. We could revise the order if the ones we come up with are not ideal to place at the end of the track. In fact, I might like to suggest that either way.

Finally, this might be a good opportunity to port new exercises, since we can do it in such a way that they apply to the track. I said on Slack that I miss the following things from the core track: Trees (like `Binary Search Tree`), Maps (like `Nucleotide Count`), higher-order functions (that aren't list combinators), `Either` (or [a variant that accumulates errors](https://hackage.haskell.org/package/validation)), `Monoid`, parsing, list comprehensions, state monads, applicative functors, QuickCheck (properties and generators).

So how about we find / come up with exercises that cover things we haven't covered already, and then think about the order on the core track?

petertseng:
If https://github.com/exercism/haskell/pull/727 is an Exercism-wide policy, we need to remove not only difference-of-squares but also sum-of-multiples, but I suppose I can defer its removal until after there's a replacement for... whatever role in the sequence it plays.

I must admit I'm better at thinking of exercises that match a given topic rather than deciding what topics should go in the core curriculum. Let's see what we came up with.

> Trees

Don't think we have a good one right now. POV is about them, but too much for an intro. `binary-search-tree` does exist but that's making students implement their own rather than  a situation where we can use trees

> Maps

maybe `etl` can be good here?

> higher-order functions (that aren't list combinators)

there is the fact that `accumulate` (poorly named!) and `strain` have students implement them. I'm hard pressed to find an exercise that would suggest their use though since there are so many ways to solve a given exercise.

> `Either`

nucleotide-count or largest-series-product perhaps? for this one there are already a few exercises that have `Either` so mostly it's grepping the repo for it and picking one

> `Monoid`

Eh there was once mention of that in binary-search tree https://github.com/exercism/haskell/issues/59 , but I'm not convinced it's strong enough to build the connectoin

> parsing

[...]

PR #789, opened 2018-12-07
Replace difference-of-squares with nucleotide-count on core track

sshine:
Nucleotide Count has several reasonable implementations and learning goals. Mentor notes are made available in exercism/website-copy#654 and the discussion to replace some core track exercises is advanced in #761.

### DIFF
--- a/config.json
+++ b/config.json
@@ -90,14 +90,13 @@
       ]
     },
     {
-      "slug": "difference-of-squares",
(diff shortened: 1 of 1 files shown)


sshine:
I'm working on a `Data.Text`-based version of Nucleotide Count, but it involves [`mono-foldable`](http://hackage.haskell.org/package/mono-foldable) which I haven't messed with before, so there's that.

PR #792, opened 2018-12-14
Leap: Add link to guards to hints

sshine:
While many students submit something with guards, some don't.

This provides learning material up-front.

I'd like to also wait for exercism/problem-specifications#1418 before merging this.


sshine:
Also, the reason why I didn't pick [LYAH's chapter on guards](http://learnyouahaskell.com/syntax-in-functions#guards-guards) is because I don't like that their first example is about fat-shaming. I've copied this link dozens of times to students already, but it isn't before I'm committing this to history that I realize how rude it is.

PR #796, opened 2019-01-30
Hamming: Add a test case to avoid wrong recursion solution

tqa236:
This solution will pass all the current test cases but fail on the new one:

```
module Hamming (distance) where


distance :: String -> String -> Maybe Int
distance [] []         = Just 0
distance (x:xs) (y:ys)
  | length(xs) /= length(ys) = Nothing
  | x /= y = fmap (1 + ) (distance xs ys)
  | x == y = distance xs ys
```

### DIFF
--- a/exercises/hamming/package.yaml
+++ b/exercises/hamming/package.yaml
@@ -1,5 +1,5 @@
 name: hamming
-version: 2.2.0.9
+version: 2.3.0.10
 
 dependencies:
   - base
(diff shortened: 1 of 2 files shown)


tqa236:
Hello,

I did all 4 points you said and request a new pull request. Please tell me if you notice anything wrong.

tqa236:
@sshine Is this PR ok to merge now or do we still need some changes?

sshine:
@tqa236: We're good. Thanks a lot, and sorry for the late response.

Issue #798, opened 2019-02-26
Restructure Haskell core track


sshine:
In an effort to restructure the Haskell core track I have followed @F3PiX'es Track Tool in its current incarnation. Step 1 took me through the current core track and yielded the initial proposal listed below.

- [x] Update `config.json` to reflect this via draft PR #800
- [x] Discuss these suggestions

---

- Switch "Bob" and "Pangram".

  They're both about handling textual data.

  Pangram has fewer gotchas, is quicker done, has more alternative solutions, and doesn't lend itself as well to `Data.Text`. (See #795.) Bob often involves more refactoring, has more learning objectives and has the extra "You could do it with `Data.Text`, too!" part.

  This way people don't try to solve Pangram with `Data.Text` which is possible but not ideal without [mono-traversable](https://hackage.haskell.org/package/mono-traversable), and they get an exercise more under their belt before being thrown into a lot of refactoring.

- Switch "Nucleotide Count" and "Sum of Multiples".

  Eventually I think we should delete "Sum of Multiples" and "Grains" (see #761).

  But since the Track Tool does not address removal yet, I want to make "Nucleotide Count" come immediately after "RNA Transcription", since they're both thematically related *and* Nucleotide Count uses `Either` while RNA Transcription uses `Maybe` for monadic error handling, so they practice the same thing in progressive difficulty.

---

The proposed track reordering looks like:

1. hello-world
2. leap
3. space-age
4. **pangram**
5. **bob**
6. collatz-conjecture
7. rna-transcription
8. **nucleotide-count**
9. *sum-of-multiples*
10. *grains*

sshine:
Related:

- #192: discussion: Making the Haskell track more interesting for beginners

  Perhaps we can get an updated graph similar to this one.

- #194: discussion: Making the Haskell track more idiomatic

petertseng:
I don't know what this Track Tool is or what steps comprise it, so it doesn't appear possible to evaluate whether the proposals meet any of its stated goals.

However, each proposal has been justified standalone and can be evaluated independently of knowledge of this Track Tool. I encourage that future proposals should continue to be justified in such a manner!

I agree with introducing `Maybe` before `Either`. This seems like a logical progression. I suspect (but don't have data to support) that other projects that educate also order them in this order.

I had to think a lot harder about pangram vs bob but I've come to agree. Both ask "is property P true of text T?" but pangram has only one P, where Bob has multiple (and not all mutually exclusive).

sshine:
[...]

PR #800, opened 2019-02-27
Switch order of current core exercises

sshine:
**Edit:** *As @petertseng points out below, embed the reasons for deciding into the commit message. As a consequence, the pull request message is updated to reflect this change. This change consists of moving the arguments from the discussion at #798 to this message instead.*

---

As part of [The Track Anatomy Project][1], the Haskell core track is
restructured in #798 using @F3PiX'es multi-stage framework referred to
as "The Track Tool".

Since The Track Tool is still under development, it isn't documented
yet. But the execution of the first step can be viewed for other
languages, e.g. exercism/javascript#590.

---

Proposal:

 - Switch "Bob" and "Pangram".

   They're both about handling textual data.

   Pangram has fewer gotchas, is quicker done, has more variations among
   its alternative solutions, and doesn't lend itself as well to `Data.Text`.
   (See #795.) Bob often involves more refactoring, has more learning
   objectives and has the extra "You could do it with Data.Text, too!" part.

   This way people don't try to solve Pangram with Data.Text which is possible
   but not ideal without mono-traversable, and they get an exercise more under
   their belt before being thrown into a lot of refactoring.

 - Switch "Nucleotide Count" and "Sum of Multiples".

   Eventually I think we should delete "Sum of Multiples" and "Grains" (see #761).

   But since The Track Tool does not address removal yet, I want to make
   "Nucleotide Count" come immediately after "RNA Transcription", since
   they're both thematically related and Nucleotide Count uses Either
   while RNA Transcription uses Maybe for monadic error handling, so they
   practice the same thing in progressive difficulty.

The proposed track reordering looks like:

1. hello-world
2. leap
3. space-age
4. **pangram**
5. **bob**
6. collatz-conjecture
7. rna-transcription
8. **nucleotide-count**
9. sum-of-multiples
10. grains

[1]: https://exercism.io/blog/track-anatomy-project

### DIFF
--- a/config.json
+++ b/config.json
@@ -37,8 +37,8 @@
       ]
     },
     {
-      "slug": "bob",
-      "uuid": "e570c730-12e5-4b9e-adb0-eae85360e1b5",
+      "slug": "pangram",
+      "uuid": "42cb6880-0c4b-4455-815b-aac94c05db18",
       "core": true,
       "unlocked_by": null,
       "difficulty": 1,
@@ -47,8 +47,8 @@
       ]
     },
     {
-      "slug": "pangram",
-      "uuid": "42cb6880-0c4b-4455-815b-aac94c05db18",
+      "slug": "bob",
+      "uuid": "e570c730-12e5-4b9e-adb0-eae85360e1b5",
       "core": true,
       "unlocked_by": null,
       "difficulty": 1,
@@ -79,6 +79,16 @@
       ]
     },
     {
+      "slug": "nucleotide-count",
+      "uuid": "00feb0e9-ae1e-43fc-a0ea-8f2d0d0b9f49",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 2,
+      "topics": [
+        "either"
+      ]
+    },
+    {
       "slug": "sum-of-multiples",
       "uuid": "e125a32e-60ca-4e59-ae35-01f3b1907ccd",
       "core": true,
@@ -87,16 +97,6 @@
       "topics": [
         "math",
         "number_theory"
-      ]
-    },
-    {
-      "slug": "nucleotide-count",
-      "uuid": "00feb0e9-ae1e-43fc-a0ea-8f2d0d0b9f49",
-      "core": true,
-      "unlocked_by": null,
-      "difficulty": 2,
-      "topics": [
-        "either"
       ]
     },
     {


petertseng:
Although I prefer to discuss a PR using comments on the PR, rather than a linked issue, I think the main objectives I wish to achieve are:

* For a given change, I want to be able to understand its reasoning and intent.
    * This remains possible because the commit message links to the issue.
    * I recommend that **the commit message be updated with a summary of the reasoning for both changes** in addition. A short summary such as "We should present `Maybe` before `Either`" and "Bob checks multiple properties of a string versus Pangram's one" seems to suffice here.
* For a given change, I'd like to make it as easy as possible to find all relevant discussion for it, so discussion for it should all happen in one place.
    * Therefore, I found it advantageous to follow suit, and since the existing discussion was in https://github.com/exercism/haskell/issues/798, I followed suit.
    * This means that I have to try to avoid discussion of the change here.

This means that, given the above are followed, I do not appear to have a reason to prefer one of the two choices (discussion in the same PR, discussion in a linked issue) over the other.

petertseng:
well, I suppose the difference is that I will discuss the *intent* of the changes in the linked issue, and discuss whether the intent is good. In the PR, I will discuss the *mechanics* of the change, and whether it actually matches the stated intent.

(I still don't mind combining the two discussions into the same place, but for PR where **both** avenues of discussion may be contentious, then it may be worth separating)

For this issue, the mechanics are fairly uncontentious, but just to be thorough, it is possible to verify that this PR does indeed only perform the two reorderings that it claimed to do.

```diff
diff -u <(ruby ../problem-specifications/tree.rb --flat <(git show master:config.json)) <(ruby ../problem-specifications/tree.rb --flat config.json)
--- /dev/fd/11	2019-03-06 01:05:18.000000000 +0000
+++ /dev/fd/12	2019-03-06 01:05:18.000000000 +0000
@@ -10,12 +10,12 @@
 hello-world
 leap
 space-age
-bob
 pangram
+bob
 collatz-conjecture
 rna-transcription
-sum-of-multiples
 nucleotide-count
+sum-of-multiples
 grains

 [side]
```

sshine:
Hi @petertseng, and thanks for your patience while best I try to place the discussion.

I've updated the commit message and the PR message to reflect the specific change.

> In the PR, I will discuss the mechanics of the change, and whether it actually matches the stated intent.
> (I still don't mind combining the two discussions into the same place, but for PR where both avenues of discussion may be contentious, then it may be worth separating)

In retrospect I think I would have preferred to keep all discussion of concrete changes to the PRs and more abstract, high-level discussion at #798. I'm not sure what's least confusing at this point; since some choices are made in a bigger picture, I'd like to go with some division a la *mechanics* vs. *abstract reasons*.

PR #802, opened 2019-03-05
rna-transcription: Improve README hint and example

### LINKED ISSUE #801: Issues with rna-transcription
Hello,
I've been doing the 'rna-transcription' exercise and I think that the tests are not consistent. It seems that when given String is not a proper DNA the programme should return 'X' but one of the test says that given 'U' the programme should return 'X' (even though the testcase is called 'invalid input').

I had a look at the example solution and it fails the test with the following error:

```
  src/DNA.hs:7:5: 
  1) toRNA correctly handles invalid input (RNA instead of DNA)
       uncaught exception: PatternMatchFail
       src/DNA.hs:(7,5)-(10,28): Non-exhaustive patterns in function fromDNA
```

b-mehta:
Fixes #801, and finishes propagating changes from #762.
Also modified `mapM` in the example solution to `traverse`, since `traverse` is more general and preferred, and `mapM` is typically just defined as an alias for it.

### DIFF
--- a/exercises/rna-transcription/README.md
+++ b/exercises/rna-transcription/README.md
@@ -18,6 +18,9 @@ each nucleotide with its complement:
 * `T` -> `A`
 * `A` -> `U`
 
+Given invalid output, your program should return the first invalid character.
+
+
 
 ## Getting Started
 
--- a/exercises/rna-transcription/.meta/hints.md
+++ b/exercises/rna-transcription/.meta/hints.md
@@ -1,0 +1,1 @@
+Given invalid output, your program should return the first invalid character.
--- a/exercises/rna-transcription/examples/success-standard/src/DNA.hs
+++ b/exercises/rna-transcription/examples/success-standard/src/DNA.hs
@@ -1,7 +1,7 @@
 module DNA (toRNA) where
 
 toRNA :: String -> Either Char String
-toRNA = mapM fromDNA
+toRNA = traverse fromDNA
   where
     fromDNA :: Char -> Either Char Char
     fromDNA 'C' = return 'G'


exercises/rna-transcription/README.md
```diff
@@ -18,6 +18,8 @@ each nucleotide with its complement:
 * `T` -> `A`
 * `A` -> `U`
 
+Given invalid output, your program should return the first invalid character.
+
```
sshine:
```suggestion


```

sshine: When you run `bin/ensure-readmes-are-updated.sh` it seems to want this extra linebreak.

b-mehta: Ah I was worried that this might happen, thanks.

petertseng:
All right, the overall changes are indeed what we want here.

Since this PR combines two changes (1. README hint; 2. mapM -> traverse), and the final state should be that each commit should contain one change, it is not possible to use GitHub's Squash and merge on this PR. Therefore the commits must be manually combined as appropriate. Could you squash the commits such that the two changes take up exactly one commit each?

(So basically, what I am saying is that the original PR submission did well to separate the two commits. All changes after the original submission should now be made as amendments to the first commit, rather than additional commits)

b-mehta:
@petertseng Is this what you meant?

petertseng:
You should remove the commit labeled `travis: Change language to bash`. That is part of another PR, not yours.

The other two commits are indeed what I meant, excellent :+1:

This is all I want to ask for. There is something I may have to ask for even though I don't want to (the "This branch is out-of-date with the base branch").

You could try to solve this by ensuring your branch contains every commit in https://github.com/exercism/haskell/commits/master . You might find it most convenient to update https://github.com/b-mehta/haskell/commits/master so that you can just `git rebase -i master`, but you might have luck just using the "Update branch" button, then using `git fetch` so you locally have the changes that land on the remote `patch-1`, then `git rebase -i 2f5ac669d3a60c33b63931bf34e86e42cb6d6342`

I will attempt to see if I can get a certain setting changed that will relieve you of having to do this, but if you want to handle this yourself, that's good too.

b-mehta:
That particular commit is there from updating the branch earlier, when my hope was that no further changes would have been necessary. I can remove it again, but then updating again would just put it back again.

petertseng:
if using `git rebase -i`, the line containig `travis: Change language to bash` should be **deleted**

petertseng:
The commit `Merge branch 'master' into patch-1` shouldn't be on this PR. Since the PR is going to be merged into master, we don't want to merge master into it. Having the extra commit is unnecessary extra.

b-mehta:
As I mentioned before, that commit came from the 'Update branch'.

petertseng:
> and the extra commit won't be shown after the merge happens

this is only true if we use Squash and merge (which we absolutely cannot; this PR should stay as two commits) or rebase and merge (which I don't prefer. it loses information of what PR these commits came from)

petertseng:
You should instead rebase on master (ensuring that the `master` being rebased on contains all the commits in https://github.com/exercism/haskell/commits/master), instead of merging master.

petertseng:
I want to use "Create a merge commit", therefore the only commits listed in https://github.com/exercism/haskell/pull/802/commits should be exactly the two commits that should appear on master.

You have correctly observed that otherwise, GitHub would ask to use Update branch, since the branch is out of date.

For both conditions to be true, the branch should be *rebased on master* (ensuring that the `master` being rebased on contains all the commits in https://github.com/exercism/haskell/commits/master ) , then it will meet both conditions:

1. up to date (GitHub does NOT say that *update branch* is necessary)
2. contains only the two commits we actually want to merge, no others like `Merge branch 'master' into patch-1`

petertseng:
I have used https://github.com/petertseng/test-repo/commits/master to demonstrate that the `Merge branch 'master' into patch-1` commit will stay, which is what we don't want. So we don't want it to show up in https://github.com/exercism/haskell/pull/802/commits

petertseng:
I can also mention what I would to do achieve the objectives. If you already have a way in mind to achieve them, you can use that way instead and ignore me:

```
git remote add exercism git@github.com:exercism/haskell.git
git fetch exercism
git checkout patch-1
git rebase -i exercism/master
# there should be two lines starting with `pick`, one for each commit
git push --force-with-lease
```

sshine:
@b-mehta: I took the liberty of experimenting with pushing to PR branches: when you created this PR, you allowed maintainers of this repository to push to `b-mehta:patch-1`. I've not tried this before, so I figured this would be a good opportunity. I've removed the merge commits by rebasing onto master as @petertseng instructed. The merge commits vanished as they were not necessary when applying the two commits on top of master.

I think the way to avoid this on PRs that have fallen behind is to *rebase* rather than *merge* the feature branch with master.

b-mehta:
Apologies both for the delay. 
@sshine: Thanks for sorting this!  Yes, I'd allowed maintainers to push for essentially this reason :)
@petertseng: Thanks for your clarification, and apologies for my misunderstanding.

PR #804, opened 2019-03-06
travis: build only one exercise

petertseng:
This serves to demonstrate the potential impact of
https://github.com/exercism/haskell/pull/751

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,6 @@ script:
         export SET_RESOLVER="--resolver ${RESOLVER}"
     fi
 
-    for exercise in ${TRAVIS_BUILD_DIR}/exercises/*/ ; do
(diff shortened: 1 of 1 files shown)


petertseng:
comparing https://travis-ci.org/exercism/haskell/builds/502376618 vs https://travis-ci.org/exercism/haskell/builds/499951244 : 3m37s vs 5m8s for building from warm cache. 20m vs 36m represents a situation where cache is cold on (I assume, but didn't check) both sides.

Note that for us it's likely the only situations where cache is cold is when we changed stack resolver, in which case we're building all exercises anyway. So the 3m37s vs 5m8s figure is the important one here.

It's all right (30% reduction, proportionally) but not game-changing.

petertseng:
This PR should not be merged.

Issue #769, opened 2018-11-15
Shouldn't all exercises have `ghc-options: -Wall` in their package.yaml?


sshine:
In `Space Age` I argue in my mentor feedback that one benefit of writing a total function,

    orbitalPeriod :: Planet -> Float
    orbitalPeriod Mercury = ...
    orbitalPeriod Venus = ...
    ...

is that GHC can warn you in case of missing or duplicate cases, and that this is useful when extending data types over time. But if you add a planet and forget to extend `orbitalPeriod`, `stack test` and `stack build` will not, actually, show this warning.

Which made me think, why don't we ad `ghc-options: -Wall` to package.yaml of all exercises?

petertseng:
Yes, not only is this a good idea on its own merits, the history supports it as well because https://github.com/exercism/exercism/pull/804 added `-Wall` to the then-instructions. It seems that was lost on the move to Stack.

sshine:
I've removed `ghc-options: -Wall` from all example package.yaml. Before the amend:

    $ git diff-tree --no-commit-id --name-only -r HEAD | wc -l
    201

    $ git diff-tree --no-commit-id --name-only -r HEAD | grep examples | wc -l
    105

After the amend:

    $ git diff-tree --no-commit-id --name-only -r HEAD | wc -l
    96

    $ git diff-tree --no-commit-id --name-only -r HEAD | grep examples | wc -l
    0

As 96 + 105 = 201, it seems that no other files were affected.

*(Actually, I did accidentally commit a rogue file and found it by doing this test. Hooray!)*

PR #807, opened 2019-03-08
binary-search: Add exercise (closes #615)

sshine:
Finishing @samosaara's work, this implements the binary-search exercise on the Haskell track using `Data.Array` from the `array` package. This seems to be the preferred choice cf. the discussion at #615.

In addition, a number of changes were made:

 - Unlock this after `collatz-conjecture` as the exercise deals with `Maybe`; placing it after the first exercise that deals with error handling seems natural.

 - Set exercise version number to 1.3.0.0 and update test suite. This addresses exercism/problem-specifications#1399 (version 1.3.0).

 - Add `ghc-options: -Wall` since #769.

 - Export function as `find`. This is what problem-specification calls it.

 - Generalise type signature in stub to better distinguish the type parameter for the index and the type parameter for the elements. One could generalise `Int` to `Ix i => i`, but this becomes slightly difficult when doing index arithmetic.

 - Make example solution work for arbitrary array bounds. Add test cases specific to the Haskell track that tests non-zero bounds. These cannot be expressed in problem-specifications right now, and they're only relevant for languages that can change the array base.

 - Add hint for optionally dealing with non-zero bounds.

 - Add hint with resources to `Data.Array`.

### DIFF
--- a/config.json
+++ b/config.json
@@ -560,6 +560,16 @@
       ]
     },
     {
+      "uuid": "e714ed04-e633-4814-8885-323968f39046",
+      "slug": "binary-search",
+      "core": false,
+      "unlocked_by": "collatz-conjecture",
+      "difficulty": 5,
+      "topics": [
+        "maybe"
+      ]
+    },
+    {
       "slug": "atbash-cipher",
       "uuid": "4b6d04f4-8cfc-4d8a-ac6e-77f3889f27c3",
       "core": false,
--- a/config.json
+++ b/config.json
@@ -560,8 +560,8 @@
       ]
     },
     {
-      "uuid": "e714ed04-e633-4814-8885-323968f39046",
       "slug": "binary-search",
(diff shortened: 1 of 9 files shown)


sshine:
Question: I disabled the Haskell-specific tests initially. Should they be enabled by default? If not, is there a better way, syntactically, to comment them out? (Simply commenting out `for_ cases $ test (-9)` makes HLint complain about a redundant `do` above.)

Comment: I wish there were better tutorials on `Data.Array`. Adding a third link with more hand-holding would be ideal. The best I've been able to find (that doesn't go into all the many irrelevant types of arrays in Haskell that only add to the confusion) is the blog post [Defeating Evil with Data Structures!](https://hackernoon.com/defeating-evil-with-data-structures-56335c0a6403), but this unfortunately is mainly about something called *segment trees*, and `Data.Array` is just a step on the way.

For now, the two links provided will have to suffice until someone finds a good tutorial.

Issue #694, opened 2018-07-21
nucleotide-count: Nucleotides have a misleading type


isovector:
The function `nucleotideCounts :: String -> Either String (Map Char Int)` I'd suggest has the wrong type. Nucleotides are certainly an enum and not a `Char`. I worry the exercise as it stands today discourages people from thinking about data modeling and instead plays into a bias to use strings more often than they should be.

petertseng:
I agree -- an enum would make sense so the type system can help make clear there are only four possible values at each position.

I would also take this opportunity to address what was said in https://github.com/exercism/haskell/issues/626#issuecomment-347012997

> While I think that enforcing invariables in the type system is a great idea, sometimes it is awkward to do that in Haskell. I don't think we should push all the exercises in this direction when it feels artificial in the language.
>
> About nucleotide-count, using length-parameterized types is idiomatic in Idris, but it is still considered type-hackery in Haskell, which is an advanced topic and would confuse students.

I still think there would be value in using a `data` for `nucleotide-count`; we don't have to do the Idris-style length-parameterisation.

I believe the files needed to edit to resolve this issue are the following:

* https://github.com/exercism/haskell/blob/master/exercises/nucleotide-count/test/Tests.hs
* https://github.com/exercism/haskell/blob/master/exercises/nucleotide-count/src/DNA.hs
* https://github.com/exercism/haskell/blob/master/exercises/nucleotide-count/examples/success-standard/src/DNA.hs

sshine:
I agree. But if we use a

    data Nucleotide = A | C | G | T

then what's the point in having a `Map Nucleotide Int` when you could just have a

    data NucleotideCount = NC { _a :: Int, _c :: Int, _g :: Int, _t :: Int }

but then you wouldn't need the `Nucleotide` data type if the original input is a `String`.

Perhaps one should keep the `Map Nucleotide Int` since that adds complexity.

sshine:
Should the tests keep asserting that the map contains all four nucleotides when a nucleotide occurred zero times?

isovector:
@sshine I personally wouldn't let the `NucleotideCount` described above through in a real code review---it doesn't compose well with the ecosystem. Eg, writing `Nucelotide -> NucelotideCount -> Int` requires a pattern match that the compiler won't help you write correctly. When in doubt, use the definitions that are the hardest to get wrong.

isovector:
@sshine nope, #714 looks great to me!

PR #808, opened 2019-03-09
Implement Resistor Colors exercise

tejasbubane:
Problem Specification: https://github.com/exercism/problem-specifications/pull/1466

### DIFF
--- a/exercises/resistor-colors/examples/success-standard/package.yaml
+++ b/exercises/resistor-colors/examples/success-standard/package.yaml
@@ -1,0 +1,16 @@
+name: resistor-colors
+
+dependencies:
+  - base
+
+library:
+  exposed-modules: ResistorColors
+  source-dirs: src
+
+tests:
+  test:
+    main: Tests.hs
+    source-dirs: test
+    dependencies:
+      - resistor-colors
+      - hspec
--- a/exercises/resistor-colors/examples/success-standard/src/ResistorColors.hs
+++ b/exercises/resistor-colors/examples/success-standard/src/ResistorColors.hs
@@ -1,0 +1,29 @@
+module ResistorColors (Color(..), value) where
+
+data Color =
+    Black
+  | Brown
+  | Red
+  | Orange
+  | Yellow
+  | Green
+  | Blue
+  | Violet
+  | Grey
+  | White
+  deriving (Eq, Show, Read)
+
+convert :: Color -> Int
+convert Black = 0
+convert Brown = 1
+convert Red = 2
+convert Orange = 3
+convert Yellow = 4
+convert Green = 5
+convert Blue = 6
+convert Violet = 7
+convert Grey = 8
+convert White = 9
+
+value :: [Color] -> Int
+value = read . concatMap (show . convert)
--- a/exercises/resistor-colors/src/ResistorColors.hs
+++ b/exercises/resistor-colors/src/ResistorColors.hs
@@ -1,0 +1,17 @@
+module ResistorColors (Color(..), value) where
+
+data Color =
+    Black
+  | Brown
+  | Red
+  | Orange
+  | Yellow
+  | Green
+  | Blue
+  | Violet
+  | Grey
+  | White
+  deriving (Eq, Show, Read)
+
+value :: [Color] -> Int
+value cs = error "You need to implement this function."
--- a/exercises/resistor-colors/test/Tests.hs
+++ b/exercises/resistor-colors/test/Tests.hs
@@ -1,0 +1,44 @@
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
+{-# LANGUAGE RecordWildCards #-}
+
+import Data.Foldable     (for_)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+
+import ResistorColors (Color(..), value)
+
+main :: IO ()
+main = hspecWith defaultConfig {configFastFail = True} specs
+
+specs :: Spec
+specs = describe "value" $ for_ cases test
+  where
+
+    test Case{..} = it explanation assertion
+      where
+        explanation = unwords [show input, "-", description]
+        assertion   = value input `shouldBe` expected
+
+data Case = Case { description :: String
+                 , input       :: [Color]
+                 , expected    :: Int
+                 }
+
+cases :: [Case]
+cases = [ Case { description = "Brown and black"
+               , input       = [Brown, Black]
+               , expected    = 10
+               }
+        , Case { description = "Blue and grey"
+               , input       = [Blue, Grey]
+               , expected    = 68
+               }
+        , Case { description = "Yellow and violet"
+               , input       = [Yellow, Violet]
+               , expected    = 47
+               }
+        , Case { description = "Orange and orange"
+               , input       = [Orange, Orange]
+               , expected    = 33
+               }
+        ]
--- a/config.json
+++ b/config.json
@@ -214,6 +214,16 @@
       ]
     },
     {
+      "slug": "resistor-colors",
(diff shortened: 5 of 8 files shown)


tejasbubane:
I keep getting this segmentation fault while running the `ensure-readmes-updated` script:

```
➜ bin/ensure-readmes-are-updated.sh
Cloning into 'problem-specifications'...
remote: Enumerating objects: 6, done.
remote: Counting objects: 100% (6/6), done.
remote: Compressing objects: 100% (6/6), done.
remote: Total 8462 (delta 0), reused 6 (delta 0), pack-reused 8456
Receiving objects: 100% (8462/8462), 1.70 MiB | 493.00 KiB/s, done.
Resolving deltas: 100% (4986/4986), done.
Checking readme for resistor-colors
bin/ensure-readmes-are-updated.sh: line 20:  1144 Segmentation fault: 11  $configlet generate . --only "$exercise" --spec-path "problem-specifications"
```

petertseng:
Since I couldn't find output containing segmentation fault in the Travis CI output, I am taking your word for it.
Since the output you have displayed says that configlet is segfaulting, it sounds like the thing to do is to find a minimal set of conditions to reproduce this and then report it to https://github.com/exercism/configlet/issues

Since I don't use configlet I regret being unable to assist in the specifics of that

exercises/resistor-colors/examples/success-standard/src/ResistorColors.hs
```diff
@@ -0,0 +1,34 @@
...
+convert Red = 2
+convert Orange = 3
+convert Yellow = 4
+convert Green = 5
+convert Blue = 6
+convert Violet = 7
+convert Grey = 8
+convert White = 9
+
+value :: [String] -> Int
```
sshine: Since we're converting `["black", "brown"]`, I think it would make sense to

1. Have error handling.
2. Return an `Integer`.

Since it makes sense to have error handling, it also makes sense to ask if we should go with `Maybe` or `Either`.

Then `readMaybe` or `readEither` from [`Text.Read`](http://hackage.haskell.org/package/base/docs/Text-Read.html) can come into use.

sshine:
>     bin/ensure-readmes-are-updated.sh: line 20:  1144 Segmentation fault: 11  $configlet generate . --only "$exercise" --spec-path "problem-specifications"

That's strange, line 20 only contains `git`, `grep`, `cut` and `sort`.

`configlet` isn't run until line 20.

What if you `echo` the line that begins with `$configlet` and run that manually?

It sounds like `configlet` fails because of the way it's compiled...

exercises/resistor-colors/examples/success-standard/src/ResistorColors.hs
```diff
@@ -0,0 +1,34 @@
...
+  | Yellow
+  | Green
+  | Blue
+  | Violet
+  | Grey
+  | White
+  deriving (Eq, Show, Read)
+
+convert :: Color -> Int
+convert Black = 0
```
petertseng: I notice that it would be *possible* to use http://hackage.haskell.org/package/base-4.12.0.0/docs/Prelude.html#t:Enum here.

I have not yet determined whether it would be a *good idea* to do so.

exercises/resistor-colors/examples/success-standard/package.yaml
```diff
@@ -0,0 +1,16 @@
+name: exercism-resistor-colors
```
petertseng: Since it's not necessary to prepend `exercism-` to the name, I think we should prefer not to do so, and only do so when it is necessary (as it is in `clock`, the only exercise that does so)

exercises/resistor-colors/test/Tests.hs
```diff
@@ -0,0 +1,44 @@
...
+        assertion   = value input `shouldBe` expected
+
+data Case = Case { description :: String
+                 , input       :: [String]
+                 , expected    :: Int
+                 }
+
+cases :: [Case]
+cases = [ Case { description = "Brown and black"
+               , input       = ["brown", "black"]
```
petertseng: Depending on what we want the exercise to teach, we might choose to instead use `Brown` and `Black` here and in all other tests. In that case, we would provide that in the stub.

The similar discussion was had for nucleotide-count in https://github.com/exercism/haskell/issues/694

exercises/resistor-colors/examples/success-standard/src/ResistorColors.hs
sshine: I had the same thought.

What you don't get with `deriving Enum` is an explicit annotation that `Black` is 0, etc.

Making an explicit `instance Enum where ...` has the drawback that you need to specify both `toEnum` and `fromEnum`, even though the exercise only needs the one. [There are ways](https://stackoverflow.com/questions/6000511/better-way-to-define-an-enum-in-haskell) to avoid that, but the first way (using a dynamic lookup table) should be discouraged, and the second way (using `Data` and `Typeable`) is probably a little too complicated to serve as a good example.

So I'm content with having a simple `convert` function.

F3PiX:
I noticed that the Readme here is an older version of the Readme we actually implemented.

F3PiX:
@tejasbubane Can you regenerate the readme with 
`configlet regenerate -o resistor-colors`? 

(thanks/cc @ErikSchierboom )

exercises/resistor-colors/test/Tests.hs
sshine: I like this idea!

This exercise was introduced quite recently as a simple exercise that can be put into the core track. If the aim is to put it into the Haskell core track, we can purpose it for anything we like. I don't think we need an extra exercise that teaches monadic error handling (we currently have three already; `collatz-conjecture`, `rna-transcription` and `grains`).

`space-age` already has a data type with multiple constructors, and I use it to teach the difference between pattern matching all constructors vs. dynamic lookup tables, avoiding duplicate code, and naming things. This exercise could do many of the same things without `Float`.

Unless we make it about something that `space-age` isn't, e.g. `Enum`, I am however not sure I see the point in making it a core exercise. @tejasbubane: Let me know what you think wrt. making this a core exercise.

ErikSchierboom:
Oh, my mistake. That command should actually be: `./bin/configlet regenerate . -o resistor-colors`
And that assumes that there is an up-to-date `problem-specifications` repository located at `../problem-specifications`

exercises/resistor-colors/test/Tests.hs
tejasbubane: Changed.

exercises/resistor-colors/examples/success-standard/package.yaml
tejasbubane: Changed.

exercises/resistor-colors/examples/success-standard/src/ResistorColors.hs
tejasbubane: From the tests in specification, I assumed the input will always be valid. Let me know if I should make this change or wait for more discussion (in the problem spec as well).

tejasbubane:
Regarding the configlet segfault issue - I had an old version (which was segfaulting) lying in my $PATH and that was picked up instead of bin/configlet: https://github.com/exercism/haskell/blob/189f2623c0b4a91238472d092628a25193d44255/bin/ensure-readmes-are-updated.sh#L2-L5

I have updated the README and the ensure script works fine now.

exercises/resistor-colors/src/ResistorColors.hs
```diff
@@ -0,0 +1,17 @@
...
+  | Orange
+  | Yellow
+  | Green
+  | Blue
+  | Violet
+  | Grey
+  | White
+  deriving (Eq, Show, Read)
+
+value :: [String] -> Int
```
sshine:
```suggestion
value :: [Color] -> Int
```

tejasbubane: Missed this one. Fixed now.

PR #810, opened 2019-03-12
Update list-ops to 2.4.0

tejasbubane:
Change:
* Add one test case for reversing nested list: https://github.com/exercism/problem-specifications/pull/1453


tejasbubane:
@sshine Done.

PR #815, opened 2019-03-21
sublist: Replace 'Sublist' type with 'Maybe Ordering' (fixes #814)

### LINKED ISSUE #814: Sublist exercise uses custom `Sublist` type instead of `Maybe Ordering`
In the [`sublist`](https://github.com/exercism/haskell/blob/master/exercises/sublist/src/Sublist.hs) exercise, I find that `data Sublist = ...` is *very* close to the built-in `Ordering` type; it just also has the `Unequal` option. Would an idiomatic solution not be `Maybe Ordering`?

sshine:


sshine:
You have a point. I'd rather be safe than sorry.

We don't currently have a good mechanism for detecting that an exercise got hard.

PR #817, opened 2019-03-25
Upgrade word-count to v1.3.0

tejasbubane:
Added two tests:

* Multiple whitespace - https://github.com/exercism/problem-specifications/pull/1023
* Empty words - https://github.com/exercism/problem-specifications/pull/1446

### DIFF
--- a/exercises/word-count/package.yaml
+++ b/exercises/word-count/package.yaml
@@ -1,5 +1,5 @@
 name: word-count
-version: 1.2.0.6
+version: 1.3.0.7
 
 dependencies:
   - base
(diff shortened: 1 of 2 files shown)


sshine:
I note that *Multiple whitespace - exercism/problem-specifications#1023* doesn't increment the version number. Treating it as part of 1.3.0, I suppose, is the easiest. The alternative would be to bump only the 4th component (1.2.0.6 → 1.2.0.7, 1.2.0.7 → 1.3.0.8). I don't know if we should have a consistent policy on this, but here is one proposal: Do not port updates until the canonical version has incremented, and treat all changes between the canonical versions as belonging to the incremented version.

petertseng:
On the bright side, hopefully https://github.com/exercism/problem-specifications/blob/master/bin/check_versions means problem-specifications won't miss incrementing a version number anymore, so we don't have to think too hard about a policy, I hope.

PR #826, opened 2019-06-21
Add "An opinionated beginner's guide to Haskell" as learning resource

sshine:
This resource was referenced in Haskell Weekly 164, 2019-06-19. It
amends the 2018 edition, and so both of these are added as resources.

They appear to address newcomers in particular.


sshine:
https://github.com/theindigamer/not-a-blog/blob/master/opinionated-haskell-guide-2019.md
https://lexi-lambda.github.io/blog/2018/02/10/an-opinionated-guide-to-haskell-in-2018/

PR #828, opened 2019-06-28
do not merge: more warnings

petertseng:
I read the opinionated guide and decided to see what happens when we add
the flags.

https://lexi-lambda.github.io/blog/2018/02/10/an-opinionated-guide-to-haskell-in-2018/


petertseng:
huh. I guess unless I accidentally didn't apply the flags, we are good. though note that nightly says `-Wunused-matches`. If things change we may need to have stubs do `-Wno-unused-matches`

PR #830, opened 2019-08-02
Bob: Update README.md: Clarify definition of question/shouting

sshine:
This addresses exercism/problem-specifications#1516.


petertseng:
I'd suggest a little something about how it was updated, like "more clearly define yelling and questions"

PR #832, opened 2019-08-11
yacht: port

guygastineau:
[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/yacht/canonical-data.json)

I did as much as I could figure out was standard for this track.

My solution might need some refactoring, but it work and shows proof of concept.

`config.json`: I put it at level 3.  You all will know better how hard this should be rated.
I also figured I'd ask here what core exercise should unlock it instead of just guessing (I'm not that far through the Haskell track myself ;) )

Please let me know what to do to get this production ready :smile:


exercises/yacht/examples/success-standard/src/Yacht.hs
```diff
@@ -0,0 +1,69 @@
...
+yacht _ []   = error "You lose!"
+yacht Ones   dice = simpleScore 1 dice
+yacht Twos   dice = simpleScore 2 dice
+yacht Threes dice = simpleScore 3 dice
+yacht Fours  dice = simpleScore 4 dice
+yacht Fives  dice = simpleScore 5 dice
+yacht Sixes  dice = simpleScore 6 dice
+
+yacht FullHouse dice
+  | s == 2 && length g == 2 && (length . head') gs == 3 = sum dice
```
sshine: I'd like more descriptive variable names than `s` and `g`.

exercises/yacht/examples/success-standard/src/Yacht.hs
```diff
@@ -0,0 +1,69 @@
...
+  | otherwise = 0
+
+yacht Choice dice = sum dice
+
+yacht Yacht (d:ice)
+  | all (== d) ice = 50
+  | otherwise      = 0
+
+simpleScore :: Int -> [Int] -> Int
+simpleScore x dice = (sum . filter (== x)) dice
```
sshine: Why not pointfree?

```suggestion
simpleScore die = sum . filter (== die)
```

exercises/yacht/examples/success-standard/src/Yacht.hs
```diff
@@ -0,0 +1,69 @@
...
+  |  isSucc dice && maximum dice == 5 = 30
+  | otherwise = 0
+
+yacht BigStraight dice
+  | isSucc dice && maximum dice == 6 = 30
+  | otherwise = 0
+
+yacht Choice dice = sum dice
+
+yacht Yacht (d:ice)
```
sshine: `d:ice` is kind of funny, but I like better that each variable has a real name. :-)
```suggestion
yacht Yacht (die:dice)
```

exercises/yacht/examples/success-standard/src/Yacht.hs
```diff
@@ -0,0 +1,69 @@
...
+  | all (== d) ice = 50
+  | otherwise      = 0
+
+simpleScore :: Int -> [Int] -> Int
+simpleScore x dice = (sum . filter (== x)) dice
+
+gsortBySize :: ((Ord a), Eq a) => [a] -> [[a]]
+gsortBySize = sortOn (length) . group . sort
```
sshine: What's `g` in `gsort`? Maybe just write `groupSortBySize` or `sortBySize`?

```suggestion
gsortBySize = sortOn length . group . sort
```

exercises/yacht/examples/success-standard/src/Yacht.hs
```diff
@@ -0,0 +1,69 @@
...
+yacht Fours  dice = simpleScore 4 dice
+yacht Fives  dice = simpleScore 5 dice
+yacht Sixes  dice = simpleScore 6 dice
+
+yacht FullHouse dice
+  | s == 2 && length g == 2 && (length . head') gs == 3 = sum dice
+  | otherwise = 0
+  where
+    s = (length . gsortBySize) dice
+    (g:gs) = gsortBySize dice
```
sshine: What if `dice` is empty?

exercises/yacht/examples/success-standard/src/Yacht.hs
```diff
@@ -0,0 +1,69 @@
...
+
+yacht FourOfAKind dice
+  | length lg == 4 = sum lg
+  | length lg == 5 = 4 *  (head dice)
```
sshine: Unnecessary double space and parenthesis.
```suggestion
  | length lg == 5 = 4 * head dice
```

exercises/yacht/examples/success-standard/src/Yacht.hs
```diff
@@ -0,0 +1,69 @@
...
+  | otherwise = 0
+  where
+    lg = largestGroup dice
+
+yacht LittleStraight dice
+  |  isSucc dice && maximum dice == 5 = 30
```
sshine: What if `dice` is empty? `maximum` is partial.
```suggestion
  | isSucc dice && maximum dice == 5 = 30
```

exercises/yacht/examples/success-standard/src/Yacht.hs
```diff
@@ -0,0 +1,69 @@
...
+  | otherwise      = 0
+
+simpleScore :: Int -> [Int] -> Int
+simpleScore x dice = (sum . filter (== x)) dice
+
+gsortBySize :: ((Ord a), Eq a) => [a] -> [[a]]
+gsortBySize = sortOn (length) . group . sort
+
+largestGroup :: ((Ord a), Eq a) => [a] -> [a]
+largestGroup = head . reverse . gsortBySize
```
sshine: `head . reverse = last`.

What if the input is empty? `last` is partial.
```suggestion
largestGroup = last . gsortBySize
```

exercises/yacht/examples/success-standard/src/Yacht.hs
```diff
@@ -0,0 +1,69 @@
...
+
+isSucc :: [Int] -> Bool
+isSucc = all (\(x,y) -> succ x == y) . pairwise . sort
+  where pairwise = zip <*> tail
+
+head' [] = []
```
sshine: Missing signature. I'm thinking maybe you could do something else here, but I can't think of what at this moment.

exercises/yacht/examples/success-standard/src/Yacht.hs
```diff
@@ -0,0 +1,69 @@
...
+
+gsortBySize :: ((Ord a), Eq a) => [a] -> [[a]]
+gsortBySize = sortOn (length) . group . sort
+
+largestGroup :: ((Ord a), Eq a) => [a] -> [a]
+largestGroup = head . reverse . gsortBySize
+
+isSucc :: [Int] -> Bool
+isSucc = all (\(x,y) -> succ x == y) . pairwise . sort
+  where pairwise = zip <*> tail
```
sshine: Neat!

exercises/yacht/package.yaml
```diff
@@ -0,0 +1,21 @@
+name:                yacht
+version:             0.1.0.0
```
sshine: See [exercise versioning](https://github.com/exercism/haskell#exercise-versioning): We would like the [canonical version](https://github.com/exercism/problem-specifications/blob/master/exercises/yacht/canonical-data.json) followed by a serial number, so 
```suggestion
version:             1.2.0.0
```

exercises/yacht/src/Yacht.hs
```diff
@@ -0,0 +1,17 @@
...
+              | Sixes
+              | FullHouse
+              | FourOfAKind
+              | LittleStraight
+              | BigStraight
+              | Choice
+              | Yacht
+
+yacht :: Category -> [Int] -> Int
+yacht _ _ = error "unimplemented"
```
sshine: Just for conformity.
```suggestion
yacht _ _ = error "You need to implement this function."
```

exercises/yacht/stack.yaml
```diff
@@ -0,0 +1 @@
+resolver: lts-14.0
```
sshine: We may want to update the resolver across all exercises, but I think there is some advantage to having the same resolver for every exercise; in particular CI builds trigger the same compiler and library downloads.
```suggestion
resolver: lts-12.4
```

exercises/yacht/stack.yaml.lock
```diff
@@ -0,0 +1,12 @@
...
+# For more information, please see the documentation at:
+#   https://docs.haskellstack.org/en/stable/lock_files
+
+packages: []
+snapshots:
+- completed:
+    size: 523443
+    url: https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/lts/14/0.yaml
+    sha256: 283773e7120f5446d961eab35ea95c9af9c24187cc178537bd29273200a05171
+  original: lts-14.0
```
sshine: This file can be removed; it isn't present in any of the other exercises, either.

exercises/yacht/test/Tests.hs
```diff
@@ -0,0 +1,167 @@
...
+
+import Yacht (yacht, Category(..))
+
+main :: IO ()
+main = hspecWith defaultConfig {configFastFail = True} specs
+
+specs :: Spec
+specs = describe "yacht" $ for_ cases test
+  where
+
```
sshine:
```suggestion
```

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: Cool. Extra space was an accident.  I tried `$ head` and it failed (that should have been an indication that I didn't need to force it's evaluation first 😁)

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: Fair enough.  It is like natural language already gives us singular and plural forms of signifiers 🤣

I'll change it.

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: Great `point`!

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: Thanks. I am finally starting to understand applicative a bit better. Still have a long way to go 🤓

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: That is exactly why I left the signature off of this.  I was hoping we might figure out how to avoid it altogether.

exercises/yacht/stack.yaml.lock
guygastineau: Oh, I didn't realize this wasn't in the repo-wide `.gitignore`.

Consider it gone.

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: Okay.

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: Ah yes, I had been using another function I built that reduced all duplicates, but I figured I didn't need both.  

Will the composition with `length` disrupt memorization?

So are we both thinking save it, then get the length and then split it into it's parts?

Also, for dice dice is empty (I ignored it since it is untested by the problem-specifications tests): it would be enough to have:
``` Haskell
yacht _ [] = error "useful error message here"
```
Before all of the other patterns for `yacht`?

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: I made this clearer, but I there are four lines in the where.  Maybe you can give me a better way.  I'll have new work up soon ;)

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: I forgot I already used the above pattern with `"You lose!"` as the error message.  I figured that message might need to change (although it is untested, so I don't think it is that important).

Does that not handle cases where dice is empty?

Also, do you have a preference for the error message?

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: I am trying out `sortDiceByGroupSize`. It might be too long, but I would expect a function named `sortDice` to give me just the sorted dice (not the dice sorted by their group size).

Thoughts?

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: I'm not completely sure what you mean I should do with `maximum` here.

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: Oh, by `last is partial` and `maximum is partial` you mean that they will throw an exception is the list is empty?

If that is what you mean should I implement safer versions as `last'` and `maximum'`?

exercises/yacht/test/Tests.hs
guygastineau: Cool. I thought it seemed odd to have the space here, but I saw it in another test, so I went with the space ;) Fixed in this upcoming round of commits :smile:

exercises/yacht/examples/success-standard/src/Yacht.hs
sshine: I agree.

guygastineau:
Okay. New changes are up :smile:

I put all of the exercise refactoring into one commit, but all separate changes are isolated to simple, small commits.

I think I got most of what you asked for in this last push.  Please let me know what is still lacking.

Inspecific, I know neither what you wanted for `last`/`maximum` nor how you want an empty case of `dice` to be handled.

guygastineau:
Looks like something from armstrong numbers was left in my work (it was my primary example).  I'll fix this.

config.json
```diff
@@ -378,6 +378,17 @@
         "random"
       ]
     },
+    {
+      "slug": "yacht",
+      "uuid": "8e129e54-6b5c-4696-84bc-588bc1c7f0f7",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 3,
```
guygastineau: Do you think this an appropriate difficulty?

config.json
```diff
@@ -378,6 +378,17 @@
         "random"
       ]
     },
+    {
+      "slug": "yacht",
+      "uuid": "8e129e54-6b5c-4696-84bc-588bc1c7f0f7",
+      "core": false,
+      "unlocked_by": null,
```
guygastineau: Which core exercise do you think should unlock this?

guygastineau:
So, the last Travis report also had one the maintenance scripts complaining about some README's needing to be updated.

Obviously that doesn't belong in this PR, so after running `bin/configlet generate .` I only grabbed the one for this exercise.  Is there something I should do about that?  Should we do another PR where those are updated?  It should mean that `problem-specifications` got updates for those exercises since they were last generated, so I would be hesitant to regenerate the readme's without updating the tests or verifying that there haven't been breaking changes to the exercises in the `problem-specifications` repository.

sshine:
Hi there. I see that Travis CI complains about something related to Armstrong Numbers.

The CI script should not fail because of READMEs needing updates for exercises that aren't touched in the PR. I can look into this issue tomorrow. And yes, we should update the READMEs in separate PRs, but we also shouldn't be forced to do this first.

Here's a quick variation of your solution to the problem as you describe it:

```haskell
yacht :: Category -> [Int] -> Int
yacht Ones dice = valueScore 1 dice
yacht Twos dice = valueScore 2 dice
yacht Threes dice = valueScore 3 dice
yacht Fours dice = valueScore 4 dice
yacht Fives dice = valueScore 5 dice
yacht Sixes dice = valueScore 6 dice

yacht FullHouse dice = case sortedDiceGroups dice of
  [(_, 3), (_, 2)] -> sum dice
  _ -> 0

yacht FourOfAKind dice = case sortedDiceGroups dice of
  [(d, 4), _] -> d * 4
  [(d, 5)] -> d * 4
  _ -> 0

yacht LittleStraight dice
  | sort dice == [1, 2, 3, 4, 5] = 30
  | otherwise = 0

yacht BigStraight dice
  | sort dice == [2, 3, 4, 5, 6] = 30
  | otherwise = 0

yacht Choice dice = sum dice

yacht Yacht dice = case sortedDiceGroups dice of
  [(_, 5)] -> 50
  _ -> 0

valueScore :: Int -> [Int] -> Int
valueScore = (sum .) . filter . (==)

sortedDiceGroups :: [Int] -> [(Int, Int)]
sortedDiceGroups = sortOn (Down . snd) . map (head &&& length) . group . sort
```

Wrt. `yacht _ [] = 0`, I don't know what the best way to do error handling here is: We don't necessarily want to complicate the exercise by using a difficult input type (e.g. `Data.Vector.Sized`) or verbose error handling (e.g. `Maybe`). The only problem case that warrants `yacht _ [] = 0` or similar is the `yacht Yacht (die:dice)` case, since it omits `yacht Yacht []`.

If we leave the exercise open-ended to deal with invalid input (other than 5d6), I think a minimally acceptable solution should be one that does not crash but provides the value 0 on invalid input.

If we transform the exercise slightly so that `yacht` is easily total, e.g. via:

```haskell
data Die = One | Two | Three | Four | Five | Six
type Roll = (Die, Die, Die, Die, Die)
type Score = Int

yacht :: Category -> Roll -> Score
...
```

we end up making the exercise a bit ugly.

I'd probably prefer the open-ended approach with extra mentoring notes.

sshine:
> If we leave the exercise open-ended to deal with invalid input (other than 5d6), I think a minimally acceptable solution should be one that does not crash but provides the value 0 on invalid input.

Perhaps a good idea, then, is to add some track-specific negative tests for invalid input to save us the trouble of having to point it out when mentoring. @petertseng, do you have a strong opinion on such a practice?

guygastineau:
Yeah, I had left `armstrong-numbers` as the name in the `package.yaml` for the example solution.

Is fixed now.

Would you prefer that my solutions uses case?  Your example is probably more elegant than mine.  I tend to avoid case in Haskell (probably to my detriment ;) )

guygastineau:
Also, so instead of the error msg just return 0? Sounds good.

So do you not want `yacht _ [] = 0` at the top? Would you prefer it just being there for `yacht Yacht [] = 0`?

sshine:
I'd prefer it being there for `yact Yacht []` only, if you previously did a `yacht Yacht (die:dice)` pattern.

exercises/yacht/examples/success-standard/src/Yacht.hs
sshine: You could avoid having `head'` by doing concrete pattern matching:

```haskell
yacht FullHouse dice = case sortDiceByGroupSize dice of
  [[_, _], [_, _, _]] -> 50
  _ -> 0
```

I once came to the conclusion that direct pattern matching was always preferrable (in a context of pattern matching much larger ASTs in compilers), and I remember a professor saying that direct pattern matching is less abstract: If you decide to change the representation, you end up having to replace code in more places than if you always address your internal representation abstractly.

You may for example want to change the internal representation so that the largest dice group is ordered first, or so that it's a `[(Die, Count)]` rather than a `[[Die]]`. So if you want to avoid direct pattern matching and partial functions, and use abstract helper functions, maybe something like:

```haskell
yacht FullHouse dice
  | (length <$> largestDiceGroup dice) == Just 3 &&
    (length <$> nextLargestDiceGroup dice) == Just 2
    = sum dice
  | otherwise = 0

largestDice :: [Int] -> Maybe [Int]
largestDice = lastMay . sortDiceByGroupSize

nextLargestDice :: [Int] -> Maybe [Int]
nextLargestDice dice = (\xs -> initMay xs >>= lastMay) . sortDiceByGroupSize
```

The functions `lastMay` and `initMay` are from the package [`safe`](https://hackage.haskell.org/package/safe/docs/Safe.html). In this case, I'd probably just go with pattern matching because with the scope of the exercise, I don't feel like creating a layer of abstraction between `yacht` and the way I represent dice groups.

config.json
sshine: An easy answer would be Bob:
- It's quite early on
- There aren't as many exercises unlocked here, so you wouldn't need to redistribute any other exercises unlocked by Bob.

config.json
sshine: Yes, I think it's fine.

exercises/yacht/examples/success-standard/src/Yacht.hs
sshine: Just in case we're not in agreement here: I suggest that the function returns 0 if the input is invalid, rather than try to narrow down the input type to exactly 5d6, and rather than make the function return `Maybe Score` or `Either String Score`.

exercises/yacht/examples/success-standard/src/Yacht.hs
guygastineau: Thanks for the advice.  I like `fmap` there 😁

sshine:
> Would you prefer that my solutions uses case?

No, I think pattern matching directly in the function argument is very neat. :-)

Should we put it after Bob in config.json and have it merged?

guygastineau:
Yeah, I'll get to it today.  I refactored some side exercises yesterday for point free style using Functor combinators!  I learned a lot, but I didn't get any work done on my PR's :smile:

guygastineau:
Hey, so I refactored the solution.  It is not exactly what you were expecting, but I wanted to see if you liked this one for the example/if was passable with some light refactoring.

I added the `unlocked_by` as bob.  I just put it right after the other side exercise unlocked by bob.

exercises/yacht/examples/success-standard/src/Yacht.hs
```diff
@@ -0,0 +1,59 @@
...
+yacht Ones   = valueScore 1
+yacht Twos   = valueScore 2
+yacht Threes = valueScore 3
+yacht Fours  = valueScore 4
+yacht Fives  = valueScore 5
+yacht Sixes  = valueScore 6
+
+yacht FullHouse = sum
+                . join
+                . (\xs -> if length xs == 2 then xs else [])
```
guygastineau: This is the that concerns me the most, but I couldn't figure out another way to do it.

exercises/yacht/examples/success-standard/src/Yacht.hs
```diff
@@ -0,0 +1,59 @@
...
+  where verify [2, 3, 4, 5, 6] = 30
+        verify _ = 0
+
+yacht Choice = sum
+
+yacht Yacht = (* 10)
+            . length
+            . join
+            . filter((== 5) . length)
+            . diceGroups
```
guygastineau: I could make a `filterJoinGroups` function to consolidate what happens for these last 3 lines and for what happens in `FourOfAKind`, but I didn't know if that was warranted.

exercises/yacht/src/Yacht.hs
```diff
@@ -0,0 +1,17 @@
...
+              | Sixes
+              | FullHouse
+              | FourOfAKind
+              | LittleStraight
+              | BigStraight
+              | Choice
+              | Yacht
+
+yacht :: Category -> [Int] -> Int
+yacht _ _ = error "You need to implement this function."
```
petertseng: according to `grep _ exercises/*/src/*.hs` we have never used `_` in a stub. I am guessing we should provide some name instead of `_` here, unless there is a reason to deviate from the existing wisdom. However if the existing wisdom and a reason for it was documented, I seem to have lost the link to the documentation.

guygastineau: Very good point.  I didn't base that decision off of looking at the other work.

I just followed the convention of not naming variables that wouldn't be used.

I see now that we are not suppose to suggest how many variable bindings they use.
That makes plenty of sense in Haskell.  I guess I expected it not to compile if nothing was accepted.  Thank you for pointing this out :smile:

guygastineau: Pushed new changes which make the stub conform with other exercise stubs ;)

guygastineau:
> valueScore :: Int -> [Int] -> Int
> valueScore = (sum .) . filter . (==)

-----
I love that so much!
I got it reduced to the following in `ghci`
``` haskell
(.) (.) (sum .) filter (==) :: (Foldable ((->) [a]), Num [a], Eq a) => a [a]
```
Is there a name for this combinatory patter?

guygastineau:
Okay, so I sent up the last commit just to provide a hopefully more legible version of this approach.  If this style is too bizarre for the example in your opinions I will happily revert it to simple pattern matches in the function signatures.

exercises/yacht/src/Yacht.hs
```diff
@@ -0,0 +1,17 @@
...
+              | Sixes
+              | FullHouse
+              | FourOfAKind
+              | LittleStraight
+              | BigStraight
+              | Choice
+              | Yacht
+
+yacht :: Category -> [Int] -> Int
+yacht = error "You need to implement this function."
```
petertseng: I thought we should provide a name for the two function parameters, such as `yacht category dice = error "You need to implement this function.". this is based on `grep error exercises/*/src/*.hs` - they all provide names. I think I was not clear in my previous comment. I am sorry for not being clear.

petertseng: I thought we should provide a name for the two function parameters, such as `yacht category dice = error "You need to implement this function."` this is based on `grep error exercises/*/src/*.hs` - they all provide names. I think I was not clear in my previous comment. I am sorry for not being clear.

petertseng: I thought we should provide a name for the two function parameters, such as `yacht category dice = error "You need to implement this function."` this is based on `grep error exercises/*/src/*.hs` - they all provide names. I think I was not clear in my previous comment. I am sorry for not being clear.

guygastineau: No problem, thank you for being clear now 😄

I'll get it changed in the morning (EST).

Thank you for helping with this PR.

exercises/yacht/src/Yacht.hs
sshine: Our current convention is to just name the input variables meaningfully. This does provide linting warnings. One could prevent those linting errors by prefixing those meaningful names with `_`s, but this does not occur in any exercise stubs, either, and would probably cause a little confusion.

guygastineau:
Okay, thank you @sshine 😄

I had a lot of fun making this solution, but I was concerned the current approach was using too many tricks.

Given that I just started learning Haskell about 1 1/2 to 2 months ago I think I am just still in a state of surprise and awe that solutions like this can exist in programming 😁

I will change it to join filter (I was trying to save line length, but it will still be under 80), so it's all good.

I'll have a new PR after I get some coffee.

guygastineau:
Thank you! of course it was okay :smile:
I was worried that I was taking advantage of the PR process to get extra mentoring in Haskell :stuck_out_tongue_closed_eyes:  I am very glad that we both saw merit in exploring the solution space.

@sshine @petertseng Thank you both very much for advice for this PR!
I will continue to port more desired exercises in my spare time :smile:

PR #835, opened 2019-08-28
tree-building: port

guygastineau:
This doesn't point to a `canonical-data.json` file, because none exists: [problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises/tree-building)

I stole all of the tests from the `go` track.  I think it would be good to have tests that verify it can handle more levels of recursion. [go tests](https://github.com/exercism/go/blob/master/exercises/tree-building/tree_test.go)

I think I probably should have used `Either` instead of `Maybe`, but now we can have a discussion about what would be best.

This is a refactoring exercise.  I might have made the gross version too gross, but it definitely can be refactored :rofl:

I think that my `correct` solution could probably do with some refactoring too ;)

PS. I meant to send this up here last week, but I've been distracted by life itself.  It feels good to get moving on this again ;)

### DIFF
--- a/exercises/tree-building/examples/success-standard/src/TreeBuilding.hs
+++ b/exercises/tree-building/examples/success-standard/src/TreeBuilding.hs
@@ -1,7 +1,7 @@
 module TreeBuilding (newTree, Record(..), Tree(..)) where
 
 import Data.List
-import Data.Ord (comparing)
+import Data.Maybe (isNothing)
 
 type Id = Int
 type Children = [Tree]
@@ -13,58 +13,58 @@ data Tree = Leaf Id | Branch Id Children deriving (Eq, Show)
 
 newTree :: [Record] -> Maybe Tree
 newTree records
-  | cycles records = Nothing
-  | succIdCheck records == True = build . groupByParent $ records
-  | otherwise = Nothing
+  | cycles records      = Nothing
+  | succIdCheck records = build . groupByParent $ records
+  | otherwise           = Nothing
 
 -- Checks for cycling in newTree
 cycles :: [Record] -> Bool
-cycles = any id . map checkRecord
-  where checkRecord =
-          \r -> case r of
-                  (Record i Nothing)  -> if' (i /= 0) True False
-                  (Record i (Just p)) -> if' (p >= i) True False
+cycles = any checkRecord
+  where checkRecord r =
+          case r of
+            (Record i Nothing)  -> if' (i /= 0) True False
+            (Record i (Just p)) -> if' (p >= i) True False
 
 build :: [ParentGrouping] -> Maybe Tree
 build [] = Nothing
 build (x:xs)
   | not . validRoot $ x = Nothing
-  | xs == []    = Just (Leaf 0)
-  | otherwise   = Just (Branch 0 (build' xs))
+  | null xs             = Just (Leaf 0)
+  | otherwise           = Just (Branch 0 (build' xs))
 
 --  Internal build
 build' :: [ParentGrouping] -> [Tree]
 build' xs
-  | length xs == 1 = map (\xid -> Leaf xid) (snd (head xs))
+  | length xs == 1 = map Leaf (snd (head xs))
   | otherwise      = buildWithChildren xs
   where
-    buildWithChildren =
-      \(y:ys) -> map (\yid ->
-                        let children = filter ((== (Just yid)) . fst) ys
-                        in if length children > 0
-                           then (Branch yid (build' children))
-                           else Leaf yid
-                     ) (snd y)
+    buildWithChildren (y:ys) =
+      map (\yid ->
+             let children = filter ((== Just yid) . fst) ys
+             in if not (null children)
+                then Branch yid (build' children)
+                else Leaf yid
+          ) (snd y)
 
 -- Validate the Root node
 validRoot :: ParentGrouping -> Bool
-validRoot (p, ids) = p == Nothing && length ids == 1 && sum ids == 0
+validRoot (p, ids) = isNothing p && length ids == 1 && sum ids == 0
 
 groupByParent :: [Record] -> [ParentGrouping]
 groupByParent = sortOn fst
               . map (\xs -> (recordParent (head xs), map recordId xs))
               . groupBy parentsEq
-              . sortBy (comparing recordParent)
-              . sortBy (comparing recordId)
+              . sortOn recordParent
+              . sortOn recordId
 
 parentsEq :: Record -> Record -> Bool
 parentsEq rx ry = recordParent rx == recordParent ry
 
 recordParent :: Record -> Maybe Int
-recordParent = \(Record _ p) -> p
+recordParent (Record _ p) = p
 
 recordId :: Record -> Id
-recordId = \(Record i _) -> i
+recordId (Record i _) = i
 
 -- verification
 succIdCheck :: [Record] -> Bool


exercises/tree-building/examples/success-standard/src/TreeBuilding.hs
```diff
@@ -0,0 +1,77 @@
+module TreeBuilding (newTree, Record(..), Tree(..)) where
+
+import Data.List
+import Data.Maybe (isNothing)
+
+type Id = Int
+type Children = [Tree]
+type ParentGrouping = (Maybe Id, [Id])
+
+data Record = Record Id (Maybe Id) deriving (Eq, Show)
```
sshine: Since you created these two as helper functions, you might as well create them as record fields instead.

```suggestion
data Record = Record
  { recordId     :: Id
  , recordParent :: Maybe Id
  } deriving (Eq, Show)
```

exercises/tree-building/examples/success-standard/src/TreeBuilding.hs
```diff
@@ -0,0 +1,77 @@
...
+cycles = any checkRecord
+  where checkRecord r =
+          case r of
+            (Record i Nothing)  -> if' (i /= 0) True False
+            (Record i (Just p)) -> if' (p >= i) True False
+
+build :: [ParentGrouping] -> Maybe Tree
+build [] = Nothing
+build (x:xs)
+  | not . validRoot $ x = Nothing
```
sshine:
```suggestion
  | not (validRoot x) = Nothing
```

exercises/tree-building/examples/success-standard/src/TreeBuilding.hs
```diff
@@ -0,0 +1,77 @@
...
+  | length xs == 1 = map Leaf (snd (head xs))
+  | otherwise      = buildWithChildren xs
+  where
+    buildWithChildren (y:ys) =
+      map (\yid ->
+             let children = filter ((== Just yid) . fst) ys
+             in if not (null children)
+                then Branch yid (build' children)
+                else Leaf yid
+          ) (snd y)
```
sshine: This function is not exhaustive.

I might write it like:

```haskell
build' :: [ParentGrouping] -> [Tree]
build' [(_, yids)] = map Leaf yids
build' ((_, yids):pgs) = map buildWithChildren yids
  where
    buildWithChildren yid =
      let children = [ child | child <- pgs, fst child == Just yid ]
      in if null children
         then Leaf yid
         else Branch yid (build' children)
```

And I might address that it's not exhaustive by concluding that it [shouldn't be empty](http://hackage.haskell.org/package/base/docs/Data-List-NonEmpty.html).

exercises/tree-building/examples/success-standard/src/TreeBuilding.hs
```diff
@@ -0,0 +1,77 @@
...
+
+-- Validate the Root node
+validRoot :: ParentGrouping -> Bool
+validRoot (p, ids) = isNothing p && length ids == 1 && sum ids == 0
+
+groupByParent :: [Record] -> [ParentGrouping]
+groupByParent = sortOn fst
+              . map (\xs -> (recordParent (head xs), map recordId xs))
```
sshine:
```haskell
import Control.Arrow ((&&&))

(\x -> (f x, g x)) = f &&& g
```
```suggestion
              . map (recordParent . head &&& map recordId))
```

exercises/tree-building/examples/success-standard/src/TreeBuilding.hs
```diff
@@ -0,0 +1,77 @@
...
+recordParent :: Record -> Maybe Int
+recordParent (Record _ p) = p
+
+recordId :: Record -> Id
+recordId (Record i _) = i
+
+-- verification
+succIdCheck :: [Record] -> Bool
+succIdCheck = all (\(x,y) -> succ x == y) . pairwise . sort . map recordId
+  where pairwise = zip <*> tail
```
sshine: I'm not sure why this check is necessary.

Can't a database skip IDs as long as they're monotonically increasing?

exercises/tree-building/src/TreeBuilding.hs
```diff
@@ -0,0 +1,49 @@
+module TreeBuilding (newTree, Record(..), Tree(..)) where
+
+import Data.List
+import Data.Ord (comparing)
+--import Data.Maybe (fromMaybe)
```
sshine: I'm not sure an unused import is necessary in the stub.

```suggestion
```

sshine:
> I see that:
>
>     hlint ${TRAVIS_BUILD_DIR} # Run `hlint` on the entire repository.
>
> returns 1 as it's exit code if any warnings are given. That seems like desirable behavior in general. A number of the warnings I put in there intentionally, because they were redundant (to give some easy fixes during refactoring).

You're quite right, with refactoring exercises we don't want `hlint` to fail in CI because of intentional mistakes. It should be possible to add a `.hlint.yaml` containing [`ignore` directives](https://github.com/ndmitchell/hlint#ignoring-hints):

```yaml
- ignore: { within: [ TreeBuilding ] }
```

But I'm not sure how to distinguish the `TreeBuilding` of the stub with the `TreeBuilding` of the example solution. [It's not possible to specify the directory in which to ignore a module](https://github.com/ndmitchell/hlint/issues/117), so perhaps we should look into replacing a repository-wide `hlint .` with local ones. I'm not really sure, though.

exercises/tree-building/examples/success-standard/package.yaml
```diff
@@ -0,0 +1,21 @@
+name:                tree-building
+version:             1.0.0.0
```
petertseng: since there is no canonical-data.json, by what's written in https://github.com/exercism/haskell/blob/master/README.md#exercise-versioning please use `0.1.0.s` instead of `1.0.0.s` version numbers

exercises/tree-building/package.yaml
```diff
@@ -0,0 +1,21 @@
+name:                tree-building
+version:             1.0.0.0
```
petertseng: since there is no canonical-data.json, by what's written in https://github.com/exercism/haskell/blob/master/README.md#exercise-versioning please use `0.1.0.s` instead of `1.0.0.s` version numbers

exercises/tree-building/examples/success-standard/package.yaml
guygastineau: Okay, cool. I wasn't sure of the protocol.

guygastineau:
Just an update. I've been super busy recently, and I will be through Saturday.
Should get some time early next week to finish up on this project :smile:

sshine:
> It appears like we can put the ignore rules in the stub file as ANN pragmas for it to apply specifically to that file.

I'm sorry that I didn't address this in my research: Yes, we could do that, but then we would distribute the pragma to the student, which would cause HLint to be disabled on their end, which we really don't want. So ideally we should find a workaround HLint not having an `--ignore-file` or `--ignore-directory` feature, which most likely will complicate the very neat `hlint .` at the root.

The long-term benefit we get from working out this part is that we make way for handing out stubs with warnings in them for other refactoring exercises. I think having obvious errors that HLint catches is a good basis for refactoring, since it encourages the use of HLint.

guygastineau:
Okay, those are some compelling reasons to get this figured out the right way regarding `hlint` 😁

guygastineau:
I'll finally have some time for this tonight !

I looked at `Data.Tree` and it is exactly what I need.

sshine:
A last thought -- sorry for spamming you -- if we're giving a

```haskell
data Thread = Leaf Id | Branch Id Children
```

in the stub, but we'd like people to use

```haskell
import Data.Tree (Tree)
type Thread = Tree Id
```

then how can we express test cases abstractly? We would need to hide the concrete constructors (`Leaf`, `Branch` for the stub, `Node` for the example) and instead export functions for extracting the posts of a thread, i.e.

```haskell
module ForumThread (Record, Thread, fromList, thread) where

...

-- In the stub
thread :: Id -> [Thread] -> Thread
thread recordId [] = Leaf recordId
thread recordId children = Branch recordId children

-- In the example
thread :: Id -> [Thread] -> Thread
thread = Node
```

Then we can express `expected` trees in the test file using e.g.

```haskell
Just (thread 0 [ thread 1 [], thread 2 [] ])
```

regardless of internal representation.

Then it can become an objective to change the internal representation to `Data.Tree` without affecting the test suite.

guygastineau:
@sshine that is amazing!

Thank you for sharing that strategy with me 😍

exercises/tree-building/examples/success-standard/package.yaml
petertseng: upon further review of other example package.yaml, I find that we just don't have `version` in example `package.yaml`, likely so we don't need to update them in multiple places. So this one should be removed

exercises/tree-building/examples/success-standard/package.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+dependencies:
+  - base
+
+library:
+  exposed-modules: TreeBuilding
+  source-dirs: src
+  ghc-options: -Wall
+  # dependencies:
```
petertseng: these lines can (and should) be removed from example's package.yaml

sshine:
@guygastineau: I feel that I've left you a bit of a mess wrt. feedback.

I'll try and summarize my proposal of what we need to do from here:

- Rename the stub module into `ForumThread`
- Rename the main type the module exports into `ForumThread`
- Rename the main function that the module exports into `fromList`
- Create and use an abstract `thread` constructor in the stub, test and example, so students can refactor `data ForumThread = ...` into `(new)type ForumThread = Tree Id` without having to change the tests.
- Fix imports and exports in the stub, test and example to something like `Record(..), Thread, thread, fromList`.
- Make a decision about some of the negative cases, as I've outlined above.

In [this gist](https://gist.github.com/sshine/243d9cfaf6d552b8f1c253aaaa22bd76) I've updated the stub file and the tests so they conform to these proposals, and the only thing I've left commented out are the negative cases.

As for your example solution: Feel free to stop refactoring whenever you feel like it. The important part here is that it's a learning experience for you and that it behaves the same as the stub file. We can always take iterations on improving it further for your sake after merging the important parts.

guygastineau:
@sshine  thank you for your detailed work.

I will see how far I can take this with your guidance ;)

I'm sorry I've left this PR up without progress for so long. A new project for a startup is taking a lot of time from my days.

guygastineau:
Hey @sshine That is a really good Gist!  I appreciate all of your guidance here ;)

I am still swamped with real work projects for a start up right now.  I could probably implement these changes over the next day or so, but if you are willing simply to change this to the better version I would be grateful.

I am sorry that I have left this PR hanging for so long on your track.

sshine:
@guygastineau: We've had PRs open for a year, it really is no inconvenience. I'll look at it if you prefer to not continue with it.

guygastineau:
@sshine your generosity of spirit is only matched by your expertise.

Though it might take me til Monday to have it in good shape I suppose I should do it (it will help help me be a better Haskeller).  Thank you for dissuading my anxieties about longer running PR's.

exercises/tree-building/examples/success-standard/src/TreeBuilding.hs
```diff
@@ -0,0 +1,77 @@
...
+type ParentGrouping = (Maybe Id, [Id])
+
+data Record = Record Id (Maybe Id) deriving (Eq, Show)
+
+data Tree = Leaf Id | Branch Id Children deriving (Eq, Show)
+
+newTree :: [Record] -> Maybe Tree
+newTree records
+  | cycles records      = Nothing
+  | succIdCheck records = build . groupByParent $ records
```
petertseng: to aid the understandability of `succIdCheck` in this context, I suggest the function be renamed to reflect what it signifies if it were to return true. For example, if I read it correctly, it seems to be `containsConsecutiveIds` ?

exercises/tree-building/examples/success-standard/package.yaml
petertseng:
```suggestion
```

exercises/tree-building/examples/success-standard/package.yaml
```diff
@@ -0,0 +1,21 @@
...
+  # - foo       # List here the packages you
```
petertseng:
```suggestion
```

exercises/tree-building/examples/success-standard/package.yaml
```diff
@@ -0,0 +1,21 @@
...
+  # - bar       # want to use in your solution.
```
petertseng:
```suggestion
```

sshine:
> would be grateful to anyone who wanted to finish this tree building port

You already did the heavy lifting.

I'm playing with stateful tree generators for this one, so it may take a couple of days.

guygastineau:
@sshine cool. I am excited to see your work.  I imagine I will learn a lot from reading it 🤓

sshine:
I've gotten seriously frustrated at [`QuickCheck-GenT`](http://hackage.haskell.org/package/QuickCheck-GenT-0.2.0/docs/QuickCheck-GenT.html). It's old, so it seems you need to run the stateful generator using something like

    instance Arbitrary Thread where
      arbitrary = (`evalState` state) <$> runGenT (sized threadGen)

I ditched the thing when I couldn't easily call on `deriving (..., MonadState ..., MonadGen ...)` having to explicitly *lift* non-`GenT` features from `QuickCheck`'s `Gen` like `shuffle`. Meh. So I'm redoing it with [`Hedgehog.Gen`](http://hackage.haskell.org/package/hedgehog/docs/Hedgehog-Gen.html) that is naturally implemented mtl-style. Hedgehog deals with generating recursive structures in a less primitive way, so I want to learn to do that part right.

guygastineau:
Well, this sure has been open for along time ;)

I have learned a lot since I tried porting this exercise.  I think I will give it another go now.

sshine:
@guygastineau: Cool :)

I think I learned something about how to not give feedback doing this one, haha. I think if you regard all of the comments above as one big brainstorm and do the exercise without constraints, maybe it will be less critical. Feel free to decide whether you want to close this one and open a new one, or just work from here and not let the comments made so far disturb your creative process too much.

exercises/tree-building/examples/success-standard/src/TreeBuilding.hs
sshine: You may like to know that `first` and `second` actually live in two places nowadays: There is a better version in [`Data.Bifunctor`](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Bifunctor.html#v:first). A bifunctor is considerably less scary than an arrow, I think.

sshine:
@guygastineau: Upon revisiting the comment thread in this PR, I see [this comment](https://github.com/exercism/haskell/pull/835#issuecomment-528348197) containing a link to a proposed copy of the handed-out solution and a test suite that is compatible with it. You would only need to make an example solution that you like and conforms to this test suite (which was made flexible wrt. internal representation, so that this could become a part of the task).

guygastineau:
@sshine hahaha.  Well, I think your feedback was useful even though some of it lost me at the time ☺️

I got interested in this again, because I thought I might be able to use recursion schemes in the example.  I'll see if it is practical with the internal representation.  Still, my point is that your advice was good, but it was over my head at the time.  The above brainstorming session makes much more sense now that I have gone on to learn and build more Haskell projects in the wild that do actual work.  My ongoing quest to learn intuitionistic type theory is also working out (though it is hard), that and some explorations in compilers has made a lot of things clearer for me ☺️

I will need to see if I can get tools here to work for me.  Last time (for some reason I can't understand) I did not reach out about the testing tools not working.  Is stack still set up to use some sandbox directory to run all of the tests?

IDK, I will re-familiarize myself with the process of testing and contributing to this project.  If I encounter similar issues using the repo's utilities I will reach out for help with that this time.

PR #836, opened 2019-08-30
travis: reject invalid UUIDs

petertseng:
Want to prevent any invalid UUIDs from being entered.

Want to put this in configlet lint, but can't:
exercism/configlet#99
exercism/configlet#168
So it will go in individual tracks' .travis.yml for now.

It appears that the site will accept pretty much arbitrary strings as
UUIDs for now, but we want to make less work for ourselves when valid
UUIDs are required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,16 @@ script:
         exit 1
     fi
  - |
+   # Check for invalid UUIDs.
+   # can be removed once `configlet lint` gains this ability.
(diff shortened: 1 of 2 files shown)


NobbZ:
I'd want to thank you for this, and if you ever switch to github workflows, you inspired me to build a workflow file, which of course can be re-used by the tracks across exercism:

https://github.com/exercism/erlang/blob/master/.github/workflows/uuid.yml

PR #837, opened 2019-09-01
READMEs from problem specifications

petertseng:
a collection of several months' worth

These should be relatively uncontroversial.
A few other batches with some controversial ones are coming up.

do not squash


sshine:
I've merge committed this PR; this is also what you'd do, right?

petertseng:
that is indeed also what I would have done.

PR #838, opened 2019-09-01
matrix 1.1.0.6: Use 1-based indices

petertseng:
The READMEs now have 1-based indices but the tests still have 0-based.

Before these can be merged, must decide between various solutions:

1. Change the tests to be one-indexed.
2. Maintain our own README with zero-based indices. You do this by placing the README we want in `.meta/description.md`. Not recommended.
    * We cannot just use hints.md since that only allows additive changes. This is a modification of existing text, which hints.md does not support.
3. Keep both the README and tests as is, but insert a statement somewhere (hints.md, or the tests) telling students to be aware of the discrepancy.
4. Suggest to problem-specifications that no, really, they should go back to using zero-based indices.
5. Your suggestion here.

### DIFF
--- a/exercises/matrix/README.md
+++ b/exercises/matrix/README.md
@@ -14,11 +14,11 @@ So given a string with embedded newlines like:
 representing this matrix:
 
 ```text
-    0  1  2
+    1  2  3
   |---------
-0 | 9  8  7
-1 | 5  3  2
-2 | 6  6  7
+1 | 9  8  7
+2 | 5  3  2
+3 | 6  6  7
 ```
 
 your code should be able to spit out:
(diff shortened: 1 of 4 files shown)


petertseng:
My own opinion: Indifferent between 1 and 3.
I will not stop someone from doing 4, but it is not high value enough to be a movement I want to drive myself.
I do not think you can convince me to do 2.

sshine:
Matrix: I wonder what kind of solution we'd like to recommend for 1-indexing. Incidentally, [`Data.Matrix.getElem`](https://hackage.haskell.org/package/matrix/docs/Data-Matrix.html#v:getElem) uses 1-indexing.

For the current example solution using `Data.Vector` it seems that we'll be doing some translation between 0- and 1-indexing. An alternative is to use a [`Data.Array`](http://hackage.haskell.org/package/array/docs/Data-Array.html) which supports arbitrary indexing with the `Ix` type class, but we'd implement a matrix using a library that is much closer to a matrix already.

Saddle-Points: We already use `Data.Array` here (both in the stub and in the example), so making the solution work with 1-indices in a nice way should be less of a hassle.

I prefer 1, but I think you should fix the example solutions however you like.

petertseng:
Very well. It was not too hard to translate between 1-based and 0-based in the example implementation. This PR now only does matrix for reasons related to how I think things should be squashed.

PR #839, opened 2019-09-01
Connect: remove obsolete sentence from description

petertseng:
The "obsolete" sentence says that the tests won't have spaces. Now it is unspecified, but it would be unclear if the README had spaces and the tests don't have spaces.

We remove spaces in tests:
https://github.com/exercism/haskell/blob/master/exercises/connect/test/Tests.hs#L19

Before deciding, must decide:

1. Change the tests to no longer remove spaces.
2. Keep both the README and tests as is, but insert a statement somewhere (hints.md, or the tests) telling students to be aware of the discrepancy.
3. Suggest to problem-specifications that no, really, they should reinstate the stipulation about spaces not being present.
4. Your suggestion here.

original commit msg:

---

The canonical test input is now similar to what is shown in the
description, but the description says it isn't.

Fixes https://github.com/exercism/problem-specifications/issues/1435.

https://github.com/exercism/problem-specifications/pull/1513

### DIFF
--- a/exercises/connect/examples/success-standard/src/Connect.hs
+++ b/exercises/connect/examples/success-standard/src/Connect.hs
@@ -21,12 +21,12 @@ parseLines [] = error "Empty lines"
 parseLines ([]:_) = error "Empty first line"
 parseLines lines@(firstLine:_) =
   let height = length lines
-      width = length firstLine
+      width = length (filter (/= ' ') firstLine)
   in A.array ((0,0),(width - 1,height - 1)) fieldAssocs
   where fieldAssocs :: [((Int,Int),Maybe Mark)]
         fieldAssocs =
           do (l,y) <- zip lines [0 ..]
-             (c,x) <- zip l [0 ..]
+             (c,x) <- zip (filter (/= ' ') l) [0 ..]
              let f = case c of
                      'O' -> Just Nought
                      'X' -> Just Cross
--- a/exercises/connect/README.md
+++ b/exercises/connect/README.md
@@ -15,8 +15,7 @@ Your goal is to build a program that given a simple representation of a board
 computes the winner (or lack thereof). Note that all games need not be "fair".
 (For example, players may have mismatched piece counts.)
 
-The boards look like this (with spaces added for readability, which won't be in
-the representation passed to your code):
+The boards look like this:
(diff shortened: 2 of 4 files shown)


petertseng:
My own opinion: Indifferent between 1 and 2.
I will not stop someone from doing 3, but it is not high value enough to be a movement I want to drive myself.

sshine:
I have no strong opinion.

I'd go with 2 as it involves the least amount of work.

sshine:
I've spent a few minutes thinking about the exercise: I think it is unsatisfying that the README and the tests describe one problem (containing spaces) and that solutions and the actual effect of the tests (`filter (/= ' ')`) is another.

For the purpose of cleaning up, which this PR is about, I'd go with 2. But I'd create an issue and label it `hacktoberfest` for fulfilling 1, since I think the example solution must also be changed.

petertseng:
Although I suppose it would have been low impact to let the discrepancy stay (I imagine the amount of traffic to this exercise is low), I decided to not leave that discrepancy around. I executed 1 ("Change the tests to no longer remove spaces.") in the cheapest way possible (the way that requires the smallest change in the solution).

I think the commits should probably be squashed (since otherwise there exists an intermediate state where the README is out of sync with the tests), but I can live with leaving them unsquashed. Not a strong preference.

exercises/connect/examples/success-standard/src/Connect.hs
```diff
@@ -21,12 +21,12 @@ parseLines [] = error "Empty lines"
 parseLines ([]:_) = error "Empty first line"
 parseLines lines@(firstLine:_) =
   let height = length lines
-      width = length firstLine
+      width = length (filter (/= ' ') firstLine)
```
petertseng: in anticipation of the question "did you have to duplicate this", I guess not; I could have done `map (filter (/= ' ')) lines` in either `winner` or `parseLines`.

`winner` didn't seem quite right since the previous structure was that only `parseLines` cares about the particular string contents of the input, but I suppose I don't care enough to say I'd refuse to break that rule.

`parseLines` is doable, just `width` wouldn't be able to use `firstLine` anymore and would probably have to use `head lines'` or something.

Issue #600, opened 2017-10-04
Property-based-tests Diamond.


Average-user:
Here [https://github.com/exercism/problem-specifications/blob/master/exercises/diamond/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/diamond/canonical-data.json) you can see that in the diamond exercise, a property-based-test is possible. So, would be great if someone can implement a property-based-test in: [https://github.com/exercism/haskell/tree/master/exercises/diamond](https://github.com/exercism/haskell/tree/master/exercises/diamond). Meanwhile the value-based-test can do the work. 

According to  [https://github.com/exercism/problem-specifications/pull/191#issuecomment-254589087](https://github.com/exercism/problem-specifications/pull/191#issuecomment-254589087) it's not only possible to implement property-based-tests, it's preferred.

rbasso:
I think it would be great to use `QuickCheck` in this exercise!

Have you identified all the properties to be tested, @Average-user?

With that, it would be easy to change the test suite using `trinary`, `octal` or `simple-linked-list` as inspiration.

Average-user:
@rbasso  If I had to do this, I will probably make my own property tests, but thats usually not welcome.

rbasso:
> If I had to do this, I will probably make my own property tests, but thats usually not welcome.

Why?

Usually the *properties* capture the essence of the solution, so I think people tend to find the same, *"fundamental"* ones, after pondering for a while.

That said, I'm still trying to find the minimal set of properties that define a *diamond*, so take everything that I just said with a grain of salt. 😄

Average-user:
>Usually the properties capture the essence of the solution

Yeah, I understand that, it is probably because of that, that I would try to make my own tests, and not use QuickCheck. But I don't know if I want to :\

rbasso:
>> Usually the properties capture the essence of the solution

> Yeah, I understand that, it is probably because of that, that I would try to make my own tests, and not use QuickCheck. But I don't know if I want to :\

To use QuickCheck, you would need to define the properties anyway, so I guess I'm probably not understanding what you mean by *"create my own property tests"* and *"to make my own tests"* that could be different from writing properties to be tested by QuickCheck, as we did in `octal`, `trinary` and `simple-linked-list`.

Average-user:
[...]

PR #842, opened 2019-09-05
Test for Nothing case in Diamond exercise

barrymoo:
When I solved this exercise, my first solution didn't handle the `Nothing` case. This PR simply adds a test to check that `diamond '1'` returns `Nothing` to hint to the user their solution is incomplete. Any comments are appreciated.

### DIFF
--- a/exercises/diamond/test/Tests.hs
+++ b/exercises/diamond/test/Tests.hs
@@ -14,46 +14,50 @@ specs = describe "diamond" $ for_ cases test
 
     test Case{..} = it description assertion
       where
-        assertion = diamond input `shouldBe` Just expected
+        assertion = diamond input `shouldBe` expected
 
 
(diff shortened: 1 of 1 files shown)


barrymoo:
Thanks for the great comments!

> If you'd like to instead write property-based tests for this or another exercise, this could be a learning experience. I did this in #735 for another exercise.

I have experience with QuickCheck. If you think this is a better idea I am willing to give it a shot.

> I would recommend that we don't try and push a negative test to the canonical tests right now because the problem-specifications repository is temporarily only accepting bug-fixes.

Just so I understand, this statement is recommending implementing PBTs yes?

If you would like to close this, feel free. I can submit a different PR with the PBTs.

sshine:
Yes, exactly! There are some thoughts on #600 to get started.

PR #843, opened 2019-09-07
Diamond: Implement property-based tests

barrymoo:
Here is the initial implementation for property-based tests for the `diamond` exercise. For any non-letter character `diamond` should produce `Nothing`. I am very open to feedback as I would like to contribute to this project.

To do:
 
- [x] Read this PR carefully #735
- [x] Read this Issue carefully #600

### DIFF
--- a/exercises/diamond/package.yaml
+++ b/exercises/diamond/package.yaml
@@ -1,5 +1,5 @@
 name: diamond
-version: 1.1.0.4
+version: 1.1.0.5
 
 dependencies:
   - base
@@ -19,3 +19,4 @@ tests:
     dependencies:
       - diamond
       - hspec
+      - QuickCheck
--- a/exercises/diamond/test/Tests.hs
+++ b/exercises/diamond/test/Tests.hs
@@ -3,15 +3,21 @@
 import Data.Foldable     (for_)
 import Test.Hspec        (Spec, describe, it, shouldBe)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+import Test.QuickCheck   (Gen, oneof, forAll)
+import Data.List         ((\\))
+import Data.Maybe        (isNothing)
 import Diamond (diamond)
 
 main :: IO ()
 main = hspecWith defaultConfig {configFastFail = True} specs
 
 specs :: Spec
-specs = describe "diamond" $ for_ cases test
+specs = describe "diamond" $ do
+  it "non-Alpha characters should produce `Nothing`" $
+    forAll genNonAlphaChars $
+      isNothing . diamond
+  for_ cases test
   where
-
     test Case{..} = it description assertion
       where
         assertion = diamond input `shouldBe` Just expected
@@ -107,3 +113,7 @@ cases = [ Case { description = "Degenerate case with a single 'A' row"
                         "                         A                         "]
                }
         ]
+
+genNonAlphaChars :: Gen Char
+genNonAlphaChars = oneof $ pure <$> nonAlphaChars
+  where nonAlphaChars = ['\0' .. '\127'] \\ (['A' .. 'Z'] ++ ['a' .. 'z'])
--- a/exercises/diamond/test/Tests.hs
+++ b/exercises/diamond/test/Tests.hs
@@ -115,5 +115,5 @@ cases = [ Case { description = "Degenerate case with a single 'A' row"
         ]
 
 genNonAlphaChars :: Gen Char
-genNonAlphaChars = oneof $ pure <$> nonAlphaChars
+genNonAlphaChar = elements nonAlphaChars
   where nonAlphaChars = ['\0' .. '\127'] \\ (['A' .. 'Z'] ++ ['a' .. 'z'])
--- a/exercises/diamond/test/Tests.hs
+++ b/exercises/diamond/test/Tests.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE LambdaCase #-}
 
 import Data.Foldable       (for_)
 import Test.Hspec          (Spec, describe, it, shouldBe, pending)
@@ -13,8 +14,8 @@ import Test.QuickCheck     ( Gen
                            , Property
                            )
 import Data.Maybe          (isNothing, isJust, fromMaybe)
-import Data.Char           (ord, isLetter, isUpper, isPrint)
-import Control.Applicative (liftA2)
+import Data.Char           (ord, isLetter, isPrint)
+import Data.List           (isSuffixOf)
 import Diamond (diamond)
 
 main :: IO ()
@@ -42,7 +43,10 @@ specs = describe "diamond" $ do
     forAllCharDiamond checkCorrectDimensions
 
   it "rows should start and end with the same character" $
-    forAllDiamond $ all checkFirstAndLastCharEq
+    forAllDiamond $
+      let headEqualsLast ys = not (null ys) && take 1 ys `isSuffixOf` ys
+      in \case [] -> False
+               xs -> all headEqualsLast $ filter isLetter <$> xs
 
   for_ cases test
   where
@@ -189,9 +193,3 @@ shrinkNonAlphaChar c = if isPrint c
                         else printableChars
   where
     printableChars = filter isPrint ['\0' .. '\127']
-
-checkFirstAndLastCharEq :: String -> Bool
--- checkFirstAndLastCharEq xs = head letters == tail letters
-checkFirstAndLastCharEq xs = liftA2 (==) head last letters
-  where
-    letters = filter isLetter xs
--- a/exercises/diamond/test/Tests.hs
+++ b/exercises/diamond/test/Tests.hs
@@ -188,8 +188,9 @@ checkCorrectDimensions c xs = length xs == dim && all ((== dim) . length) xs
     dim = 2 * position c + 1
 
 shrinkNonAlphaChar :: Char -> String
-shrinkNonAlphaChar c = if isPrint c
-                        then takeWhile (/= c) printableChars
-                        else printableChars
+shrinkNonAlphaChar c =
+  if isPrint c
+   then takeWhile (/= c) printableChars
+   else printableChars
   where
     printableChars = filter isPrint ['\0' .. '\127']
--- a/exercises/diamond/test/Tests.hs
+++ b/exercises/diamond/test/Tests.hs
@@ -17,7 +17,7 @@ import Test.QuickCheck     ( Gen
                            , discard
                            )
 import Data.Maybe          (isNothing, isJust)
-import Data.Char           (ord, isLetter, isPrint, isSpace)
+import Data.Char           (isLetter, isPrint, isSpace)
 import Data.List           (isSuffixOf)
 import Diamond (diamond)
 
--- a/exercises/diamond/test/Tests.hs
+++ b/exercises/diamond/test/Tests.hs
@@ -3,10 +3,10 @@
 import Data.Char         (isLetter, isPrint, isSpace)
 import Data.Foldable     (for_)
 import Data.List         (isSuffixOf)
-import Data.Maybe        (isNothing, isJust)
+import Data.Maybe        (isJust, isNothing)
 import Test.Hspec        (Spec, describe, it, shouldBe)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
-import Test.QuickCheck   (arbitraryASCIIChar, counterexample, conjoin,
+import Test.QuickCheck   (arbitraryASCIIChar, conjoin, counterexample,
                           discard, elements, forAll, forAllShrink, Gen,
                           Property, suchThat, Testable, (===))
 
--- a/exercises/diamond/test/Tests.hs
+++ b/exercises/diamond/test/Tests.hs
@@ -12,6 +12,7 @@ import Test.QuickCheck     ( Gen
                            , suchThat
                            , Testable
                            , Property
+                           , (===)
                            )
 import Data.Maybe          (isNothing, isJust, fromMaybe)
 import Data.Char           (ord, isLetter, isPrint, isSpace)
@@ -34,10 +35,9 @@ specs = describe "diamond" $ do
     forAllDiamond $ odd . length
 
   it "should have equal top and bottom" $
-    forAllDiamond topEqualToBottom
-
-  it "should have the correct middle row" $
-    forAllCharDiamond checkMiddle
+    forAllDiamond $ \diamond ->
+      let halfRoundDown = length diamond `div` 2
+       in take halfRoundDown diamond === take halfRoundDown (reverse diamond)
 
   it "should have the same width and height" $
     forAllCharDiamond checkCorrectDimensions
@@ -163,24 +163,8 @@ forAllDiamond = forAllCharDiamond . const
 forAllCharDiamond :: Testable prop => (Char -> [String] -> prop) -> Property
 forAllCharDiamond p = forAll genDiamond $ uncurry p . fmap (fromMaybe [""])
 
-topLength :: [String] -> Int
-topLength = (`div` 2) . length
-
 position :: Char -> Int
 position c = ord c - ord 'A'
-
-topEqualToBottom :: [String] -> Bool
-topEqualToBottom xs = take len xs == take len (reverse xs)
-  where
-    len = topLength xs
-
-checkMiddle :: Char -> [String] -> Bool
-checkMiddle c xs = getMiddle xs == middle
-  where
-    getMiddle ys = ys !! topLength ys
-    leftSide = c : replicate (position c) ' '
-    rightSide = tail . reverse $ leftSide
-    middle = leftSide ++ rightSide
 
 checkCorrectDimensions :: Char -> [String] -> Bool
 checkCorrectDimensions c xs = length xs == dim && all ((== dim) . length) xs
--- a/exercises/diamond/test/Tests.hs
+++ b/exercises/diamond/test/Tests.hs
@@ -1,24 +1,16 @@
 {-# LANGUAGE RecordWildCards #-}
 
+import Data.Char         (isLetter, isPrint, isSpace)
 import Data.Foldable     (for_)
+import Data.List         (isSuffixOf)
+import Data.Maybe        (isNothing, isJust)
 import Test.Hspec        (Spec, describe, it, shouldBe)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
-import Test.QuickCheck   ( Gen
-                         , elements
-                         , forAll
-                         , forAllShrink
-                         , arbitraryASCIIChar
-                         , suchThat
-                         , Testable
-                         , Property
-                         , (===)
-                         , counterexample
-                         , conjoin
-                         , discard
+import Test.QuickCheck   ( arbitraryASCIIChar, counterexample, conjoin
+                         , discard, elements, forAll, forAllShrink, Gen
+                         , Property, suchThat, Testable, (===)
                          )
-import Data.Maybe        (isNothing, isJust)
-import Data.Char         (isLetter, isPrint, isSpace)
-import Data.List         (isSuffixOf)
+
 import Diamond (diamond)
 
 main :: IO ()
--- a/exercises/diamond/test/Tests.hs
+++ b/exercises/diamond/test/Tests.hs
@@ -159,7 +159,7 @@ forAllDiamond p = forAll genDiamond $ maybe discard p
 shrinkNonAlphaChar :: Char -> String
 shrinkNonAlphaChar c =
   if isPrint c
-   then takeWhile (/= c) printableChars
-   else printableChars
+  then takeWhile (/= c) printableChars
+  else printableChars
   where
     printableChars = filter isPrint ['\0' .. '\127']
--- a/exercises/diamond/test/Tests.hs
+++ b/exercises/diamond/test/Tests.hs
@@ -3,7 +3,7 @@
 import Data.Foldable     (for_)
 import Test.Hspec        (Spec, describe, it, shouldBe)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
-import Test.QuickCheck   (Gen, oneof, forAll)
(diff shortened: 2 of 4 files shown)


barrymoo:
Question: Would you prefer I convert the current unit tests to property based ones? I could easily generate a `Map Char String` which contains all of the results and implement one PBT for all characters.

exercises/diamond/test/Tests.hs
```diff
@@ -3,15 +3,21 @@
...
 main :: IO ()
 main = hspecWith defaultConfig {configFastFail = True} specs
 
 specs :: Spec
-specs = describe "diamond" $ for_ cases test
+specs = describe "diamond" $ do
+  it "non-Alpha characters should produce `Nothing`" $
+    forAll genNonAlphaChars $
+      isNothing . diamond
+  for_ cases test
```
sshine: In #735 we move the property-based tests last, but considering [Mark Seemann's blog post](https://blog.ploeh.dk/2015/01/10/diamond-kata-with-fscheck/) on exactly this:

> The no-brain approach to Example-Driven Development is to start with 'A', then 'B', and so on. Exactly as Seb Rose predicts, when you approach the problem like this, when you reach 'C', it no longer seems reasonable to hard-code the responses, but then the entire complexity of the problem hits you all at once. It's quite hard to do incremental development by going through the 'A', 'B', 'C' progression.

So I think their position first are fine, as long as the generators are listed last.

```suggestion

    for_ cases test
```

exercises/diamond/test/Tests.hs
```diff
@@ -3,15 +3,21 @@
...
 import Diamond (diamond)
 
 main :: IO ()
 main = hspecWith defaultConfig {configFastFail = True} specs
 
 specs :: Spec
-specs = describe "diamond" $ for_ cases test
+specs = describe "diamond" $ do
+  it "non-Alpha characters should produce `Nothing`" $
+    forAll genNonAlphaChars $
```
sshine: I would probably not go with a plural name; it is a generator of a single non-alpha character, so either `nonAlphaCharGen` or `genNonAlphaChar`, whichever convention is picked.

```suggestion
    forAll genNonAlphaChar $
```

exercises/diamond/test/Tests.hs
```diff
@@ -107,3 +113,7 @@ cases = [ Case { description = "Degenerate case with a single 'A' row"
                         "                         A                         "]
                }
         ]
+
+genNonAlphaChars :: Gen Char
+genNonAlphaChars = oneof $ pure <$> nonAlphaChars
```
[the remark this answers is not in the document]
barrymoo: This is a great function to know about!

barrymoo:
I added some of the basic PBTs from [Mark Seemann's blog post](https://blog.ploeh.dk/2015/01/10/diamond-kata-with-fscheck/). I can implement some of the trickier ones if you think it is necessary.

exercises/diamond/test/Tests.hs
```diff
@@ -3,15 +3,35 @@
...
+specs = describe "diamond" $ do
+  it "non-Alpha characters should produce `Nothing`" $
+    forAll genNonAlphaChar $
+      isNothing . diamond
+
+  it "Length of a diamond should be odd" $
+    forAll genAlphaChar $ odd . length . fromMaybe [""] . diamond
+
+  it "Top and bottom of a diamond should be equal" $
+    forAll genAlphaChar $ topEqualToBottom . fromMaybe [""] . diamond
```
sshine: This looks quite good.

I wonder how it might look like if we describe the actual property definition up here...

```haskell
  it "Top and bottom of a diamond should be equal" $
    forAllDiamond $ \dia ->
      let half = topLength dia
      in take half dia == take half (reverse dia)
```

exercises/diamond/test/Tests.hs
```diff
@@ -3,15 +3,32 @@
...
 import Diamond (diamond)
 
 main :: IO ()
 main = hspecWith defaultConfig {configFastFail = True} specs
 
 specs :: Spec
-specs = describe "diamond" $ for_ cases test
-  where
+specs = describe "diamond" $ do
+  it "non-Alpha characters should produce `Nothing`" $
```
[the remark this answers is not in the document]
sshine: A mindless omission: The point of the combinator's name, `it`, is to produce a sentence in the output that begins with "diamond ...". I would not demand that we respect this, as the only artifact is in a sentence like "diamond Check the middle of the diamond is correct" when a test fails. Those sentences show up on their own (indented) lines otherwise.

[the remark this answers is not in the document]
barrymoo: I didn't know (or realize) this. I will try to rewrite the descriptions with this in mind.

exercises/diamond/test/Tests.hs
```diff
@@ -3,15 +3,35 @@
...
+    forAll genNonAlphaChar $
+      isNothing . diamond
+
+  it "Length of a diamond should be odd" $
+    forAll genAlphaChar $ odd . length . fromMaybe [""] . diamond
```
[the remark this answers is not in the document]
barrymoo: This is a fantastic suggestion.

exercises/diamond/test/Tests.hs
barrymoo: For the simpler PBTs, I think this style would be a bit less burdensome for future readers.

barrymoo:
@sshine Thanks for the awesome feedback! I will work on this Thursday (7/12/19)

exercises/diamond/test/Tests.hs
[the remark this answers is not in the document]
sshine: Hspec is inspired by [RSpec](https://rspec.info/about/) which calls itself "behavior-driven development".

exercises/diamond/test/Tests.hs
sshine: Since we're using equality here, you can also consider [`===`](https://hackage.haskell.org/package/QuickCheck-2.13.2/docs/Test-QuickCheck.html#v:-61--61--61-). If it doesn't print the counter-example nicely, [`counterexample`](https://hackage.haskell.org/package/QuickCheck-2.13.2/docs/Test-QuickCheck.html#v:counterexample) can be used to pretty-print the upper and lower part for better visual feedback.

exercises/diamond/test/Tests.hs
```diff
@@ -107,3 +134,40 @@ cases = [ Case { description = "Degenerate case with a single 'A' row"
                         "                         A                         "]
                }
         ]
+
+genNonAlphaChar :: Gen Char
+genNonAlphaChar = arbitraryASCIIChar `suchThat` (not . isLetter)
+
+genAlphaChar :: Gen Char
+genAlphaChar = arbitraryASCIIChar `suchThat` isUpper
```
sshine: As I wrote in [this comment](https://github.com/exercism/haskell/pull/843#discussion_r321981808), ``gen `suchThat` predicate`` has a drawback when `predicate` accepts only a few of the values that `gen` generates, since at best it will be inefficient at generating, and at worst it will give up. You can see this in [the source code for `suchThat`](http://hackage.haskell.org/package/QuickCheck-2.13.2/docs/src/Test.QuickCheck.Gen.html#suchThat):

```haskell
-- | Generates a value that satisfies a predicate.
suchThat :: Gen a -> (a -> Bool) -> Gen a
gen `suchThat` p =
  do mx <- gen `suchThatMaybe` p
     case mx of
       Just x  -> return x
       Nothing -> sized (\n -> resize (n+1) (gen `suchThat` p))

-- | Tries to generate a value that satisfies a predicate.
-- If it fails to do so after enough attempts, returns @Nothing@.
suchThatMaybe :: Gen a -> (a -> Bool) -> Gen (Maybe a)
gen `suchThatMaybe` p = sized (\n -> try n (2*n))
 where
  try m n
    | m > n = return Nothing
    | otherwise = do
        x <- resize m gen
        if p x then return (Just x) else try (m+1) n
```

Here is a micro-benchmark using [Criterion](http://www.serpentine.com/criterion/tutorial.html):

```haskell
module Main where

import Test.QuickCheck
import Criterion
import Criterion.Main
import Data.Char (isLetter, isUpper)
import Data.List ((\\))

main :: IO ()
main = defaultMain
  [ bgroup "genUpper"
      [ bench "suchThat"        $ genUpperSuchThat       `nfQC` isUpper
      , bench "elements"        $ genUpperElements       `nfQC` isUpper
      , bench "elements/filter" $ genUpperElementsFilter `nfQC` isUpper
      ]
  , bgroup "genNonAlpha"
      [ bench "suchThat"        $ genNonLetterSuchThat       `nfQC` (not . isUpper)
      , bench "elements"        $ genNonLetterElements       `nfQC` (not . isUpper)
      , bench "elements/filter" $ genNonLetterElementsFilter `nfQC` (not . isUpper)
      ]
  ]

nfQC :: Gen Char -> (Char -> Bool) -> Benchmarkable
nfQC gen = nfIO . quickCheckWith args . forAll gen
  where
    args = stdArgs { chatty = False }

genUpperSuchThat, genUpperElements, genUpperElementsFilter :: Gen Char
genUpperSuchThat = arbitraryASCIIChar `suchThat` isUpper
genUpperElements = elements ['A'..'Z']
genUpperElementsFilter = elements . filter isUpper $ ['\0'..'\127']

genNonLetterSuchThat, genNonLetterElements, genNonLetterElementsFilter :: Gen Char
genNonLetterSuchThat = arbitraryASCIIChar `suchThat` (not . isLetter)
genNonLetterElements = elements (['\0'..'\127'] \\ (['A'..'Z'] ++ ['a'..'z']))
genNonLetterElementsFilter = elements . filter (not . isLetter) $ ['\0'..'\127']
```

and compiling and running it gives:

```
$ stack ghc -- -O2 AsciiCharBenchmark.hs --output gen.html
[1 of 1] Compiling Main             ( AsciiCharBenchmark.hs, AsciiCharBenchmark.o )
Linking AsciiCharBenchmark ...

$ ./AsciiCharBenchmark 
benchmarking genUpper/suchThat
time                 213.5 μs   (212.0 μs .. 216.1 μs)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 218.1 μs   (215.8 μs .. 221.8 μs)
std dev              9.044 μs   (6.284 μs .. 13.24 μs)
variance introduced by outliers: 39% (moderately inflated)

benchmarking genUpper/elements
time                 89.69 μs   (88.31 μs .. 91.90 μs)
                     0.994 R²   (0.987 R² .. 1.000 R²)
mean                 89.43 μs   (88.29 μs .. 92.76 μs)
std dev              5.595 μs   (1.949 μs .. 10.41 μs)
variance introduced by outliers: 63% (severely inflated)

benchmarking genUpper/elements/filter
time                 89.38 μs   (89.07 μs .. 89.73 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 89.09 μs   (88.92 μs .. 89.31 μs)
std dev              609.7 ns   (441.7 ns .. 825.1 ns)

benchmarking genNonAlpha/suchThat
time                 124.2 μs   (121.9 μs .. 127.2 μs)
                     0.998 R²   (0.997 R² .. 1.000 R²)
mean                 122.4 μs   (121.8 μs .. 123.7 μs)
std dev              2.859 μs   (1.572 μs .. 5.033 μs)
variance introduced by outliers: 19% (moderately inflated)

benchmarking genNonAlpha/elements
time                 95.06 μs   (94.96 μs .. 95.17 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 95.06 μs   (94.96 μs .. 95.21 μs)
std dev              404.3 ns   (251.1 ns .. 626.1 ns)

benchmarking genNonAlpha/elements/filter
time                 95.00 μs   (94.85 μs .. 95.15 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 95.22 μs   (95.07 μs .. 95.56 μs)
std dev              707.7 ns   (442.0 ns .. 1.240 μs)
```

With HTML output [here](https://simonshine.dk/gen.html) (the colors seem a bit off).

As you can see, `suchThat` is less efficient when only 26/128 generated elements are accepted by its right-hand side predicate, as when (128 - 2*26)/128 = 76/128 generated elements are accepted. Generating from a positive set is always faster. So while it doesn't really matter here, the point I wanted to make was that you should be careful with `suchThat`.

I said something wrong to you yesterday, being that `suchThat` would eventually make the test give up. But as evidenced from:

```haskell
     case mx of
       Just x  -> return x
       Nothing -> sized (\n -> resize (n+1) (gen `suchThat` p))
```

it actually keeps growing the generator's size parameter.

barrymoo: Considering this, I decided that the `genNonAlphaChar` is more declarative using `suchThat`. However, `genAlphaChar` is cleaner using `elements`. If we consider the performance data `elements` is more compelling for the latter case too.

barrymoo:
@sshine I went through all of your comments and I think this PR is in much better shape. Thanks again for the great comments!

sshine:
@barrymoo: Thanks for your latest pushes! I'll look at this PR this weekend.

exercises/diamond/test/Tests.hs
```diff
@@ -107,3 +142,57 @@ cases = [ Case { description = "Degenerate case with a single 'A' row"
...
+  where
+    len = topLength xs
+
+checkMiddle :: Char -> [String] -> Bool
+checkMiddle c xs = getMiddle xs == middle
+  where
+    getMiddle ys = ys !! topLength ys
+    leftSide = c : replicate (position c) ' '
+    rightSide = tail . reverse $ leftSide
+    middle = leftSide ++ rightSide
```
sshine: Along with all the other predicates at the bottom of the file, I will recommend that `checkMiddle` is also moved up.

I will also say the way this property is expressed is very close to being a re-implementation.

An exaggerated example of what I mean is:

Imagine that you property-test `isLeapYear` in the following way:

```haskell
newtype Year = Year Int deriving (Eq, Show)
instance Arbitrary Year where
  arbitrary = Year <$> (genYear :: Gen Int)

prop_isLeapYear :: Year -> Bool
prop_isLeapYear (Year year)
  | year `rem` 400 = isLeapYear year
  | year `rem` 100 = not (isLeapYear year)
  | year `rem`   4 = isLeapYear year
  | otherwise      = False 
```

The property then becomes "the student solution is equivalent to an embedded reference solution". Having a competing implementation that you can check against is indeed a very effective way to verify that your implementation is correct. But embedding it in a test suite for an exercise is not so ideal.

Having re-read [Mark Seemann's blog post](https://blog.ploeh.dk/2015/01/10/diamond-kata-with-fscheck/), I can't really find any properties that he tests that are equivalent to checking the middle part, so perhaps it is fine to skip on this.

exercises/diamond/package.yaml
```diff
@@ -19,3 +19,4 @@ tests:
     dependencies:
       - diamond
       - hspec
+      - QuickCheck
```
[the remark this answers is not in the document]
barrymoo: Should be settled via 642e1e0

exercises/diamond/test/Tests.hs
```diff
@@ -1,17 +1,52 @@
...
+import Test.QuickCheck     ( Gen
+                           , elements
+                           , forAll
+                           , forAllShrink
+                           , arbitraryASCIIChar
+                           , suchThat
+                           , Testable
+                           , Property
+                           )
+import Data.List           (partition)
```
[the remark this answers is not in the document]
barrymoo: I think this was an artifact from a previous discussion we had. `filter` obviously makes more sense.

exercises/diamond/test/Tests.hs
```diff
@@ -107,3 +142,57 @@ cases = [ Case { description = "Degenerate case with a single 'A' row"
...
+shrinkNonAlphaChar :: Char -> String
+shrinkNonAlphaChar c = if isPrint c
+                        then takeWhile (/= c) printableChars
+                        else printableChars
+  where
+    (printableChars, _) = partition isPrint ['\0' .. '\127']
+
+checkFirstAndLastCharEq :: String -> Bool
+-- checkFirstAndLastCharEq xs = head letters == tail letters
+checkFirstAndLastCharEq xs = liftA2 (==) head last letters
```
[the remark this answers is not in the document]
barrymoo: Resolved by 1e9ee3c97eeeb6d4312351fdc6621acde1e8ea8a

exercises/diamond/test/Tests.hs
```diff
@@ -1,17 +1,52 @@
...
+    forAllDiamond topEqualToBottom
+
+  it "should have the correct middle row" $
+    forAllCharDiamond checkMiddle
+
+  it "should have the same width and height" $
+    forAllCharDiamond checkCorrectDimensions
+
+  it "rows should start and end with the same character" $
+    forAllDiamond $ all checkFirstAndLastCharEq
```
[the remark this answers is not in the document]
barrymoo: See 4554d266a7ffe4d4c84943c1ba71aa9b001fd40c

exercises/diamond/test/Tests.hs
```diff
@@ -1,17 +1,52 @@
...
+      isNothing . diamond
+
+  it "should produce a value for an alpha character" $
+    forAll genAlphaChar $ isJust . diamond
+
+  it "should have an odd number of rows" $
+    forAllDiamond $ odd . length
+
+  it "should have equal top and bottom" $
+    forAllDiamond topEqualToBottom
```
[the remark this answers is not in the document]
barrymoo: See f7b40857f6a8d11b081efca432b4cfb405201095

exercises/diamond/test/Tests.hs
barrymoo: I am fine removing this, see f7b40857f6a8d11b081efca432b4cfb405201095

exercises/diamond/test/Tests.hs
```diff
@@ -1,17 +1,52 @@
...
+
+  it "should have the correct middle row" $
+    forAllCharDiamond checkMiddle
+
+  it "should have the same width and height" $
+    forAllCharDiamond checkCorrectDimensions
```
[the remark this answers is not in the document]
barrymoo: See eb61cce77a024f868f7bd58d78c5ab3524b8c44d

exercises/diamond/test/Tests.hs
```diff
@@ -1,17 +1,62 @@
...
+  it "should produce a value for an alpha character" $
+    forAll genAlphaChar $ isJust . diamond
+
+  it "should have an odd number of rows" $
+    forAllDiamond $ odd . length
+
+  it "should have equal top and bottom" $
+    forAllDiamond $ \rows ->
+      let halfRoundDown = length rows `div` 2
+       in take halfRoundDown rows === take halfRoundDown (reverse rows)
```
petertseng: mark for single

exercises/diamond/test/Tests.hs
```diff
@@ -1,17 +1,62 @@
...
+            (concat [ "The length of row with index "
+                    , show idx
+                    , " is not equal to the height" ])
+            (length row === length rows)
+      in conjoin $ zipWith sameHeightWidth [0 :: Int ..] rows
+
+  it "rows should start and end with the same letter" $
+    forAllDiamond $
+      let headEqualsLast row = not (null row) && take 1 row `isSuffixOf` row
+       in (&&) <$> not . null <*> all (headEqualsLast . filter (not . isSpace))
```
petertseng: mark for single

exercises/diamond/test/Tests.hs
```diff
@@ -107,3 +152,23 @@ cases = [ Case { description = "Degenerate case with a single 'A' row"
...
+genDiamond :: Gen (Maybe [String])
+genDiamond = diamond <$> genAlphaChar
+
+forAllDiamond :: Testable prop => ([String] -> prop) -> Property
+forAllDiamond p = forAll genDiamond $ maybe discard p
+
+shrinkNonAlphaChar :: Char -> String
+shrinkNonAlphaChar c =
+  if isPrint c
+   then takeWhile (/= c) printableChars
```
petertseng: mark for single

exercises/diamond/test/Tests.hs
```diff
@@ -107,3 +152,23 @@ cases = [ Case { description = "Degenerate case with a single 'A' row"
...
+   else printableChars
```
petertseng: mark for single

exercises/diamond/test/Tests.hs
```diff
@@ -1,17 +1,62 @@
...
-import Test.Hspec        (Spec, describe, it, shouldBe)
-import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+import Data.Foldable       (for_)
+import Test.Hspec          (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner   (configFastFail, defaultConfig, hspecWith)
+import Test.QuickCheck     ( Gen
+                           , elements
+                           , forAll
+                           , forAllShrink
+                           , arbitraryASCIIChar
```
petertseng: order? arbitraryASCIIChar < forAllShrink

exercises/diamond/test/Tests.hs
```diff
@@ -1,17 +1,62 @@
...
+                           , suchThat
+                           , Testable
+                           , Property
+                           , (===)
+                           , counterexample
+                           , conjoin
+                           , discard
+                           )
+import Data.Maybe          (isNothing, isJust)
```
petertseng: order? isJust < isNothing

exercises/diamond/test/Tests.hs
```diff
@@ -1,17 +1,62 @@
 {-# LANGUAGE RecordWildCards #-}
 
-import Data.Foldable     (for_)
-import Test.Hspec        (Spec, describe, it, shouldBe)
-import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+import Data.Foldable       (for_)
+import Test.Hspec          (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner   (configFastFail, defaultConfig, hspecWith)
```
petertseng: it doesn't seem necessary to increase the indentation. no `import` being used here is longer than `Test.Hspec.Runner`, is it?

sshine:
Excellent work, @barrymoo. Thanks a lot for your contribution.

Catch you on the twitch stream. :-)

Issue #470, opened 2017-01-15
When do we want to update our Stack snapshot?


petertseng:
In #435 we update to lts-7.9.

Is there a rule for how often we want to do this? For example, lts-7.15 is out now.

Or is it just "we update whenever someone feels like it"?

Let's see when a few versions were released since 7.9:

* 2016-11-14
* 2016-11-22
* 2016-11-28
* 2016-12-04
* 2016-12-12
* 2016-12-18
* 2017-01-08

Okay, so that means (normally) weekly. I don't think we particularly want to update every week. So, do we want to adopt some sort of thing like "let's update every month", or do we want it to be looser, like "we'll update if something is broken, otherwise we won't"?

rbasso:
Any snapshot update generates additional downloads and compilation to the users, so I always avoided updating too frequently.

- Minor number updates usually require small downloads and some additional compilation, but they are generally safe. I guess they are supposed to only add new packages and make non-breaking updates to the existing packages (changes to the patch number).

- Major number updates require a lot of downloads and compilation, in some cases forcing download of a new GCC. Most of the time they reduce the number of available package.

I believe we can periodically update minor numbers, to allow more packages and also receive some non-breaking updates, but the major number bumps should be decided in a by-case basis.

We may also need to clear and rebuild the Travis-CI cache from time to time because of the updates, to check that everything builds from nothing - as expected - and to reduce the cache size, that keeps increasing as we build multiple snapshots. Usually I only do that when we change `.travis.yml`.

petertseng:
> I believe we can periodically update minor numbers, to allow more packages and also receive some non-breaking updates, but the major number bumps should be decided in a by-case basis.

Makes sense. How often do you think minor updates should be done? A month? Two months? Longer? Is there even a need for a set schedule?

> We may also need to clear and rebuild the Travis-CI cache from time to time because of the updates, to check that everything builds from nothing - as expected - and to reduce the cache size, that keeps increasing as we build multiple snapshots. 

Hmm... seems like it would be good to have these maintenance tasks written down somewhere so that we can be reminded of them (and/or any future maintainers)

rbasso:
> Is there even a need for a set schedule?

For minor updates, anytime someone feels it is too old seems good enough IMHO.This way is is up the one opening the PR if it is a week or a few months. :)

> Hmm... seems like it would be good to have these maintenance tasks written down somewhere so that we can be reminded of them (and/or any future maintainers)

[...]

PR #846, opened 2019-09-26
stack.yaml: lts-12.4 -> lts-14.7

petertseng:
https://www.stackage.org/diff/lts-12.4/lts-14.7

Last update 14 months ago
https://github.com/exercism/haskell/pull/697

I think it would be interesting to get on GHC 8.6.x.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ cache:
     - $HOME/.foldercache # Per exercise `.stack-work` cache.
 
 env:
- - RESOLVER="lts-12.4" CURRENT="YES"  # Equal to each stack.yaml.
+ - RESOLVER="lts-14.7" CURRENT="YES"  # Equal to each stack.yaml.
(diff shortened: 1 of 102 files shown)


petertseng:
I see that this subjects every exercise to the README check, and I'm not interested in fixing resistor-colour-duo's README since a change to the tests will be needed. I will take an appropriate action at some point, maybe after work.

petertseng:
Huh, this one went a lot smoother than 12.4 did.

sshine:
I've taken the liberty to assume so.

petertseng:
I do think the merge commit was the better of the choices.

As you can see, the stack version is updated very irregularly, "whenever someone feels like it". It was discussed briefly in https://github.com/exercism/haskell/issues/470 but the conclusion there reached was simply "left to the choice of whoever wants to open the PR", which happened to be me this time.

PR #850, opened 2019-10-01
Bank Account: extend hint with extra help

### LINKED ISSUE #799: Bank Account: Hard to guess approach to exercise
Going over the [Bank Account](https://github.com/exercism/haskell/tree/master/exercises/bank-account) exercise, there is only one hint in the exercise text, "from multiple threads/processes (terminology depends on your programming language)".

The [example solution](https://github.com/exercism/haskell/blob/master/exercises/bank-account/examples/success-standard/src/BankAccount.hs) uses `TVar`s from `Control.Concurrent.STM.TVar`, but if one googles threading and Haskell, there are plenty of other things from `Control.Concurrent` one may run into and try to make work.

- [x] See if there are any tutorials on concurrency in Haskell that apply to this exercise.
- [ ] Apply a *track-specific hint* in `bank-account/.meta/hints.md` and/or refine the stub.
  (An example of a hint for another exercise: [Bob](https://github.com/exercism/haskell/blob/master/exercises/bob/.meta/hints.md).)

ppartarr:
The [Concurrent Haskell](https://en.wikipedia.org/wiki/Concurrent_Haskell) wikipedia page has an in-depth example of  `Control.Concurrent.STM` and should guide the user in the right direction.

The second point links to the [STM](https://hackage.haskell.org/package/stm) package to point the user to `TVar`. Maybe this is too explicit?

Closes #799

### DIFF
--- a/exercises/bank-account/.meta/hints.md
+++ b/exercises/bank-account/.meta/hints.md
@@ -14,3 +14,8 @@ The initial balance of the bank account should be 0.
 You will find a dummy data declaration and type signatures already in place,
 but it is up to you to define the functions and create a meaningful data type,
 newtype or type synonym.
+
+If you need help, here are some additional resources:
+
+- Read about [concurrency](https://en.wikipedia.org/wiki/Concurrent_Haskell) in Haskell.
+- Look into the [Software Transactional Memory](https://hackage.haskell.org/package/stm) package and its Transactional Variables.
--- a/exercises/bank-account/README.md
+++ b/exercises/bank-account/README.md
@@ -43,6 +43,11 @@ You will find a dummy data declaration and type signatures already in place,
 but it is up to you to define the functions and create a meaningful data type,
 newtype or type synonym.
 
+If you need help, here are some additional resources:
+
+- Read about [concurrency](https://en.wikipedia.org/wiki/Concurrent_Haskell) in Haskell.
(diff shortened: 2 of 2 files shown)


sshine:
Disclaimer: I didn't solve the exercise, and I never wrote concurrent Haskell code.

Wow, a Wikipedia article that works like a Haskell tutorial.

And the example is even bank account-ish.

@petertseng: Do you find that this is too helpful?

I wonder if either we want to add `stm` to package.yaml or indicate that we'd like the student to do that. Because of a bug reported in exercism/exercism#4753, making a change to package.yaml will not propagate to the copy the student downloads until the test file is altered.

ppartarr:
The build was failing because I didn't run `configlet` in my initial commit so the push force was to to fix this

petertseng:
EBWODP:

> If the README should try to prevent a fuller view of the options,

I meant If the README should try to **provide** a fuller view of the options,

ppartarr:
I decided to go ahead with the `MVar` suggestion since it is usually the preferred approach to writing concurrent Haskell. The README mentions to "watch out for concurrent transactions` and using `STM` is a foolproof solution whereas `IORef` might fall short if the student had to build on the solution of this exercise to add a second account, as you mention.

It might be worth introducing the concept of `IORef` and its limitations with regards to concurrency in another exercise but that is probably out of the scope of this PR.

sshine:
Thanks for providing this evaluation. I'll merge this.

Issue #841, opened 2019-09-04
Port test suite so it supports Data.Text


sshine:
In #780, exercises that might be applicable for teaching the `Data.Text` module were found. In some of those cases, the test suite was extended with `OverloadedStrings` so that both `String` and `Text` solutions work, and an example solution was provided. For a notable example of this, one of the early exercises, [Bob](https://github.com/exercism/haskell/tree/master/exercises/bob), had [an example](https://github.com/exercism/haskell/tree/master/exercises/bob/examples/success-text) added, its [test suite](https://github.com/exercism/haskell/blob/master/exercises/bob/test/Tests.hs) updated, and had [the following hint](https://github.com/exercism/haskell/blob/master/exercises/bob/.meta/hints.md) added.

This issue contains the remaining exercises for which `Data.Text` may be applicable. Contributors are free to pick an exercise they're comfortable with, **(1)** make an example solution, **(2)** make the test suite compatible with multiple string types, **(3)** add a hint, and **optionally (4)** make `Data.Text` the default in the stub file.

List of exercises that don't yet support `Data.Text` but could:
- [ ] `rotational-cipher`: Easy.
- [ ] `raindrops`: Easy.
- [ ] `diamond`: Easy.
- [ ] `twelve-days`: Easy.
- [ ] `atbash-cipher`: Easy.
- [ ] `run-length-encoding`: Doable.
- [ ] `ocr-numbers`: Doable.
- [ ] `rail-fence-cipher`: Doable.
- [ ] `luhn`: Easy, but hardly relevant as the exercise mainly consists of working on integers.
- [x] `series`: Easy. Test suite already contains `OverloadedStrings` and already has `ByteString -> Seq (Seq a)`. So a `Data.Text` example is probably not necessary unless the stub should encourage something other than `String` by default.

I've looked at the example solution(s) for each of these and made a quick estimate where "Easy" means that it would be quick to translate this to `Data.Text` by finding equivalent combinators, and "Doable" means that the contributor probably needs to learn (part of) the algorithm in the exercise to port it. The estimation was made very quickly and subjectively, so your mileage may vary.

barrymoo:
I can add `Data.Text` support to `diamond`. I am working my way through the exercises and I don't want to contibute to ones I haven't completed. That being said, the next closest exercises are `raindrops` and `twelve-days`.

sshine:
[...]

PR #855, opened 2019-10-04
Initial diamond text support in test suite

barrymoo:
Add `Data.Text` support for `diamond` from #841

@sshine Should I just modify the `README.md` or add a `.meta/hints.md` similar to https://github.com/exercism/haskell/blob/master/exercises/bob/.meta/hints.md?

### DIFF
--- a/exercises/diamond/.meta/hints.md
+++ b/exercises/diamond/.meta/hints.md
@@ -29,7 +29,7 @@ import           Data.Text (Text)
 - You can now write e.g. `diamond :: Char -> Maybe [Text]` and refer to
   `Data.Text` combinators as e.g. `T.pack`,
 - Look up the documentation for
-  [`Data.Text`](https://hackage.haskell.org/package/text-1.2.3.1/docs/Data-Text.html),
+  [`Data.Text`](https://hackage.haskell.org/package/text/docs/Data-Text.html),
 - You can then replace all occurrences of `String` with `Text` in Diamond.hs:
 
 ```haskell
--- a/exercises/diamond/.meta/hints.md
+++ b/exercises/diamond/.meta/hints.md
@@ -1,0 +1,39 @@
+## Hints
+
+You need to implement the `diamond` function which prints a diamond starting at
+`A` with the given character at its widest points. You can use the provided
+signature if you are unsure about the types, but don't let it restrict your
+creativity:
+
+```haskell
+diamond :: Char -> Maybe [String]
+```
+
+This exercise works with textual data. For historical reasons, Haskell's
+`String` type is synonymous with `[Char]`, a list of characters. For more
+efficient handling of textual data, the `Text` type can be used.
+
+As an optional extension to this exercise, you can
+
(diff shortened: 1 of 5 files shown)


sshine:
@barrymoo: What would you prefer?

petertseng:
By the way, just so my position is known:

I agree this is unfortunate. I'm tentatively open to any of the proposed solutions, and don't have any to add at this moment.

I think we're waiting for something to be added to the README, yes?

sshine:
Yes, I'm sorry, I completely forgot that. @barrymoo:

- [ ] Decide which option you prefer to go with
- [ ] Add a `.meta/hints.md` that resembles e.g. #853
- [ ] At the track repo's root, `configlet generate . --only diamond`
- [ ] Commit and push both of the updated `.meta/hints.md` and `README.md`

This will merge the hint with README.md in such a way that passes CI.

barrymoo:
@sshine

> Decide which option you prefer to go with

I really don't think this is a decision I should make. I am split between `string-conversions` and `Text`-only. Many folks just starting out (including me) might prefer to use `String` because they are comfortable with it. On the other hand, learning and using `Text` is likely what you will encounter in the "real" world.

barrymoo:
The `text` package actually depends on `bytestring` anyway (http://hackage.haskell.org/package/text). I think this makes me lean towards the `string-conversions` package.

sshine:
@barrymoo: I'd go with `string-conversions` then.

barrymoo:
I'll update this PR tonight with `string-conversions` and metadata.

sshine:
As a side note unrelated to the completion of this pull request, `string-conversions` has two interfaces, the other one of which is in the module `Data.String.Conversions.Monomorphic` in which there's a function [`toString`](https://hackage.haskell.org/package/string-conversions-0.4.0.1/docs/Data-String-Conversions-Monomorphic.html#v:toString)

```haskell
toString :: ConvertibleStrings a String => a -> String
```

that fixates the output type. Some people (such as the author of the [`string-fromto`](https://github.com/clovyr/string-fromto#motivation) package) go even further and eliminate the use of type classes here. I've not personally experienced annoyance at type-inference ambiguity when converting strings, but I probably haven't written enough Haskell.

A comparable example you may have run into is when you write `x ^ 10` and GHC will say:

```
$ stack ghc -- -Weverything Pow.hs
...
Pow.hs:5:15: warning: [-Wtype-defaults]
    • Defaulting the following constraints to type ‘Integer’
        (Integral b0) arising from a use of ‘^’ at Pow.hs:5:15-20
        (Num b0) arising from the literal ‘10’ at Pow.hs:5:19-20
    • In the first argument of ‘print’, namely ‘(2 ^ 10)’
      In the expression: print (2 ^ 10)
      In an equation for ‘main’: main = print (2 ^ 10)
  |
5 | main = print (2 ^ 10)
  |               ^^^^^^
```

because "*what `10`???*"

```
$ stack ghci
λ> :t 10
10 :: Num p => p
```

PR #869, opened 2019-10-14
resistor-color-trio: Add exercise (with extensions)

sshine:
This adds the exercise from exercism/problem-specifications#1551.

In addition to the canonical exercise, this implementation extends:
 - Add a "zero" test case (Black, Black, Black).

 - Add a `Data.Text` hint in `.meta/hints.md`.

 - Export both a `label` function *and* an `ohms` function: While the
   exercise is mainly about printing a formatted string, this nudges
   the student towards separating print and calculate.

 - For totality, since `Resistor` allows expressing resistors in the
   megaohm and gigaohm ranges, extend the exercise so that the correct
   unit beyond kiloohms are returned.

 - Extended unit tests are added, and coverage is described as such:

   Canonical tests are marked with 'x'.
   Extended tests are marked with 'y'.
   Optional (commented out) tests are marked with 'z'.

   ```
   Color       1, 2      3
   -------------------------
   Black       xyyyyy    xyy
   Brown       xyz       x
   Red         xyz       xz
   Orange      xxz       x
   Yellow      xyz       x
   Green       xyz       yz
   Blue        xyz       y
   Violet      xy        y
   Grey        xy        yz
   White       yy        y
   ```

   The over-representation of Black is partly caused by the fact that
   canonical and extended tests avoid edge cases like "4.7 kiloohms".
   There is some didactial point in incremental improvements, and as
   such, canonical and extended tests partially validates solutions.

   Optional (commented out) tests make the exercise considerably more
   difficult, so they are accompanied with a disclaimer that students
   may choose freely to include them or not.

   This way, students who are oblivious to the edge cases will not be
   aware of them, and students who are not (either by thinking or by
   inspecting the unit tests) will be able to choose.

### DIFF
--- a/exercises/resistor-color-trio/test/Tests.hs
+++ b/exercises/resistor-color-trio/test/Tests.hs
@@ -39,6 +39,7 @@ cases = [ Case { input         = Resistor (Black, Black, Black)
                , expectedLabel = "0 ohms"
                , expectedOhms  = 0
                }
+
         , Case { input         = Resistor (Black, Yellow, Black)
                , expectedLabel = "4 ohms"
                , expectedOhms  = 4
@@ -63,6 +64,7 @@ cases = [ Case { input         = Resistor (Black, Black, Black)
                , expectedLabel = "970 kiloohms"
                , expectedOhms  = 970e3
                }
+
         , Case { input         = Resistor (White, Black, Green)
                , expectedLabel = "9 megaohms"
                , expectedOhms  = 9e6
--- a/exercises/resistor-color-trio/test/Tests.hs
+++ b/exercises/resistor-color-trio/test/Tests.hs
@@ -1,0 +1,104 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE NumDecimals #-}
+
+import Data.Foldable     (for_)
+import Test.Hspec        (Spec, describe, it, shouldBe)
+import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+import Data.String       (fromString)
+
+import ResistorColors (Color(..), Resistor(..), label, ohms)
+
+main :: IO ()
+main = hspecWith defaultConfig {configFastFail = True} specs
+
+specs :: Spec
+specs = do
+  describe "label" $ for_ cases testLabel
+  describe "ohms" $ for_ cases testOhms
+  where
+    testLabel Case{..} = it (description input) $
+      label input `shouldBe` fromString expectedLabel
+      where
+        description Resistor{..} =
+          show bands ++ " should give " ++ expectedLabel
+
+    testOhms Case{..} = it (description input) $
+      ohms input `shouldBe` expectedOhms
+      where
+        description Resistor{..} =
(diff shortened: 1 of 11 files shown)


sshine:
This PR is currently marked as WIP because the example solutions don't handle the optional tests.

sshine:
Currently no special-cases for `"1 ohm"`, `"1 kiloohm"`, `"1 megaohm"` and `"1 gigaohm"`.

I'm unsure if they should be added as additional optional tests, or simply not be mentioned at all.

sshine:
Also, I learned something very cool: `1.2e3 :: Int` using the `NumDecimals` pragma.

sshine:
I've realized that the derived `Eq` instance is problematic:

```haskell
let ten1 = Resistor (Brown, Black, Black)
    ten2 = Resistor (Black, Brown, Brown)

ohms ten1 == ohms ten2 -- True
ten1      == ten2      -- False
```

Fortunately, since we only test `label` and `ohms`, we never compare `Color` or `Resistor`.

I have an idea for making a subsequent refactoring exercise where this insufficiency could come handy.

sshine:
I've not enabled the optional tests, so CI does not reveal that the example solutions work for them.

The following demonstrates that this is the case

## success-standard

```
$ stack test
...
label
  ...
  (Brown,Red,Red) should give 1.2 kiloohms
  (Orange,Yellow,Green) should give 3.4 megaohms
  (Green,Blue,Grey) should give 5.6 gigaohms

ohms
  ...
  (Brown,Red,Red) should give 1200
  (Orange,Yellow,Green) should give 3400000
  (Green,Blue,Grey) should give 5600000000

Finished in 0.0013 seconds
30 examples, 0 failures
```

## success-text

The exact same.

exercises/resistor-color-trio/test/Tests.hs
```diff
@@ -0,0 +1,106 @@
...
+               }
+        , Case { input         = Resistor (Green, Brown, Orange)
+               , expectedLabel = "51 kiloohms"
+               , expectedOhms  = 51e3
+               }
+        , Case { input         = Resistor (White, Violet, Yellow)
+               , expectedLabel = "970 kiloohms"
+               , expectedOhms  = 970e3
+               }
+
```
petertseng: ~~I see that there is a newline here, perhaps because of the change in SI prefix between kilo and mega. I wondered why there is not a newline between (no prefix) and kilo, and no newline between mega and giga?~~

Nope, I was wrong because I didn't read the commit message, which was easily fixed by my reading the commit message. I didn't even think of the explanation in the commit message because there are five cases in the canonical data and six cases between the newlines here. I have no change to suggest. It is indeed useful to have every colour represented

exercises/resistor-color-trio/test/Tests.hs
```diff
@@ -0,0 +1,106 @@
...
+               , expectedLabel = "3.4 megaohms"
+               , expectedOhms  = 3.4e6
+               }
+        , Case { input         = Resistor (Green, Blue, Grey)
+               , expectedLabel = "5.6 gigaohms"
+               , expectedOhms  = 5.6e9
+               }
+        -}
+        ]
+
```
petertseng: could remove this completely empty line

(but still do leave a trailing newline)

Issue #852, opened 2019-10-02
Bump canonical exercise versions


sshine:
Most exercises on the Haskell track have unit tests that are based on *canonical tests* shared between all Exercism tracks in [exercism/problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises). When alterations are made to these tests, the unit tests that mirror them on the Haskell track come out of sync.

The task at hand is to pick an exercise where versions have come out of sync and make the necessary changes to the Haskell test suite for that exercise. A list of exercises for which this is relevant can be generated by the following site: https://tracks.exercism.io/haskell/master (updates automatically) or using e.g. [`up-to-date/hs.rb`](https://github.com/petertseng/exercism-problem-specifications/tree/up-to-date/up-to-date).

Protocol: To avoid race conditions, you can declare that you intend to bump a specific exercise here before submitting a PR. One PR per version bump per exercise. Most exercises are only one version bump behind.

An extract from #849 made Oct 1, 2019, **updated Oct 17**:

- [ ] palindrome-products: 1.1.0.4 < 1.2.0
- [x] forth: 1.7.0.11 < 1.7.1
- [x] pangram: 1.4.1.11 < 2.0.0
- [ ] phone-number: 1.6.1.8 < 1.7.0
- [x] clock: 2.3.0.8 < 2.4.0
- [x] space-age: 1.1.0.5 < 1.2.0
- [x] robot-simulator: 3.1.0.7 < 3.2.0
- [ ] luhn: 1.3.0.6 < 1.6.1
- [ ] anagram: 1.4.0.8 < 1.5.0
- [ ] queen-attack: 2.2.0.7 < 2.3.0
- [x] acronym: 1.6.0.10 < 1.7.0
- [ ] allergies: 1.2.0.7 < 2.0.0
- [x] matrix: 1.1.0.7 < 1.3.0
- [ ] simple-cipher: 1.2.0.5 < 2.0.0
- [x] word-count: 1.3.0.7 < 1.4.0
- [x] armstrong-numbers: 1.0.0.2 < 1.1.0
- [x] leap: 1.5.1.9 < 1.6.0
- [x] etl: 1.0.0.5 < 2.0.1
- [ ] ~~beer-song~~
- [ ] ~~food-chain~~
- [ ] ~~house~~
- [ ] ~~triangle~~

[...]

PR #871, opened 2019-10-14
upgrade pangram tests to 2.0.0

barrymoo:
This PR updates the `pangram` tests original mentioned in https://github.com/exercism/haskell/issues/852

I found this commit https://github.com/exercism/problem-specifications/commit/f68b632f2578d9969074657d764b32f61f53c008 describing the recent changes to the tests. It seems that the descriptions are the only thing that changed. Is the goal to match the descriptions? Or match the canonical tests described in the specifications?

### DIFF
--- a/exercises/pangram/package.yaml
+++ b/exercises/pangram/package.yaml
@@ -1,5 +1,5 @@
 name: pangram
-version: 1.4.1.11
+version: 2.0.0.12
 
 dependencies:
   - base
--- a/exercises/pangram/test/Tests.hs
+++ b/exercises/pangram/test/Tests.hs
@@ -22,58 +22,58 @@ data Case = Case { description :: String
                  }
 
 cases :: [Case]
-cases = [ Case { description = "sentence empty"
+cases = [ Case { description = "with empty sentence"
                , input       = ""
                , expected    = False
                }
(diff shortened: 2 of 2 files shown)


petertseng:
The thing that is most important is the content. As we can see from `git show f68b632f2578d9969074657d764b32f61f53c008 -w`, only descriptions and indentation were changed in 2.0.0. I would say that unless a particular description doesn't make sense for the language that we might as well try to match the descriptions as well.

The original motivation for the changed descriptions was that descriptions that are too long don't work well for some tracks. This track does not appear to have such a problem, so if there are some description changes that actually lose some information, we can keep them.

On the other hand if there are some changes whose purpose is to be better English, I would say to probably take them. "sentence empty" vs "empty sentence" would be in that category, I suppose?

petertseng:
EBWODP:

> take them

That was not the best choice of words. I meant "apply them to this track"

barrymoo:
Great, I'll go through and compare the descriptions to see if any make sense to incorporate.

barrymoo:
In my previous PR #843 @sshine helped me with the `it` function from `hspec`. I modified the descriptions such that they read "pangram ...", e.g. "pangram with empty sentence".

PR #873, opened 2019-10-17
Update track README.md

sshine:
The updated README.md can be viewed here:

https://github.com/sshine/exercism-haskell/tree/simplify-readme#exercism-haskell-track

This pull request proposes the following:

- That we remove the 'Asking for help' section.
- That headers are bumped from `####` to `###` and from `###` to `##`.
- That we no longer link directly to the 'Getting involved' in a section of its own.
- That we add a 'Git and GitHub' help section at the top.
- That we fix a few broken links
- That we link to https://tracks.exercism.io/haskell when appropriate.
- That 'How to contribute' subheaders are written in present tense.

This PR breaks with the principle of one-change-only, unless one views updating README.md as that one change. I am willing to split them into separate PRs or remove individual proposed changes should anyone object.


README.md
```diff
@@ -24,12 +23,19 @@ Exercism exercises in Haskell
...
-### Reporting or fixing bugs
-Typical examples for a bug: A typo, a missing test case, an unclear or ambiguous
-problem description.
-- If you are unsure whether you have really found a bug [just ask](https://github.com/exercism/haskell/issues/new).
-- To report a bug you can [write an issue](#writing-an-issue).
-- If you already have a fix for it you may [write a pull request](#writing-a-pull-request).
+### Git and GitHub
+If you would like to contribute but lack experience with git and/or GitHub, try these resources:
+
+- [GitHub Help: Creating an issue](https://help.github.com/articles/creating-a-pull-request-from-a-fork/)
```
petertseng: the link target doesn't seem to match the text? creating an issue vs creating a pull request

README.md
```diff
@@ -25,21 +24,13 @@ Exercism exercises in Haskell
 - [Running tests](#running-tests)
 - [Running HLint](#running-hlint)
 
-### Asking for help
-When in doubt just ask!
-- [gitter support](https://gitter.im/exercism/support): more specific/technical questions.
-- [gitter dev](https://gitter.im/exercism/dev): technical questions.
```
petertseng: empirically, looking at all instances of the world "Haskell" occurring in exercism/dev dating back to January 2016, I count exactly one instance where it was a contributor looking to contribute. So at the rate of one contributor every 3.75 years, It seems we don't lose much by removing this link. Having them write an issue seems to work fine.

I have no data on exercism/support since I'm not interested in that channel.

README.md
sshine: Thanks for spotting.

PR #882, opened 2019-11-10
Add QuickCheck test for reshaping in "matrix" exercise

### LINKED ISSUE #881: Matrix exercise is under-specified by test suite
The only test for the `reshape` function in the test suite reshapes to a symmetric matrix:
```
reshape (2, 2) (intMatrix "1 2 3 4") `shouldBe` intMatrix "1 2\n3 4"
```
As such, there is nothing to prevent someone from writing a function that inverts the desired output dimensions  (possibly speaking from experience here.  >_>).  Either a case with non-equal rows/columns should be added or (probably better) the property that `shape . reshape dimension $ matrix == dimension` for all valid `dimension`/`matrix`.

Edit:  Hit send early.

SiriusStarr:
Per #881, adds a QuickCheck test for the property that `shape . reshape dimension $ matrix == dimension` for all valid `dimension`/`matrix`.

Closes #881.

### DIFF
--- a/exercises/matrix/test/Tests.hs
+++ b/exercises/matrix/test/Tests.hs
@@ -1,7 +1,7 @@
 import Control.Arrow     ((&&&))
 import Test.Hspec        (Spec, it, shouldBe, shouldNotBe)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
-import Test.QuickCheck
+import Test.QuickCheck   (Gen, choose, elements, forAllShow, listOf1, vector)
 
 import qualified Data.Vector as Vector (fromList)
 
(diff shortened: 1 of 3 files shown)


exercises/matrix/test/Tests.hs
```diff
@@ -1,6 +1,7 @@
 import Control.Arrow     ((&&&))
 import Test.Hspec        (Spec, it, shouldBe, shouldNotBe)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
+import Test.QuickCheck
```
sshine:
```suggestion
import Test.QuickCheck   (Gen, choose, elements, forAllShow, listOf1, vector)
```

exercises/matrix/test/Tests.hs
```diff
@@ -79,5 +80,28 @@ specs = do
...
+
+        matrixAndNewShape m = (\c -> (m, (size `quot` c, c))) <$> elements factors
+          where
+            size = rows m * cols m
+            factors = [n | n <- [1 .. size], size `rem` n == 0]
+
+        nonEmptyMatrix :: Gen (Matrix Int)
+        nonEmptyMatrix =
+          choose (1, 10)
+            >>= (fmap fromList . listOf1 . Test.QuickCheck.vector)
```
sshine:
```suggestion
            >>= (fmap fromList . listOf1 . vector)
```

SiriusStarr:
@sshine Changes should be as you requested.  Thanks for the feedback!  (Not having explicit imports was a dumb oversight, haha, I had intended to do that once I was done and forgot, clearly.)

SiriusStarr:
Whoops, didn't realize there were example solutions or I would have made certain to change that, haha.

sshine:
Apparently I've also asked you to do something that caused an overlap between two `vector` identifiers. Sorry about that. I think it's suitable for a squash-merge once CI has run.

SiriusStarr:
Yep, that was why I'd left `vector` qualified originally.  All good!

Issue #884, opened 2019-12-04
Participate in Exercism Research Experiment 1


sshine:
After the announcement of [Exercism Research Experiment](https://exercism.io/blog/announcing-exercism-research), a track-like repository is created in [`exercism/research_experiment_1`](https://github.com/exercism/research_experiment_1/) where some exercises are created with the following hypothesis:

>  The first experiment, which we aim to launch in the first week of January, will be centred around investigating how someone’s previous programming experience impacts the reusability of the code they produce in different languages.

In order to participate in this experiment, it is necessary to create a Haskell test runner (see e.g. [`exercism/rust-test-runner`](https://github.com/exercism/rust-test-runner) for comparison) and a number of exercises, `haskell-1-a`, `haskell-1-b`, `haskell-2-a`, `haskell-2-b`. The launch of this experiment is the first week of January, so this will have to be pushed over Christmas.

Note that I tried to get away with copy-pasting Erik's work in exercism/research_experiment_1#10, which is not an option. We're allowed to make minor variations as long as they're different problems. So I'll try and revise the `haskell-1-a` and `haskell-1-b` that were initially rejected.

PR #886, opened 2019-12-12
config.json: add "online_editor" section

sshine:
As part of "Research Experiment 1" (#884), and in correspondence with a similar pull request exercism/ruby#1016, this adds track configuration for the upcoming online editor.

Since this configuration does not allow context-sensitive indentation, any choice of "2 spaces" or "4 spaces" seems like an arbitrary default.


sshine:
Thanks. I agree with this. Sorry I haven't gotten back to fixing the CI yet.

sshine:
I'm closing this for now and will reopen it when `configlet fmt` is compatible with this change.

(It seems that tracks that employ this change don't have `configlet fmt` checks during CI.)

PR #894, opened 2020-02-24
bank-account: Fixed tests

nicuveo:
There were two major issues with those tests. First and foremost: they
were compiled without `-threaded`, meaning there was no actual
concurrency during the test runs, making the test that spawns twenty
threads meaningless.

Secondly, there's a behaviour that wasn't tested: if an account
exists, we expect increment and balance to never return
`Nothing`. That should only ever happen if the account is closed. This
commit therefore also adds a test that mixes increments and balance
checks to ensure all of them return `Just`.

This commit also increases the number of parallel threads to 50 (which
was the amount after which I could reliably make the tests fail on my
machine).

I have tried an alternative solution: use Control.Concurrent.QSemN to
release all threads "at the same time"; but since taking resources
from the semaphore is an atomic operation this wasn't working (each
thread was grabbing a resource one by one).

### DIFF
--- a/exercises/bank-account/examples/success-standard/src/BankAccount.hs
+++ b/exercises/bank-account/examples/success-standard/src/BankAccount.hs
@@ -3,7 +3,7 @@ module BankAccount ( BankAccount
                    , getBalance, incrementBalance) where
 import Control.Concurrent.STM (TVar, atomically, newTVar, readTVar, readTVarIO, writeTVar)
 
-newtype BankAccount = BankAccount { unBankAccount :: TVar (Maybe Int) }
+newtype BankAccount = BankAccount { unBankAccount :: TVar (Maybe Integer) }
 
 openAccount :: IO BankAccount
 openAccount = atomically $ BankAccount <$> newTVar (Just 0)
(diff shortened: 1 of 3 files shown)


exercises/bank-account/examples/success-standard/src/BankAccount.hs
```diff
@@ -3,18 +3,18 @@ module BankAccount ( BankAccount
...
 
 closeAccount :: BankAccount -> IO ()
 closeAccount = atomically . flip writeTVar Nothing . unBankAccount
 
-getBalance :: BankAccount -> IO (Maybe Int)
+getBalance :: BankAccount -> IO (Maybe Integer)
 getBalance = readTVarIO . unBankAccount
 
-incrementBalance :: BankAccount -> Int -> IO (Maybe Int)
+incrementBalance :: BankAccount -> Integer -> IO (Maybe Integer)
```
sshine: The PR/commit do not mention why `Int` was changed into `Integer`.

Investigating, it seems that the stub has used `Integer` ever since it was introduced in 0bfdca039de71a02f6352df14ef9175e44f4df51 (2016-07-13), but the example solution of d956febd4043e1091300b1c3440a1fbd28dd52c1 (2013-10-25) uses `Int`.

So on the premise of consistency alone, this is a reasonable change.

nicuveo: Ah, yes, sorry! Yup, I was just fixing the inconsistency between stub and example.

nicuveo:
Bumped the version number, renamed the commit to have a 'bank-account:' prefix.

And yeah, running 50 threads in the hope of creating a collision is inelegant, but I couldn't find an elegant solution that does not put additional restrictions on the user's code... I tried the semaphore approach, but it didn't work either. Hopefully, this will be consistent enough to find the most glaring bugs.

sshine:
Since the force-push I've rebased rather than merged, but I have come to realize that the `bank-account: bumped serial version number` commit is probably a bit on its own, which would have continued to warrant a merge commit. Sorry about that.

PR #896, opened 2020-03-03
dominoes: added double-orientation test

nicuveo:
I found a case where a student's code would mistakenly use the same
domino in both ways. With their code, the chain
`[(1, 2), (2, 3), (2,1)]` would be considered valid, as they would
first match (1,2) with (2,3), then swap the domino and match (3,2)
with (2,1).

Surprisingly, none of the tests currently catch this, hence this
patch.

### DIFF
--- a/exercises/dominoes/package.yaml
+++ b/exercises/dominoes/package.yaml
@@ -1,5 +1,5 @@
 name: dominoes
-version: 2.1.0.7
+version: 2.1.0.8
 
 dependencies:
   - base
--- a/exercises/dominoes/test/Tests.hs
+++ b/exercises/dominoes/test/Tests.hs
@@ -74,6 +74,10 @@ cases = [ Case { description = "empty input = empty output"
                , input       = [(1, 2), (1, 3), (2, 3)]
                , expected    = True
                }
+        , Case { description = "cannot use the same domino in both directions"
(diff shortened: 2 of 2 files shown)


nicuveo:
Yup, the fetch-configlet commit was only included in this branch to make the tests pass; I will rebase as soon as that other PR is submitted.

For the version number, I wasn't sure what to do, since I'm suggesting two changes. If you approve of both, I could merge them in one PR that also bumps the version number?

nicuveo:
I don't think I have the permissions to tag this PR as `hold`. Can it still be done after it's merged?

sshine:
> I don't think I have the permissions to tag this PR as `hold`.
>
> Can it still be done after it's merged?

I'm sorry for any lack of clarity.

What I intended to do was this: Merge this PR so that the Haskell track may benefit from the unit test right away, and potentially create a pull request in the exercism/problem-specifications repository, mark *that* pull request as `hold`, and wait for the exercism/problem-specifications lockdown to get resolved.

Since there is no clear indication of when that will happen, the consensus right now is that track maintainers should apply whatever additional unit tests when they feel like i. My additional opinion, which seems to be reflected among other contributors, is that submitting a pull request to exercism/problem-specifications and putting it on hold functions as a signal to other tracks that an additional unit test is in the pipeline. So we're sort of queuing them up.

PR #897, opened 2020-03-03
dominoes: add long test with no solution

nicuveo:
This is meant to discourage students from using a solution that simply
tries permutations of the domino chain.


nicuveo:
I am not sure I understand why you think it is unwarranted to propagate this test? This simply tests point 1: that the submitted code claims there is a chain if and only if there actually is one. I might be reading the comment wrong?

sshine:
I'm sorry, you're right. That test case does make sense to propagate. I'm not sure what I was thinking.

PR #898, opened 2020-03-03
Fix fetch-configlet.

nicuveo:
Travis builds are failing because of fetch-configlet failures, as tracked in https://github.com/exercism/configlet/issues/173. This patch is a copy of https://github.com/exercism/cpp/pull/344, which fixed the cpp track.


nicuveo:
Yep, your version looks much better; I'll amend this PR.

PR #900, opened 2020-03-14
Luhn: Update tests to 1.7.0 (1 changed tests, 2 new tests)

tejasbubane:
Changes included:

1. exercism/problem-specifications#1480: 1.4.0 to 1.5.0:
   Change "055a 444 285" test into "055b 444 285" for subtle reasons.
   New test name: using ascii value for doubled non-digit isn't allowed

2. exercism/problem-specifications#1500: 1.5.0 to 1.6.0:
   Add test case: valid number with an odd number of spaces

3. exercism/problem-specifications#1635: 1.6.0 to 1.7.0:
   Add test case: invalid long number with an even remainder

*(Edit by @sshine: Added short description of the addressed commits.)*

### DIFF
--- a/exercises/luhn/test/Tests.hs
+++ b/exercises/luhn/test/Tests.hs
@@ -48,14 +48,22 @@ cases = [ Case { description = "single digit strings can not be valid"
                , input       = "8273 1232 7352 0569"
                , expected    = False
                }
+        , Case { description = "invalid long number with an even remainder"
+               , input       = "1 2345 6789 1234 5678 9012"
+               , expected    = False
+               }
         , Case { description = "valid number with an even number of digits"
                , input       = "095 245 88"
                , expected    = True
                }
+        , Case { description = "valid number with an odd number of spaces"
+               , input       = "234 567 891 234"
+               , expected    = True
+               }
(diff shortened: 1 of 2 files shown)


exercises/luhn/test/Tests.hs
```diff
@@ -48,14 +48,22 @@ cases = [ Case { description = "single digit strings can not be valid"
...
         , Case { description = "valid number with an even number of digits"
                , input       = "095 245 88"
                , expected    = True
                }
+        , Case { description = "valid number with an odd number of spaces"
+               , input       = "234 567 891 234"
+               , expected    = True
+               }
 -- This track is not testing these cases, since we would rather focus on the algorithm,
 -- and because it seems strange to be unable to distinguish between well-formed invalid input and malformed input.
```
sshine: I'm remarking that the comment above was left by 610ba692834d5e7a7724e7ead0cff7877615010b.

tejasbubane:
>  A finer granularity comes at the cost of time spent now, but makes both reading `git blame` later and reviewing now much easier.

I'll be mindful of this in future. Thanks! 👍

PR #905, opened 2020-04-17
stack.yaml: lts-14.7 -> lts-15.8

tejasbubane:
https://www.stackage.org/diff/lts-14.7/lts-15.8

Last update 6 months ago #846

I have confirmed that no README check is needed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ cache:
     - $HOME/.foldercache # Per exercise `.stack-work` cache.
 
 env:
- - RESOLVER="lts-14.7" CURRENT="YES"  # Equal to each stack.yaml.
+ - RESOLVER="lts-15.8" CURRENT="YES"  # Equal to each stack.yaml.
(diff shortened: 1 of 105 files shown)


tejasbubane:
@sshine Build passed.

sshine:
@tejasbubane: Excellent.

@kytrinyx, @iHiD: Could we please get "merge commit" and "rebase and merge" back? I requested "merge commit" back some weeks ago, but this resulted in losing "rebase and merge" instead, and we haven't had that many PRs since then. In this particular case I'd like "merge commit", but most often I'd like "rebase and merge".

iHiD:
All three are back on now :)

sshine:
@ihid: "Create a merge commit" still says "Not enabled for this branch."

In the process of learning why that is, in case this particular PR remembers things from before the setting was flipped, shall I bring it up again if/when another PR cannot be merged, rather than we try and fix our merge options on a commit made before in time from changing the setting?

sshine:
I've rebased because both commits explain themselves well.

PR #906, opened 2020-04-18
nucleotide-count: use findWithDefault to accept sparse maps

### LINKED ISSUE #902: nucleotide-count: awkward to require filling out sparse `Map` datastructure
The tests reject solutions that don't fill out returned map with `0` for missing nucleotides.  @sshine and I were talking about this while I was doing the track exercises and we thought it might be nice to remove this pain point. Some possible approaches:

* Have solution return a function `Nucleotide -> Int` which leaves the student free to choose whatever representation.
* Rewrite the success predicate `matchesMap` to be more liberal.
* Use a dense structure like `array` instead of `map`.
* Something else?

sshine:
Bump version from 1.3.0.7 to 1.3.0.8.

This was created by @jitwit in #903 and accidentally closed by a force-push by me.

This closes #902.

### DIFF
--- a/exercises/nucleotide-count/test/Tests.hs
+++ b/exercises/nucleotide-count/test/Tests.hs
@@ -1,8 +1,9 @@
 {-# OPTIONS_GHC -fno-warn-type-defaults #-}
+{-# language LambdaCase #-}
 
 import Data.Either       (isLeft)
-import Data.Map          (fromList)
-import Test.Hspec        (Spec, describe, it, shouldBe, shouldSatisfy)
+import Data.Map          (findWithDefault)
+import Test.Hspec        (Spec, describe, it, shouldSatisfy)
 import Test.Hspec.Runner (configFastFail, defaultConfig, hspecWith)
 
 import DNA (nucleotideCounts, Nucleotide(..))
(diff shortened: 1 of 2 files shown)


exercises/nucleotide-count/test/Tests.hs
```diff
@@ -13,7 +14,9 @@ main = hspecWith defaultConfig {configFastFail = True} specs
 specs :: Spec
 specs = do
 
-          let x `matchesMap` y = x `shouldBe` (Right . fromList) y
+          let x `matchesMap` y = shouldSatisfy x $ \case
+                Right count -> and [ findWithDefault 0 n count == c | (n,c) <- y ]
```
jitwit, line 18: I ran `hlint` locally and didn't get any hints/warnings/complaints for this as is. Are there settings for stricter and harsher linting? 

I'd be totally fine with adapting to either suggestion you made in https://github.com/jitwit/haskell/pull/1/files. I am a fan of list comprehensions in general (and I think the combination of getting `count` in scope with pattern matching with comprehension is quite natural), so I'd be happy keeping as is, though I agree that the `and`  is slightly awkward...

sshine: I thought it might have been [the .hlint.yaml of Kowainik's relude](https://github.com/kowainik/relude/blob/master/.hlint.yaml), but I can't point at the particular rule, so it might have been a different one. (This .hlint.yaml assumes the existence of various relude-specific definitions.)

This solution is quite fine!

PR #910, opened 2020-04-20
clock: Improve hint

### LINKED ISSUE #908: Feedback: Exercise Clock is a little confusing
Right at the beginning it has `Implement a clock that handles times without dates.` I can't tell if this means without a date library, or just dateless (like from 11:59 am -11:59 pm).

It also doesn't have clear instructions on what fromHourMin and toString are supposed to do. I can infer fromHourMin from the type signature but toString didn't seem to have a place.

Also as this is the first challenge creating data types, it would be nice if there were a simple example right in the instructions.

kevin-meyers:
Closes #908 
I added a few things:
- Updated implementation lines to specifiy the type of clock, and that you can add or subtract both hours and minutes from it.
- Added descriptions for the remaining 2 functions: toString and fromHourMin
- Added a resource from learn you a haskell for creating and using custom types
- Added restrictions to not use the inbuilt data.dates package. I followed the format of other problems.

### DIFF
--- a/exercises/clock/README.md
+++ b/exercises/clock/README.md
@@ -1,8 +1,8 @@
 # Clock
 
-Implement a clock that handles times without dates.
+Implement a 24 hour clock that handles times.
 
-You should be able to add and subtract minutes to it.
+You should be able to add and subtract hours and minutes to it.
 
 Two clocks that represent the same time should be equal to each other.
 
@@ -18,9 +18,20 @@ add an `Eq` instance, and implement the functions:
 `addDelta` adds a duration, expressed in hours and minutes, to a given time,
 represented by an instance of `Clock`.
 
+`fromHourMin` takes an hour and minute, and returns an instance of `Clock` with 
+those hours and minutes.
+
+`toString` takes an instance of `Clock` and returns a string representation 
+of the clock, in 0 padded format like "08:03" or "22:35"
+
 You will find a dummy data declaration and type signatures already in place,
 but it is up to you to define the functions and create a meaningful data type,
 newtype or type synonym.
+
+If you need help getting started with Types, take a look at:
+- [Data Types in 5 Steps][types]
+
+[types]: https://mmhaskell.com/blog/2017/12/24/haskell-data-types-in-5-steps
 
 
 
--- a/exercises/clock/README.md
+++ b/exercises/clock/README.md
@@ -1,10 +1,12 @@
 # Clock
 
-Implement a 24 hour clock that handles times.
+Implement a clock that handles times without dates.
 
-You should be able to add and subtract hours and minutes to it.
+You should be able to add and subtract minutes to it.
 
 Two clocks that represent the same time should be equal to each other.
+
+It's a 24 hour clock going from "00:00" to "23:59".
 
 ## Hints
 
@@ -22,7 +24,7 @@ represented by an instance of `Clock`.
 those hours and minutes.
 
 `toString` takes an instance of `Clock` and returns a string representation 
-of the clock, in 0 padded format like "08:03" or "22:35"
+of the clock, in 0-padded format like "08:03" or "22:35"
 
 You will find a dummy data declaration and type signatures already in place,
 but it is up to you to define the functions and create a meaningful data type,
--- a/exercises/clock/README.md
+++ b/exercises/clock/README.md
@@ -6,9 +6,9 @@ You should be able to add and subtract minutes to it.
 
 Two clocks that represent the same time should be equal to each other.
 
-It's a 24 hour clock going from "00:00" to "23:59".
-
 ## Hints
+
+It's a 24 hour clock going from "00:00" to "23:59".
 
 To complete this exercise you need to define the data type `Clock`,
 add an `Eq` instance, and implement the functions:
@@ -30,7 +30,7 @@ You will find a dummy data declaration and type signatures already in place,
 but it is up to you to define the functions and create a meaningful data type,
 newtype or type synonym.
 
-If you need help getting started with Types, take a look at:
+If you need help getting started with types, take a look at:
 - [Data Types in 5 Steps][types]
 
 [types]: https://mmhaskell.com/blog/2017/12/24/haskell-data-types-in-5-steps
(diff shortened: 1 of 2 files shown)


kevin-meyers:
I am new to contributing both to exercism and in general, can someone stop by and let me know why editing the README is causing failing tests?

exercises/clock/README.md
```diff
@@ -18,11 +18,24 @@ add an `Eq` instance, and implement the functions:
 `addDelta` adds a duration, expressed in hours and minutes, to a given time,
 represented by an instance of `Clock`.
 
+`fromHourMin` takes an hour and minute, and returns an instance of `Clock` with 
+those hours and minutes.
+
+`toString` takes an instance of `Clock` and returns a string representation 
+of the clock, in 0 padded format like "08:03" or "22:35"
```
sshine, line 25:
```suggestion
of the clock, in 0-padded format like "08:03" or "22:35"
```

exercises/clock/README.md
```diff
@@ -18,11 +18,24 @@ add an `Eq` instance, and implement the functions:
...
+
 You will find a dummy data declaration and type signatures already in place,
 but it is up to you to define the functions and create a meaningful data type,
 newtype or type synonym.
 
+If you need help getting started with Types, take a look at:
+- [Making Our Own Types and Typeclasses][types]
+
+[types]: http://learnyouahaskell.com/making-our-own-types-and-typeclasses
```
sshine, line 34: I don't think this chapter is ideal for addressing this exercise.

The first half of the chapter only deals with `data` definitions, and about halfway through it addresses `type` aliases. But since this exercise is best solved with a `newtype`, and this chapter doesn't even mention `newtype`, it is perhaps more confusing than clarifying.

Having a hyperlink is good, but preferrably to something that focuses on the particular learning goals of the exercise, rather than types in general.

I would recommend either finding another source or removing this hint.

kevin-meyers: Got it, you're right! I used it when I solved the problem but I had to jump down at least halfway because I knew where what I was looking for was. If I can't find a better resource thats more concise I'll remove it.

kevin-meyers:
Thank you for getting back to me @sshine !  I appreciate you breaking it down for me, I did find .meta/hints when I was digging around. But since it looked like it had the same contents in both hints and the README i ignored it.

I believe I understand correctly: Upstream the problem description that is shared between haskell/python etc, all of them may benefit from the change. But because [I assume v3 is in progress] the change is quickest made local to haskell here in the hint file.

One quick question though, what is your comment on restrictions?

kevin-meyers:
Also I wanted to let you know I appreciate that you took the time to explain the steps to me instead of just tackling it on your own, even though it wouldve taken you no time at all.

exercises/clock/README.md
```diff
@@ -18,11 +18,24 @@ add an `Eq` instance, and implement the functions:
...
+
+## Restrictions
 
+Don't use the Data.Dates package, create the functions yourself!
```
sshine, line 38: > One quick question though, what is your comment on restrictions?

So that comment disappeared because of a bug in GitHub, it seems.

I'll try to repeat it.

You may be confusing "package" and "module". I'm not sure exactly what package or module you're referring to, since the most used date/time package, `time`, has the module `Data.Time.Clock`, and [various other popular time libraries](https://jacobstanley.io/3-packages-you-need-to-know-about-before-processing-timestamps-in-haskell/) have different ones.

I'd get rid of this restriction because using library functions [encourages fluency](https://exercism.io/values). Alternatively, if mentors report that a majority of students solve this exercise in a too simple manner, one could consider deprecating it like "gigasecond" was deprecated in #280 (see argument in #230).

```suggestion
(remove this; I cannot seem to suggest emptiness)
```

kevin-meyers: Sounds good, I like that philosophy

kevin-meyers:
I will have time to fix it all up either in a couple hours tonight or in the morning!

exercises/clock/README.md
kevin-meyers: So I found a good resource https://mmhaskell.com/blog/2017/12/24/haskell-data-types-in-5-steps

I am going to replace it for now so let me know what you think. (or if its just better off removed). My motivation to adding something was driven because I didn't know how to get the value of 'hour' or something out of a clock object.

kevin-meyers:
Okay @sshine so I what I believed is right, and when I ran ./bin/ensure-readmes-are-updated.sh, it generated a large problems-specifications folder, but what was inside wasnt too interesting, and it didnt seem to have a place in the repo either... So I just confirmed that it worked and removed the problems-specifications folder.

I might need your eyes again haha!

sshine:
@kevin-meyers: What that script does is create a copy of problem-specifications, and rebuilds README.md's based on the concatenation I described. What I personally do is move the problem-specifications git repo one level out and leave a symlink inside the exercism haskell repo. But you can just keep a copy of problem-specifications in the track repo, or delete it again.

What that does, if you inspect the script, is that it re-generates the README. So after you have changed the hint file and regenerate the README, your README will contain the hint in such a way that the CI script will regard them as having been built properly.

The part of bin/ensure-readmes-are-updated.sh that does this is:

```bash
$configlet generate . --only "$exercise" --spec-path "problem-specifications"
```

This assumes you have `configlet` installed.

kevin-meyers:
I found configlet (i didnt realize it was a part of exercism so i was scouring the internet!), and installed it. Managed to see the generated files! now I am making sure it looks correct and passes the tests.

kevin-meyers:
I put the changes in hints, and the descriptions changes in the problem specifications file. I then used the command you provided me with configlet and the path of problem-specifications to generate the README as described!

I get the feeling it will still break though but I remain optimistic.

kevin-meyers:
okay so it says its out of date, that might be because I edited the problem specifications and shouldnt have... So I feel like I can only change everything below the clock basic description. I am going to try leaving the problem specifications alone, and just "add" information above `## hints` as if it were a continuation of description.

kevin-meyers:
Sorry about running the integration tests a couple times here. If this one doesn't click I will wait until you are available again. Thank you so much for your patience

kevin-meyers:
This has been a long and ardous journey haha! Thank you so much for taking the time to guide me through the process and philosophies behind making changes as little as sentence positioning.

I am continuing to learn haskell :muscle:, I look forward to contributing more in the future (hopefully with less hand holding required!)

I am a huge fan of the platform, and couldn't have asked for a better way to get in gear.

Issue #864, opened 2019-10-11
package.yaml much behind on some exercises


chevinbrown:
I've had trouble running `stack test` on the hello-world exercise with ghc `8.6.5` and `8.8.1` (which is all I tried before finding `8.4.4` worked). 

The installation guide may need to specify a ghc version.

sshine:
Hi @chevinbrown, and thanks for reporting any difficulties with the track.

`stack test` is supposed to look in stack.yaml for a `resolver: lts-...` and use the GHC version for that [Stackage.org LTS version](https://www.stackage.org/). Incidentally, I saw your issue just as I was going to comment on #846, where @petertseng bumps the LTS to 14.7, corresponding to GHC 8.6, on this exact matter:

Because of a behavior in the current website, students don't get updated copies of handout files unless the test file specifically was touched. That means, for example for hello-world, that the stack.yaml students get is two years old, because this was when test/Tests.hs was last changed.

The effect of some exercises using GHC 8.2, some 8.4, and some 8.6 is pretty bad: It potentially means that students will download new GHC versions when they advance between single exercises.

A simple solution could be to mass update all test/Tests.hs files with a linebreak.

chevinbrown:
Makes sense. This is pretty unattainable for _students_ to understand, IMO. 
I've never done any haskell myself, so I had to hunt these things down _without any knowledge of the ecosystem_. 😕 

As a student, I feel like a "recommended version" could be a great starting point. I can see this being a barrier to entry for less experienced or less persistent students.

Would it be helpful for me to submit a PR that ensures consistency across all exercises?

chevinbrown:
I did just update `hello-world` `stack.yaml` to `resolver: lts-14.8` with ghc `8.6.5` and we're back on track.

sshine:
> "recommended version"

I'm not sure what you mean, exactly. Maybe you could elaborate? The track currently recommends (and really necessitates) the use of Stack (since no .cabal file is provided independently of package.yaml), and the "Guides" section on the track instructs how to compile and test code.

> Would it be helpful for me to submit a PR that ensures consistency across all exercises?

Let's wait a short while. I'd like to investigate the cost of fixing the bug rather than working around it.

sshine:
> Would it be helpful for me to submit a PR that ensures consistency across all exercises?

I don't have the capacity to fix the bug, so let's aim for a workaround.

Since we're going through all this trouble, we might as well copy #846 and bump LTS to 14.9.

[...]

PR #922, opened 2020-11-09
fix stack.yaml to stop breaking Haskell LSP

dvthai89:
I encountered exactly the problem raised in this [issue](https://github.com/haskell/haskell-language-server/issues/511) so I thought I should fix it to avoid trouble for latecomers.

### DIFF
--- a/exercises/perfect-numbers/stack.yaml
+++ b/exercises/perfect-numbers/stack.yaml
@@ -1,2 +1,3 @@
 resolver: lts-15.8
 packages:
+- .
--- a/exercises/perfect-numbers/stack.yaml
+++ b/exercises/perfect-numbers/stack.yaml
@@ -1,1 +1,2 @@
 resolver: lts-15.8
+packages:


exercises/perfect-numbers/stack.yaml
```diff
@@ -1 +1,2 @@
 resolver: lts-15.8
+packages:
```
sshine, line 2:
```suggestion
packages:
- .
```

sshine (changes requested):
I can reproduce this problem. Thanks for reporting it.

As [this commenter](https://github.com/haskell/haskell-language-server/issues/511#issuecomment-713736086) points out, `packages: [.]` is optional in stack.yaml. Leaving out `packages:` is interpreted by the Stack tool as an implicit `packages: [.]`, which is somehow lost in the LSP pipeline. As [this commenter](https://github.com/haskell/haskell-language-server/issues/511#issuecomment-715319236) points out, it might be a bug in hie-bios, which is a sub-component of the LSP plugin.

So your proposal appears to be that we patch all stack.yaml's with explicit `packages: [.]` values to improve the LSP experience. In that case, there should be a difference between `packages:` and `packages: [.]`, since the project should provide its own root directory as a package.

I've tested this and it appears that I get other error messages if the `.` package is omitted.

In that case:
- The cost of making `packages: [.]` explicit is a slight increase in verbosity.
- We should want to do this for all exercises, not single ones.
- Updating files for distribution to students is a bit complicated. The system around the website will only register updates if they also occur in the .hs file. So what #864 addresses is to add a bunch of commented-out checksums to .hs files when surrounding files are updated. #877 attempts this, but the effort got stranded.

My proposal is:
- Let this PR add
  ```
  packages:
  - .
  ```
  to all exercises, disregarding that this won't actually propagate to students on its own.
- Subsequently fix and merge #877.

Would you be interested in generalising the solution to all exercises in this PR?

exercises/perfect-numbers/stack.yaml
dvthai89: ok will fix all `stack.yaml` files.

PR #925, opened 2020-11-14
Improve the way CI checks that stack resolvers are synced

sshine:
As witnessed in #924, it is not possible to have lines in stack.yaml that don't match `^resolver: ${expected_stack}\$`.

This PR attempts to rephrase the way we make this check so that stack.yaml can also contain other lines, e.g. `packages: ...`.

(This is performed in incremental steps to trigger CI in each step.)

*Edit:* This check was originally added in #661.


sshine:
- [ ] Make `yq` available on Travis CI image, use another way to extract `resolver`, or wait until Travis CI is replaced with GitHub Actions. If #924 eventually waits for this PR, pick whatever choice is most convenient at that point.

bin/sync-stack-resolvers.sh
```diff
@@ -0,0 +1,19 @@
...
+    if ! [ "$exercise_stack" = "$expected_stack" ]; then
+      differing_stack="$differing_stack $(basename "$exercise")"
+    fi
+done
+if [ -n "$differing_stack" ]; then
+    echo "The following exercises have a different stack.yaml resolver:$differing_stack"
+    echo "They should instead be $expected_stack"
+    exit 1
+fi
+
```
petertseng, line 19: extra empty line can be removed (so that the file ends with just one newline instead of two)

.travis.yml
```diff
@@ -36,22 +39,7 @@ install:
...
-        if grep -v "^resolver: ${expected_stack}\$" $exercise/stack.yaml; then
-          differing_stack="$differing_stack $(basename "$exercise")"
-        fi
-    done
-    if [ -n "$differing_stack" ]; then
-        echo "The following exercises have a different stack.yaml resolver:$differing_stack"
-        echo "They should instead be $expected_stack"
-        exit 1
-    fi
+ - "sh ./bin/sync-stack-resolvers.sh"
```
petertseng, line 42: the verb "sync" to me implies that this file would actually change files such that the `resolver` are the same, but that may just be a biased interpretation of the word from my POV. If you ask me, I would make it `ensure-stack-resolvers-synced`.

petertseng (approved):
my comments are optional

petertseng (approved):

Issue #921, opened 2020-10-15
Moving from Travis to GitHub Actions


ErikSchierboom:
Hello 🙂 

Over the last few months we've been transferring all our CI from Travis to [GitHub Actions (GHA)](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions). We've found that GHA are easier to work with, more reliable, and much much faster. 

Based on our success with GHA and increasing intermittent failures on Travis, we have now decided to try and remove Travis from Exercism's org altogether and [shift everything to GHA](https://github.com/exercism/exercism/issues/5396). This issue acts as a call to action if your track is still using Travis. 

For most CI checks this should be a transposing from Travis' syntax to [GHA syntax](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions), and hopefully quite straightforward (see [this PR for an example](https://github.com/exercism/csharp/pull/1338/files)).  However, if you do encounter any issues doing this, please ask on Slack where lots of us now have experience with GHA, or post a comment here and I'll tag relevant people. This would also make a good Hacktoberfest issue for anyone interested in making their first contribution 🙂 

If you've already switched this track to GHA, please feel free to close this issue and ignore it.

Thanks!

ErikSchierboom:
Thanks to @SleeplessByte, we now have a really nice document and some starting workflows that you can use to help switch to GitHub Actions: https://github.com/exercism/problem-specifications/tree/master/workflows#workflow-templates

petertseng:
To work on this issue, what you're going to want to do is to move each of the checks in https://github.com/exercism/haskell/blob/master/.travis.yml over, and I recommend doing them piecemeal. For example here is one that will just do `configlet lint`

```yaml
name: Tests
on:
  pull_request:
  push:
    branches:
    - YOURBRANCHNAME

jobs:
  track-config:
    name: Check track configuration
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - name: Fetch
        run: bin/fetch-configlet
      - name: Lint
        run: bin/configlet lint .
```

(Notice where it says YOURBRANCHNAME - use your branch name there so GHA runs on your branch on your repo)

The checks to be moved are as follows. If you have a branch with that particular check moved over, check the box add a link to that branch if you have edit access, otherwise post a comment linking to that branch. When they're all done we'll combine them all.

- [ ] Ensure READMEs updated
- [ ] configlet fmt
[...]

PR #930, opened 2020-11-17
Move from Travis CI to GitHub Actions

petertseng:
Closes https://github.com/exercism/haskell/issues/921

The Exercism organisation is moving to GitHub Actions.
It runs faster.

These commits should probably be squashed, however they have been designed so that even if the person who merges it accidentally choose not to squash that they will each still make sense as standalone commits.

The commit initially at the head of this branch will cause the build to fail as verification that incorrect code is still rejected. The commit will be reverted as soon as that is verified.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,3 +33,77 @@ jobs:
             echo "$bad_uuid"
             exit 1
           fi
+
+  exercises:
+    name: Check exercises
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        resolver:
+          - lts-from-exercises
+          - nightly
+    # Discussion seems to indicate that continue-on-error is not exactly equivalent to Travis's allow_failures, but it's the best we can do for now.
+    # If failures on nightly cause too much confusion in CI status,
+    # consider just not running against the nightly resolver.
+    # https://github.com/actions/toolkit/issues/399
+    continue-on-error: ${{ matrix.resolver == 'nightly' }}
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install yq (for stack resolvers)
+        run: |
+         sudo add-apt-repository -y ppa:rmescandon/yq
+         sudo apt-get -q update
+         sudo apt-get -y install yq
+
+      - name: Resolve resolver
+        id: resolve-resolver
+        run: |
+          # We need this in two places:
+          # 1. cache key
+          # 2. --resolver arg to installing hlint
+          # It's unsafe to just use lts-from-exercises as the cache key,
+          # because that would cause caches from differing LTSes to collide
+          # (if we have updated the LTS version of all exercises between job runs).
+          if [ "${{ matrix.resolver }}" = "lts-from-exercises" ]; then
+            resolver=$(yq read "$(ls -1 exercises/*/stack.yaml | head -1)" resolver)
+          else
+            resolver=${{ matrix.resolver }}
+          fi
+          echo "::set-output name=resolver::$resolver"
+
+      - name: Week number
+        id: week-no
+        run: |
+          # We use this to form part of the cache key.
+          # This will get us to periodically rebuild from scratch,
+          # just to make sure that everything still works from scratch.
+          # Because we also have a weekly build at the start of every week,
+          # most of the time caches should hit.
+          echo "::set-output name=week-no::$(date -u +%Y-%U)"
+
+      - uses: actions/cache@v2
+        id: cache
+        with:
+          path: |
+            ~/.stack
+            ~/.local/bin/hlint
+          key: ${{ steps.resolve-resolver.outputs.resolver }}-${{ steps.week-no.outputs.week-no }}-${{ hashFiles('exercises/**/package.yaml') }}
+          restore-keys: ${{ steps.resolve-resolver.outputs.resolver }}-${{ steps.week-no.outputs.week-no }}-
+
+      - uses: actions/setup-haskell@v1
+        with:
+          enable-stack: true
+          stack-version: 'latest'
+
+      - name: Install hlint
+        run: |
+          if [ -x ${HOME}/.local/bin/hlint ]; then
+            echo already installed
+          else
+            stack --resolver ${{ steps.resolve-resolver.outputs.resolver }} install hlint
+          fi
+          ${HOME}/.local/bin/hlint --version
+
+      - name: HLint
+        run: ${HOME}/.local/bin/hlint ${{ github.workspace }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,6 +16,12 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Fetch
+        run: bin/fetch-configlet
+
+      - name: Lint
+        run: bin/configlet lint .
+
       - name: Check for invalid UUIDs
         # can be removed once `configlet lint` gains this ability.
         # Check issue https://github.com/exercism/configlet/issues/99
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,29 @@
+name: Tests
+on:
+  pull_request:
+  push:
+    branches:
+    - gha
+    - master
+  schedule:
+    # Weekly.
+    - cron: "0 0 * * 0"
+
+jobs:
+  track-config:
+    name: Check track configuration
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Check for invalid UUIDs
+        # can be removed once `configlet lint` gains this ability.
+        # Check issue https://github.com/exercism/configlet/issues/99
+        run: |
+          uuids=$(jq --raw-output '.exercises | map(.uuid) | .[]' config.json)
+          bad_uuid=$(echo "$uuids" | grep -vE '^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$' || true)
+          if [ -n "$bad_uuid" ]; then
+            echo "invalid UUIDs found! please correct these to be valid UUIDs:"
+            echo "$bad_uuid"
+            exit 1
+          fi
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,7 +3,6 @@ on:
   pull_request:
   push:
     branches:
-    - gha
     - master
   schedule:
     # Weekly.
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,6 +14,12 @@ jobs:
     name: Check track configuration
     runs-on: ubuntu-latest
     steps:
+      # master is needed in addition to HEAD,
+      # because the README check only checks exercises changed since master.
+      - uses: actions/checkout@v2
+        with:
+          ref: master
+
       - uses: actions/checkout@v2
 
       - name: Fetch
@@ -21,6 +27,21 @@ jobs:
 
       - name: Lint
         run: bin/configlet lint .
+
+      - name: Ensure READMEs are updated
+        run: sh ./bin/ensure-readmes-are-updated.sh
+
+      - name: Check configlet fmt
+        run: sh ./bin/check-configlet-fmt.sh
+
+      - name: Install yq (for stack resolvers)
+        run: |
+         sudo add-apt-repository -y ppa:rmescandon/yq
+         sudo apt-get -q update
+         sudo apt-get -y install yq
+
+      - name: Ensure stack resolvers synced
+        run: sh ./bin/ensure-stack-resolvers-synced.sh
 
       - name: Check for invalid UUIDs
         # can be removed once `configlet lint` gains this ability.
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -108,6 +108,7 @@ jobs:
         with:
           path: |
             ~/.stack
+            ~/.foldercache
             ~/.local/bin/hlint
           key: ${{ steps.resolve-resolver.outputs.resolver }}-${{ steps.week-no.outputs.week-no }}-${{ hashFiles('exercises/**/package.yaml') }}
           restore-keys: ${{ steps.resolve-resolver.outputs.resolver }}-${{ steps.week-no.outputs.week-no }}-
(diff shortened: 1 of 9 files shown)


petertseng:
Confirmed: https://github.com/petertseng/exercism-haskell/actions/runs/367459920 indicates failure.

petertseng:
Deliberate failure: https://github.com/petertseng/exercism-haskell/actions/runs/368342710
Success: https://github.com/petertseng/exercism-haskell/actions/runs/368343647

.travis.yml
```diff
@@ -1,67 +1,6 @@
...
-    # Explicit set exercises' resolver only if it's not the current one.
-    if [ "${CURRENT}" != "YES" ]; then
-        export SET_RESOLVER="--resolver ${RESOLVER}"
-    fi
-
-    for exercise in ${TRAVIS_BUILD_DIR}/exercises/*/ ; do
-        time bin/test-stub $exercise
-        time bin/test-all-examples $exercise
-    done
+ - /bin/true
```
petertseng, line 6: not yet able to remove this file completely. it's a required check, so removing it means can't merge this PR. Will need to ask for the new GitHub Actions checks to be made required, and Travis to be made not required, after this PR is merged.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,93 @@
...
+            ~/.foldercache
+      - uses: actions/setup-haskell@v1
+        with:
+          enable-stack: true
+          stack-version: 'latest'
+      - name: Install hlint
+        run: stack --resolver ${{ matrix.resolver }} install hlint
+        if: matrix.resolver != 'lts-from-exercises'
+      - name: Install hlint
+        run: 'stack --resolver $(cat exercises/*/stack.yaml | head -1 | cut -d: -f2) install hlint'
```
sshine, line 79: Since we have yq now:

```suggestion
        run: 'stack --resolver "$(yq read "$(ls -1 exercises/*/stack.yaml | head -1)" resolver)" install hlint'
```

This is the second place we do this. Perhaps a third place warrants us to place it in a fixed location.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,93 @@
...
+          # A change in any one exercise's package.yaml invalidates the cache for all other exercises.
+          # We are not sure there is a way around that or a better way to do things.
+          # Reason:
+          # By the https://github.com/actions/cache/blob/main/src/save.ts,
+          # the cache is not saved if it has already been generated.
+          # Thus the cache would slowly drift out of date unless we key on package.yaml contents.
+          key: ${{ matrix.resolver }}-${{ hashFiles('exercises/*/package.yaml') }}
+          path: |
+            ~/.stack
+            ~/.foldercache
```
sshine, line 70: Can you verify / correct my assumptions:
 - Travis currently caches all external dependencies but completely rebuilds all exercise projects,
 - The bin/test-stub script creates a build dir shared between an exercise's project and its examples,
 - We can't easily achieve this with actions/cache@v2 because the cache can't be invalidated per exercise?

What if:
 - We add the current week number ([somewhat close example](https://github.com/actions/cache#creating-a-cache-key)) to the cache key,
 - We add .stack-work/ for each exercise as part of the cache path?
   (That'd be in bin/test-stub's `$buildfolder`, I suppose?)

Then CI will only rebuild affected modules in any exercise, and we won't prove that it can be built from scratch. But any change to package.yaml, the stub, an example solution, stack.yaml, or the like would trigger a rebuild of the affected parts.

And weekly we'd get a complete rebuild in case something slipped `stack`.

Alternatively, what if we add .stack-work/ for each exercise as part of the cache part, but we run a `stack clean` before `stack build` / `stack test`?

sshine: (Side note: Doesn't `hashFiles('exercises/*/package.yaml')` refer only to exercise project files and not the ones located underneath examples/? So in case an example solution is changed, this wouldn't trigger a cache rebuild in the same way that a change in the exercise's package.yaml would.)

sshine (approved):
Wow, what an effort.

This review consists of more questions than observations.

I'm very fine with what is currently committed.

If we can reach a better solution for the cache, that'd be fine.

But it can also be improved upon independently of the main migration.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,91 @@
+name: Tests
+on:
+  pull_request:
+  push:
+    branches:
+    - gha
```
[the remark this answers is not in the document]
sshine: Disregard the comment about the `gha` branch, since it has been removed.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,93 @@
+name: Tests
+on:
+  pull_request:
+  push:
+    branches:
+    - master
+  schedule:
+    # Weekly.
+    - cron: "0 0 * * 0"
```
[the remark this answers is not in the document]
sshine: If we have a weekly cron job that fails when READMEs aren't updated, I'd like if it were separate from the one that confirms that exercise projects compile. I presume we perpetually desync from problem-specifications; eventually having this in a separate check means we could probably trigger automated PRs if that were relevant.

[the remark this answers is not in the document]
petertseng: Ah, you're absolutely right. I wouldn't want "readmes are desynced" to cause the same amount of alarm in the weekly run as "exercises stopped compiling".

For better or for worse, here is the current situation: https://github.com/exercism/haskell/blob/master/bin/ensure-readmes-are-updated.sh#L9 diffs HEAD against master. For a weekly run, HEAD is master, therefore no READMEs are checked. So currently, I consider the main thrust of this comment to be upheld. However, the auxiliary task of "we could have a separate task that checks our sync status vs problem-specifications" is not, and that could be a useful later addition.

.github/workflows/tests.yml
petertseng: it comes at the modest cost of having to install `yq` in this `job` as well (an install in an earlier `step` in the same `job` will persist for all later steps, but installs do not persist across `job`s), but that cost is acceptable for the clarity and utility that `yq` brings (and the possibility that we want to use it elsewhere in this same `job`, so ultimately I agree it is better to take your suggestion. Coming right up.

.github/workflows/tests.yml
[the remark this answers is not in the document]
petertseng: You are correct although I will need to briefly restore testing on the `gha` branch when I make changes to this PR so that I can make sure everything still works

.github/workflows/tests.yml
petertseng: > Doesn't `hashFiles('exercises/*/package.yaml')` refer only to exercise project files and not the ones located underneath examples/

Hmm, you are right. It looks like I should use `exercises/**/package.yaml` instead, then.

petertseng: > Travis currently caches all external dependencies but completely rebuilds all exercise projects,

I think that is true, depending on how complete we're talking. I'll explain more below.

> The bin/test-stub script creates a build dir shared between an exercise's project and its examples,

I think we managed to keep all of the build dirs separate, but it's been a while. I am seeing we have:
https://github.com/exercism/haskell/blob/3b9167ce1686821b65c6a99502d1622bd249a83f/bin/test-stub#L41-L43
https://github.com/exercism/haskell/blob/3b9167ce1686821b65c6a99502d1622bd249a83f/bin/test-example#L42-L44
(examplename is "stub" for the stub).
So the intent of all that is that all the examples+stub get their own separate .stack-work, but that because of the symlink they all end up in ~/.foldercache, where they'll all get cached.

> We can't easily achieve this with actions/cache@v2 because the cache can't be invalidated per exercise?

Well, this is an interesting question. For each `step` that's an instance of `actions/cache@v2`, I get to provide a list of paths but only one cache key. The one cache key is why I cannot say the cache is only valid for this exercise but still valid for these other ones. I have not experimented with whether it works to provide multiple instances of `action/cache@v2` in the same job, but I think that has its own questions:

* If we want one instance of `action/cache` per exercise, that may imply editing this workflow whenever an exercise is added? Or some other idea I haven't thought of.
* Say two exercises depend on the same external dependency - ideally it'd remain possible to only build that dependency once between those two exercises. Not sure whether that remains possible (could be, just haven't thought about it).

> What if:
> We add the current week number to the cache key,

Good idea. I think periodically having a completely clean rebuild is probably helpful. And the fact that a weekly build happens at the start of a week means that most of the time, we should get cache hits and not notice the complete rebuilds.

> We add .stack-work/ for each exercise as part of the cache path?

I think that got achieved by the symlinking above.

petertseng:
Another question I was thinking of was whether to include hlint as part of the cache - right now it's about 1-2 minutes to install hlint, which isn't too bad, but I think contributors would be happy if we reduce the time it takes before they can see useful results, so my instinct is to cache it.

I would like to include the `exercises/**/package.yaml` discussed at https://github.com/exercism/haskell/pull/930#discussion_r525644091 in this PR before merging, since I think that's essential behaviour that it's better to get right the first time rather than leaving for later. That means that cache key will change and thus I'll be going through another full rebuild on my `gha` branch, and those have been taking about 30-35 minutes, it looks like. I will probably merge after confirming that all works, and we can see whether there's anything else that can be done with the cache later.

petertseng:
I have determined via https://github.com/petertseng/gha-multi-cache-test/actions that it is indeed possible to have multiple instances of `actions/cache` within the same `job`. I'm not sure that helps us here, but at least it's good to know.

petertseng:
The summary of the changes (reminder that the word "force-pushed" links to a diff) is

* one extra newline between each step to make it easier to tell where a new one begins
* (temporary) enable runs on gha
* I realized the cache key can't use "lts-from-exercises" because then if we upgrade the LTS version, caches will collide, so I have to read the actual version from a file and use it in the cache key.
* week number added to cache key
* cache key's `exercises/*/package.yaml` now `exercises/**/package.yaml` instead
* cache hlint and don't re-install it if it's already there.

petertseng:
> I have determined via https://github.com/petertseng/gha-multi-cache-test/actions that it is indeed possible to have multiple instances of `actions/cache` within the same `job`. I'm not sure that helps us here, but at least it's good to know.

Oh, I figured out one application. Exercises' `package.yaml`s changing doesn't need to invalidate `hlint`.

petertseng:
And more changes:

* Realised that it's also necessary to check out master so that the README checker can compare against it.
* And because of the way multiple `checkout` invocations work (clears existing refs), I need to say origin/master instead of master in the README checker
* As I alluded to, I made hlint's cache separate from the rest.

petertseng:
I have done some further reading on how caches work and I believe most of my concerns have been resolved.

I'm now using restore-keys as described in https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows . At first I didn't realise that they allowed *partial* matches. That solved my concerns with "oh no, changing one exercise invalidates the entire cache" - with restore-keys, changing one exercise's package.yaml will start from the previous cache and save a new cache, as I had hoped it would. And I've kept the behaviour of keying by week number because I think it remains a good idea to make sure building from scratch works.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,143 @@
...
+
+      - uses: actions/cache@v2
+        id: cache
+        with:
+          path: |
+            ~/.stack
+            ~/.foldercache
+            ~/.local/bin/hlint
+          key: ${{ steps.resolve-resolver.outputs.resolver }}-${{ steps.week-no.outputs.week-no }}-${{ hashFiles('exercises/**/package.yaml') }}
+          restore-keys: ${{ steps.resolve-resolver.outputs.resolver }}-${{ steps.week-no.outputs.week-no }}-
```
sshine, line 125: > with restore-keys, changing one exercise's package.yaml will start from the previous cache and save a new cache

I understand this now, too. Thanks!

IIRC, when a restore-key only uses `resolver` and `week-no`, any changes that accumulate over one week without resolver updates, beyond the first change, won't get cached until the week later. This is probably acceptable.

Even though https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows has one example of

```
restore-keys: |
  npm-foobar-${{ hashFiles('package-lock.json') }}
  npm-foobar-
  npm-
```

I can't imagine how the first restore-key would work if `hashFiles` hits the new file(s) and not the old file(s). One could create an identifier based on the previous hash using `git` commands, rather than the current hash to create a chain of hashes.

sshine (approved):
>  I've kept the behaviour of keying by week number because it remains a good idea to make sure building from scratch works.

OK.

> Ran for 16 sec

We've really improved Travis build times here.

.github/workflows/tests.yml
sshine: I forgot to say: ...but I don't think we need to do that at this point.

petertseng:
So, looking at https://github.com/petertseng/exercism-haskell/actions vs https://travis-ci.org/github/exercism/haskell/branches, I think runs with warmed cache take about 5 minutes on GHA (e.g. https://github.com/petertseng/exercism-haskell/actions/runs/369630190) vs about 7 minutes on Travis (e.g. https://travis-ci.org/github/exercism/haskell/builds/743252519). I think runs from scratch take about 30 minutes on GHA (e.g. https://github.com/petertseng/exercism-haskell/actions/runs/369565614) vs about 40 minutes on Travis (e.g. https://travis-ci.org/github/exercism/haskell/builds/743551717). So definitely an improvement!

petertseng:
Even though I understand the comment about improving Travis times was probably in jest, it nevertheless remains true that GitHub Actions were an improvement anyway.

I made one whitespace change since the last approval and made a few final runs to make sure everything still works. Will look over it one last time before merging.

```diff
$ git diff -u  de0442d8e9ba99fa2061b42592527b646199fbf0 gha
diff --git a/.github/workflows/tests.yml b/.github/workflows/tests.yml
index 099d845..487d0bf 100644
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -144,7 +144,7 @@ jobs:
       - name: Check exercises
         run: |
           # Explicit set exercises' resolver only if it's not the current one.
-          if [ "${{ matrix.resolver}}" != "lts-from-exercises" ]; then
+          if [ "${{ matrix.resolver }}" != "lts-from-exercises" ]; then
             export SET_RESOLVER="--resolver ${{ matrix.resolver }}"
           fi

```

.github/workflows/tests.yml
petertseng: >  any changes that accumulate over one week without resolver updates, beyond the first change, won't get cached until the week later. This is probably acceptable.

That is also my understanding, and I too think we can accept that.

PR #931, opened 2020-11-18
ensure-readmes-are-updated: must quote $configlet

petertseng:
Otherwise, in the case where `which configlet` has empty output, then
the line `if [ ! -x $configlet ]` is results in `if [ ! -x ]`. Observe
the following:

    $ cat a.sh
    if [ ! -x ]; then
      echo "NOT EXECUTABLE"
    else
      echo "EXECUTABLE"
    fi
    $ sh a.sh
    EXECUTABLE

That means, were $configlet not quoted, configlet not being in the PATH
would mean that the script would proceed (not exit) with $configlet
being empty string. That eventually results in a "generate not found" on
line 38 (which starts with `$configlet generate`).

Compare to:

    $ cat b.sh
    if [ ! -x "" ]; then
      echo "NOT EXECUTABLE"
    else
      echo "EXECUTABLE"
    fi
    $ sh b.sh
    NOT EXECUTABLE

This is the correct behavior, so we must quote $configlet making
"$configlet".

### DIFF
--- a/bin/ensure-readmes-are-updated.sh
+++ b/bin/ensure-readmes-are-updated.sh
@@ -13,10 +13,10 @@ if git log master..HEAD | grep -q "$override_message"; then
 fi
 
 configlet=$(which configlet)
-if [ ! -x $configlet ]; then
+if [ ! -x "$configlet" ]; then
     configlet=bin/configlet
 fi
-if [ ! -x $configlet ]; then
+if [ ! -x "$configlet" ]; then
    echo "Improper configuration; configlet should exist in bin/ when this script is run"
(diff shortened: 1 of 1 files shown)


petertseng:
Had missed discovering this because I never ran this locally, but it was discovered on the move to GitHub Actions because I saw https://github.com/petertseng/exercism-haskell/runs/1416510699 says

```
./bin/ensure-readmes-are-updated.sh: 38: ./bin/ensure-readmes-are-updated.sh: generate: not found
```

sshine (approved):

Issue #952, opened 2021-02-09
Unbreak CI after v3 merge


sshine:
#936 caused CI to break on this track. [This comment](https://github.com/exercism/haskell/pull/936#pullrequestreview-579962014) highlights "Why does CI fail in this PR at the time of review?"

petertseng:
fixed in https://github.com/exercism/haskell/pull/981, thanks to the work done in https://github.com/exercism/haskell/pull/958

PR #958, opened 2021-02-26
Unbreak CI after v3 merge (addresses #952)

sshine:
After #936, CI no longer works.

This PR addresses the `master` -> `main` branch rename, as well as the fact that `configlet lint` will currently break because `config.json` does not conform to the v3 format. Rather than to make it conform, we add `continue-on-error: true` to the GitHub Action so that the GitHub Actions failures that affect exercises can be addressed.


.github/workflows/tests.yml
```diff
@@ -132,7 +132,7 @@ jobs:
             export SET_RESOLVER="--resolver ${{ matrix.resolver }}"
           fi
 
-          for exercise in ${{ github.workspace }}/exercises/*/; do
+          for exercise in ${{ github.workspace }}/exercises/practice/; do
```
Javran, line 135: I just tested locally that: `for exercise in ${{ github.workspace }}/exercises/practice/*; do` worked for me (that `*` is needed for bash expansion)

sshine:
This PR has a little scope creep, but mostly fixes CI.

Unfortunately I will not be able to finish it.

I am closing this PR to motivate that it is either reopened by someone who wants to perform the last bits, or redone.

Javran:
Thanks for all the work! I'm interested and might take a look - seems like tests are all passing - could you be more specific about what needs to be done in this last bit?

PR #968, opened 2021-05-04
🤖 Add stub files to all exercises

ErikSchierboom:
In this PR, we created (empty) stub files for all exercises that didn't yet have them.

The lack of stub file generates an unnecessary pain point within Exercism, contributing a significant proportion of support requests, making things more complex for our students, and hindering our ability to automatically run test-suites and provide automated analysis of solutions.

The original discussion for this is at exercism/discussions#238.

**We will automatically merge this PR in two week's time if it has not been merged by track maintainers at that point :slightly_smiling_face:**

## Tracking

https://github.com/exercism/v3-launch/issues/33


petertseng:
All exercises in this track already have stub files.

If they're not named what they should be, that's an interesting complaint, but the fix to that should be to rename the existing one, not create new ones. Creating a new empty one will confuse the student further because then there will be two files in the src directory and the student won't know which one is supposed to be edited.

Here is why all exercises in this track already have stub files:

The number of exercises is:

```
$ ls -1d exercises/practice/* | wc -l
101
```

The number of .hs files in a src directory is:

```
$ ls exercises/practice/*/src/*.hs | wc -l
101
```

But we need to make sure those files map 1-to-1 with exercises, rather than some exercises having multiple and some having zero. So extract the exercise name of each.

```
$ ls exercises/practice/*/src/*.hs | cut -d/ -f3 | sort | uniq | wc -l
101
```

Okay, so each existing exercise does indeed have a stub file.

ErikSchierboom:
@petertseng No idea what went wrong there. Sorry!

PR #983, opened 2021-06-10
Mark sshine as alumnus

petertseng:
Originally submitted in #962, but closed because of the deletion of fork.

It seems that the resignation is intended because the fork has been deleted and sshine left the Exercism GitHub organisation.


iHiD (approved):

iHiD:
Thank you @petertseng, and thank you @sshine for all your hard work.

Issue #912, opened 2020-04-26
Problems running stack test


BRobersonO:
I'm all the sudden having trouble running the test for my solutions to Haskell exercises. All of my old solutions work fine, but when I download a new exercise and run "stack test" I get this:

> Downloading lts-15.8 build plan ...
> RedownloadInvalidResponse Request {
>   host                 = "raw.githubusercontent.com"
>   port                 = 443
>   secure               = True
>   requestHeaders       = []
>   path                 = "/fpco/lts-haskell/master//lts-15.8.yaml"
>   queryString          = ""
>   method               = "GET"
>   proxy                = Nothing
>   rawBody              = False
>   redirectCount        = 10
>   responseTimeout      = ResponseTimeoutDefault
>   requestVersion       = HTTP/1.1
> }
>  "C:\\sr\\build-plan\\lts-15.8.yaml" (Response {responseStatus = Status {statusCode = 404, statusMessage = "Not Found"}, responseVersion = HTTP/1.1, responseHeaders = [("Content-Security-Policy","default-src 'none'; style-src 'unsafe-inline'; sandbox"),("Strict-Transport-Security","max-age=31536000"),("X-Content-Type-Options","nosniff"),("X-Frame-Options","deny"),("X-XSS-Protection","1; mode=block"),("Content-Type","text/plain; charset=utf-8"),("Via","1.1 varnish (Varnish/6.0)"),("X-GitHub-Request-Id","BBD0:46BA:4B1134:57CE23:5EA5C9A6"),("Content-Length","14"),("Accept-Ranges","bytes"),("Date","Sun, 26 Apr 2020 17:49:54 GMT"),("Via","1.1 varnish"),("Connection","keep-alive"),("X-Served-By","cache-mia17625-MIA"),("X-Cache","HIT, HIT"),("X-Cache-Hits","2, 1"),("X-Timer","S1587923394.196306,VS0,VE1"),("Vary","Authorization,Accept-Encoding"),("Access-Control-Allow-Origin","*"),("X-Fastly-Request-ID","efa65387234a2d82f547702cc50fe3a699751fbe"),("Expires","Sun, 26 Apr 2020 17:54:54 GMT"),("Source-Age","28")], responseBody = (), responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose})

Am I missing something really obvious?

BRobersonO:
Running "stack update" and trying again didn't work.

sshine:
[...]

PR #985, opened 2021-06-26
exercise_readme: add 404 error

petertseng:
Seen in:
https://github.com/exercism/haskell/issues/912
https://github.com/exercism/exercism/issues/5504

With solution reported in the latter.

This PR shows the change I would like to make to the docs, but I am no longer sure whether this is the correct place to make this change. I can't really find evidence that this file is used.


petertseng:
I put it in https://github.com/exercism/haskell/pull/997

PR #986, opened 2021-06-26
exercise_readme: explain options for upgrading

petertseng:
This PR shows the change I would like to make to the docs, but I am no longer sure whether this is the correct place to make this change. I can't really find evidence that this file is used.


petertseng:
I put it in https://github.com/exercism/haskell/pull/997

PR #990, opened 2021-07-07
tests: Use `yq eval` of yq 4.x

petertseng:
The installed yq version was changed to 4.x.
https://mikefarah.gitbook.io/yq/v/v4.x/upgrading-from-v3
We must use eval instead of read, and we must say `.resolver` before the
list of files to read, rather than `resolver` after the list of files to
read.


petertseng:
I put these in #991

PR #991, opened 2021-07-07
🤖 Simplify `hello-world` stub

ErikSchierboom:
Simplify the `hello-world` exercise's stub. The goal of this is to make things easier for students to get started and to introduce as little syntax as possible.

See https://github.com/exercism/v3-launch/issues/46

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -121,7 +121,7 @@ jobs:
           key: ${{ steps.resolve-resolver.outputs.resolver }}-${{ steps.week-no.outputs.week-no }}-${{ hashFiles('exercises/**/package.yaml') }}
           restore-keys: ${{ steps.resolve-resolver.outputs.resolver }}-${{ steps.week-no.outputs.week-no }}-
 
-      - uses: actions/setup-haskell@v1
+      - uses: haskell/actions/setup@v1
         with:
           enable-stack: true
           stack-version: 'latest'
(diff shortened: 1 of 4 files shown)


petertseng:
Interesting question  when we are in a situation where both of the following are true:

* CI is currently failing, and requisite changes need to be made to make it succeed
* there is only one active maintainer for a track

That sole maintainer cannot self-approve, but of course CI needs to be fixed to be able to continue work on the track - it's not allowed to merge PRs without successful CI, and it should never be allowed - CI should always succeed before merging PRs, without exception, and I do not _currently_ see a reason why I should be willing to compromise on this. (I could be convinced, but it would have to be very good.

This sole maintainer could piggyback any requisite changes onto others' PRs, as I have now done with this one. It's not really the best of situations though since it leaves nobody reviewing the additions of that sole maintainer.

This sole maintainer could solicit the help of https://github.com/orgs/exercism/teams/reviewers. Certainly workable for changes that are general and don't require knowledge specific to the track. I wonder what should happen for changes that do call for language-specific or track-specific judgment, though.

petertseng (approved):

ErikSchierboom:
> This sole maintainer could solicit the help of https://github.com/orgs/exercism/teams/reviewers. Certainly workable for changes that are general and don't require knowledge specific to the track. I wonder what should happen for changes that do call for language-specific or track-specific judgment, though.

The reviewers team is indeed really your only option. With v3, we'll make sure that for repos with fewer than 2 active maintainers, the reviewers team has write permissions (meaning they can approve PRs). As for the track/language-specific judgment, one can only hope that there is a reviewer familiar with the language.

Issue #895, opened 2020-03-02
crypto-square: tests do not actually verify whitespace


nicuveo:
The problem statement explains that text should be justified, and the [canonical tests](https://github.com/exercism/problem-specifications/blob/master/exercises/crypto-square/canonical-data.json) do seem to suggest that we should verify that (some "expected" strings have trailing whitespace).

However, we currently do not compare the strings: we do several tests to provide very detailed error messages... but we ignore all those whitespace issues in the process of doing so.

I would be fairly easy to fix: just one additional test that checks that the two strings are identical. But I chose to open an issue rather than just create a PR, to discuss how to best approach this.

sshine:
While non-core exercises (the ones that are often associated with a low chance of immediate or any feedback, and which I've proposed that we're able to disable feedback for in exercism/exercism#5153) don't receive a lot of attention, you are quite right.

You are welcome to submit an additional test. It would probably be a worthy addition to the canonical tests also, even though they are still locked down. So let's put a pin in that for now.

sshine:
At this point, the problem-specifications repository [has been re-opened](https://github.com/exercism/problem-specifications/issues/1686).

Before then it was always an option to just add a test in this repository only.

Would you be interested in doing either at this point, @nicuveo?

petertseng:
I see that the decision to ignore whitespace was a deliberate decision, with the reasoning in https://github.com/exercism/haskell/issues/259. However, my judgment will be that as long as our output is a String, it's better to keep it as the problem description says, for it would be against the description if a student emitted an output with a lot of leading spaces (among others). If we change the output to a [String], then we could freely choose whether to require the trailing spaces or forbid them.

petertseng:
I considered changing this in https://github.com/exercism/problem-specifications/pull/1827, but my main observation is that since this is a ciphering exercise, the focus isn't on human understandability of the output.

PR #997, opened 2021-08-01
Check whitespace in Crypto Square exercise

ashaindlin:
Fixes https://github.com/exercism/haskell/issues/895. The canonical test data itself is fine; the problem was that the test suite was ignoring whitespace when comparing the actual result to the expected result.

(It's been a few months without a reply from the creator of that issue, so I figured it would be ok to fix it myself, since I ran into this today -- apologies if not!)

### DIFF
--- a/exercises/practice/crypto-square/examples/success-standard/src/CryptoSquare.hs
+++ b/exercises/practice/crypto-square/examples/success-standard/src/CryptoSquare.hs
@@ -6,6 +6,7 @@ import Data.List.Split (chunksOf)
 
 encode :: String -> String
 encode = unwords
+       . justify
        . transpose
        . (squareSize >>= chunksOf)
        . map toLower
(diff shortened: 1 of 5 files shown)


exercises/practice/crypto-square/examples/success-standard/src/CryptoSquare.hs
```diff
@@ -6,10 +6,17 @@ import Data.List.Split (chunksOf)
...
+       . justify
        . transpose
        . (squareSize >>= chunksOf)
        . map toLower
        . filter isAlphaNum
   where
     squareSize :: String -> Int
     squareSize = ceiling . (sqrt :: Double -> Double) . fromIntegral . length
+    justify :: [String] -> [String]
+    justify [] = []
```
petertseng, line 18: I was indecisive on whether to add this. As written, this code would never pass an empty list to `justify`, so this case is unnecessary. However, I decided to add it to make `justify` potentially usable in other contexts, which seems like a fine idea. After all, the way `transpose` is used in `encode`, we would never pass an empty list to `transpose`, but `transpose` still would know how to handle one. In the same way, we can make `justify` know how to handle an empty list even if in the context of `encode` it will never need it.

petertseng (approved):
Upon review of the situation, I see that we deliberately made the choice to ignore whitespaces, but I think so long as we are outputting a String instead of a `[String]`, we should not ignore them. I accept, with thanks!

petertseng:
I considered changing this in https://github.com/exercism/problem-specifications/pull/1827, but my main observation is that since this is a ciphering exercise, the focus isn't on human understandability of the output.

ashaindlin:
Great, glad I could help! 👍

PR #1002, opened 2021-09-16
README: Fix link formatting; diamond: hlint: use tail . reverse instead of reverse . init

kukimik:


petertseng (approved):
good, thank you.

PR #1015, opened 2021-11-02
move example files under .meta

### LINKED ISSUE #1013: Properly prevent all files under `examples` from being downloaded by CLI
As just one example, try:

    exercism download --exercise=accumulate --track=haskell

Notice that among the files it writes are these:

    examples/fail-reverse/package.yaml
    examples/fail-reverse/src/Accumulate.hs
    examples/success-standard/package.yaml

We would instead hope that none of these files are written.

See how https://github.com/exercism/haskell/blob/main/exercises/practice/accumulate/.meta/config.json only lists examples/success-standard/src/Accumulate.hs as an example file. It should additionally list those three other files.

This same operation should be repeated for all exercises.
At the very least, all exercises will need this for their package.yaml.
Additionally, exercises with multiple examples will need all their examples added.
It is, of course, recommended to script this operation rather than attempt to perform it by hand.

For CI to enforce this in the future, see #1014

### LINKED ISSUE #1014: Add CI check that ensures example files are listed as such
We should make the effects of #1013 enforced by CI.

The intent is that each file under the `examples` directory should be listed as an `example` file and therefore be excluded from being downloaded by the CLI. A CI check should enforce that.

petertseng:
The aim is twofold:

* These files shouldn't be downloaded by the CLI. They currently are.
* There should be little maintenance burden; we wouldn't want to have to
  explicitly declare every example file in the config.json (we do have
  to declare a single one, but that is fine; we just don't want to have
  to declare all the rest of them).

this was accomplished with the following commands:

    for i in */; do git mv $i/examples $i/.meta; done
    sed -i'' -e 's,"examples/,".meta/examples/,' */.meta/config.json

Closes #1013: Files under
.meta will never be shown to the student
https://github.com/exercism/website/blob/c861c76b1aaff97a719e3dfb94884b65f2744017/app/models/git/exercise.rb#L210

Closes #1014: Since files
under .meta are not shown to the student, no CI check will ever be
needed.

I have no choice but to ask either @ErikSchierboom or @iHiD for a review here, because 

1. I am a solo track maintainer and there is no other reviewer to review this PR
2. https://github.com/orgs/exercism/teams/reviewers/members doesn't have write access to this repo, so their review wouldn't allow this to be merged
 
But in addition to that, you two may have an idea of whether this is the recommended way to proceed for example files (place them under .meta). It seems that if I don't place the example files under .meta I have to add every single one of them to config.json, so it's easiest to put them in .meta.


ErikSchierboom (approved):

iHiD:
Yep - this is great :)

Erik - are there any consequences you can think of of this? There aren't that I can think of as these files aren't used inside the website, the CLI or the tooling yet.

ErikSchierboom:
@iHiD Nope. Should be good

PR #1022, opened 2021-11-25
diamond: Add Text example

petertseng:
Low-effort, but it's just making sure the tests work with this type.


ErikSchierboom (approved):

petertseng:
This will be useful if someone is to try to rewrite the test suite without `string-conversions` - the rewritten test suite must still work with this solution (since the instructions suggest that the student could try to do this)

Issue #938, opened 2021-01-29
Add key features


ErikSchierboom:
_This issue is part of the migration to v3. You can read full details about the various changes [here](https://github.com/exercism/v3-launch/)._

In Exercism v3, each track must specify exactly six "key features". Exercism uses these features to highlight the most interesting, unique or "best" features of a language to a student.

Key features are specified in the top-level `"key_features"` field in the track's `config.json` file and are defined as an array of objects, as specified in [the spec](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/config-json.md#key-features).

## Goal

The `"key_features"` field in the `config.json` file should be updated to describe the six "key features" of this track. See [the spec](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/config-json.md#key-features).

## Example

```json
{
  "key_features": [
    {
      "icon": "features-oop",
      "title": "Modern",
      "content": "C# is a modern, fast-evolving language."
    },
    {
      "icon": "features-strongly-typed",
      "title": "Cross-platform",
      "content": "C# runs on almost any platform and chipset."
    },
    {
      "icon": "features-functional",
      "title": "Multi-paradigm",
      "content": "C# is primarily an object-oriented language, but also has lots of functional features."
    },
    {
      "icon": "features-lazy",
      "title": "General purpose",
      "content": "C# can be used for a wide variety of workloads, like websites, console applications, and even games."
    },
    {
      "icon": "features-declarative",
      "title": "Tooling",
      "content": "C# has excellent tooling, with linting and advanced refactoring options built-in."
    },
    {
      "icon": "features-generic",
      "title": "Documentation",
      "content": "Documentation is excellent and exhaustive, making it easy to get started with C#."
    }
  ]
}
```

## Tracking

https://github.com/exercism/v3-launch/issues/5

sshine:
So this invites for a discussion of what `key_feature`s best characterise Haskell.

The format is:

```json
{
    "title": "Adjective",
    "content": "Some catchphrase about adjective",
    "icon": "some-icon-name"
}
```

There is still no list of icons.

As inspiration, https://github.com/exercism/haskell/blob/main/docs/ABOUT.md contains a description that has a list of adjectives:

> Haskell is a general-purpose programming language known for being purely functional, non-strict with strong static typing and for having type inference.
>
[...]

PR #953, opened 2021-02-10
Add key features

sshine:
This uses the `features-functional`, `features-lazy`, and `features-strongly-typed` icons present in the example text:

https://github.com/exercism/docs/blob/main/anatomy/tracks/config-json.md#key-features

The three others, `features-type-inference`, `features-type-classes`, and ~`features-product-types`~ `features-algebraic-data-types` are proposals to have icons made.

Note that they don't exactly correspond, but close enough:

- "Purely functional" becomes `features-functional`, not exact but close.
- "Non-strict" becomes `features-lazy`, obligatory [non-strict vs. lazy](https://wiki.haskell.org/Lazy_vs._non-strict) remark
- "Strong, static typic" becomes `features-strongly-typed`, quite close
- `features-type-inference` was made up by me
- `features-type-classes` was made up by me (could have been `features-typed-overloading`, but this is more precise)
- ~`features-product-types`~ was made up by me (~could have been~ *is* `features-algebraic-data-types`)

Note that:
- These were proposed in #938.
- CI is currently broken, so perhaps wait with merging this until CI has been fixed.
- I have not assessed how much CI / configlet v3 checks the validity of this.
- When merging, this should be squash merged.

@ErikSchierboom: How does this process work?

PR #1028, opened 2021-12-08
Added key-features to config.json

hraftery:
Reviving PR #953 by @sshine, which was created to resolve issue #938: Add key features. The PR already had all six key features written but was closed before it got merged. I started where that PR ended, with the same goal, and have made the following changes:

Removed "features" prefix from icon names so they match names in the icon library.
Replaced language-specific jargon, as encouraged in the spec.
Changed "lazy" icon for the "Non-strict" feature to "evolving" to match an existing icon that vaguely conveys the concept (ie. planting a seed that will sprout later). The original pull request suggested that referring to icons that don't yet exist is okay, but I'm not confident about the outcome of that, so have erred on the side of working now, modifiable later.
Same for "strongly-typed" -> "statically-typed".
Same for "type-inference" -> "fun".
Same for "type-classes" -> "extensible".
Same for "algebraic-data-types" -> "powerful".

### DIFF
--- a/config.json
+++ b/config.json
@@ -1197,7 +1197,38 @@
     ]
   },
   "concepts": [],
-  "key_features": [],
+  "key_features": [
+    {
+      "title": "Purely functional",
+      "content": "Pure functions always produce the same output for the same input, so are more predictable.",
+      "icon": "functional"
+    },
+    {
+      "title": "Non-strict",
+      "content": "Lazy evaluation only provides results as needed, allowing for infinite data structures.",
+      "icon": "evolving"
+    },
+    {
+      "title": "Strong, static typing",
+      "content": "Strong, static types mean more errors are caught before the program is run.",
+      "icon": "statically-typed"
+    },
+    {
+      "title": "Type inference",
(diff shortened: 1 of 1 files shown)


petertseng (approved):
So! Since the previous PR had already been approved by Erik, and my review of this one in comparison with the previous one sees that this is a pure improvement to the previous (the text is more accessible to those without specific knowledge), I honestly don't see any reason to delay merging this one. So really, let's just go ahead!

petertseng:
Nice commit message too, good detail and context. I'm going to rewrap it a bit to match https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting, but I've otherwise left the content completely unchanged. I'm honestly  not sure how many people using this repo care about the line length, but I figure it can't hurt.

petertseng:
It looks like this went live on https://exercism.org/tracks/haskell/about now? It's great to have it, so thank you.

ErikSchierboom:
This is a great list!

> It looks like this went live on https://exercism.org/tracks/haskell/about now?

Yes. This updates via a webhook, so updates are almost instantaneous.

PR #1030, opened 2021-12-11
Adds concept Basics 2

pwadsworth:
Adds last three 'basic' concepts:
- Module import
- Int / Float / String conversion
- Signatures declaration
- Exercise Betty's Bike Shop to practice the concepts


pwadsworth:
Is there a way to run these checks locally or in my fork before I submit a pull request?

petertseng:
I had thought the track had taken efforts to make sure that the scripts in https://github.com/exercism/haskell/tree/main/bin were all runnable locally. Was this not the case, and if not, what happened when trying to run them?

pwadsworth:
> 



> I had thought the track had taken efforts to make sure that the scripts in https://github.com/exercism/haskell/tree/main/bin were all runnable locally. Was this not the case, and if not, what happened when trying to run them?

That might well be the case, I don't know how try to run them locally. Can you point me in the right direction? Google had too much similar information to figure out what might be the right approach.

petertseng:
Ah. Well the first thing to mention is that they are all **shell scripts**, so hopefully that keyword gives a few results. If you need to know what arguments each of those scripts takes, https://github.com/exercism/haskell/blob/main/.github/workflows/tests.yml can help. For example line 158 says that test-all-examples takes the exercise directory as its argument, so typically when I run it locally I use a command like `sh bin/test-all-examples exercises/concept/lucians-luscious-lasagna`

```
$ sh bin/test-all-examples exercises/concept/lucians-luscious-lasagna                                         
testing /home/pt/exercism/haskell/exercises/concept/lucians-luscious-lasagna/.meta/exemplar - should succeed

expectedMinutesInOven
preparationTimeInMinutes
elapsedTimeInMinutes

Finished in 0.0017 seconds
3 examples, 0 failures
```

pwadsworth:
Thanks @petertseng. I have the test working locally now.

petertseng:
I will review it when I can. December is a busy month personally. An interesting idea that might help you get pieces of your PR merged faster would be to separate out some of the changes that can be their own small PR. For example, the change to the README can be its own PR. The change of pwadsworth to pwad in lucian's lucsious lasagna could be its own PR too. That way those small PRs can get reviewed and merged quite quickly - since they're small they're easy to review, and since they're independent of this PR it makes sense to do it

PR #1031, opened 2021-12-16
 Adds concept Basics 2

pwadsworth:
Adds last three 'basic' concepts: 
- Module import Int 
- Float 
- String conversion 
- Signatures declaration
- Exercise Betty's Bike Shop to practice the concepts

Tried pulling the original PR to break it into smaller parts, but it didn't work out.


concepts/basics2/about.md
```diff
@@ -0,0 +1,105 @@
+# Haskell Syntax Basics 2
```
ErikSchierboom, line 1: Concepts should be be quite granular. There only really should be one basics concept, further concepts should be individual concepts. Currently, this documents lists several. Could you split them up into individual concepts? As such, you might find that there need to be more exercises or perhaps the exercise needs to be tweaked a bit.

concepts/basics2/about.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Exporting functions
+
+Exporting functions was covered in the first concept, here is a quick refresher.
```
ErikSchierboom, line 5: I would argue against having these "refresher" parts of the documentation, as the concept docs should really be standalone documents. If you do feel it is needed to link to a previous concept, I'd suggest using a [concept widget](https://exercism.org/docs/building/markdown/widgets#h-concept-widget) for that.

This case is slightly special I think, where re-hashing the existing information from the basics document makes sense (but without this line). In general, you should consider the concept documents to:

- Focus on just one concept
- Only contain information on that document. Linking to other concepts can be done via concept widgets

An exercise can still unlock multiple concepts, so you're not limited to just one concept in a concept exercise.

exercises/concept/bettys-bike-shop/.docs/introduction.md
```diff
@@ -0,0 +1,105 @@
+# Haskell Syntax Basics 2
```
ErikSchierboom, line 1: This introduction is quite big. As per [the docs](https://exercism.org/docs/building/tracks/concept-exercises#h-file-docs-introduction-md), the introduction document should only contain information that is needed to understand the fundamentals of the concept and solve the exercise.

ErikSchierboom: To me, it also looks like this exercise might be doing a bit too much. It is introducing: 

- Modules
- Imports
- Exports
- Ints and Float
- Strings

I think this is too much for a single exercise. There are several options to fix this issue:

1. Make this exercise solely about modules and add prerequisites for the `integers`, `floating-point-numbers` and `strings` concepts.
2. Make this exercise about modules and strings and add prerequisites for the `integers` and `floating-point-numbers` concepts.
3. Don't use floating-point-numbers in the exercise and just use integers. Normally, you'd still want to add a `integers` prerequisite. However, as the `basics` concept gives students a minimal understanding of integers, you could get away with not having that here (this is an exception though).
4. Don't use floats or integers at all and just have strings.

Some of the above options could even be combined.

ErikSchierboom (commented):
Thanks so much for working on this! I've not done an in-depth review of the text itself, but only added some higher-level comments. Let me know if you have any questions. I'm currently having a vacation, so I'll be slow to respond (might take days or even a week, depending on the circumstances).

concepts/basics2/about.md
pwadsworth: Understood. I agree it is better to cover a single concept and will use widgets to link to 'refreshers'. In this case however, since the concepts are so basic and universal, I followed the example of the Elm track introducing them together to make a more interesting module and exercise. 
I will remove that first line, add a widget and limit the discussion to the bare minimum, if any, on function export.

exercises/concept/bettys-bike-shop/.docs/introduction.md
pwadsworth: Ok, I will make the introduction for the exercises to only what is needed for the exercise, and break-up the basics-2 concept into integers, floating-point-numbers and strings. I will also break-up the exercise to use only the concepts introduced to that point, unless I find already made exercise for the Int and Float concepts.

pwadsworth:
I am closing this pull request to rework this into small bits both in the number of concepts added at one time and the pull requests. I will resubmit in multiple pull requests to make it easier to review.

ErikSchierboom:
Thanks a lot! Sorry for all the hassle. The first couple of Concept Exercises always take the longest to build.

pwadsworth:
No problem, I am learning a lot from the feedback and the concepts are looking much better now that I separated them.

PR #1032, opened 2021-12-27
Adds Types concept

pwadsworth:
Adds:
- Types concept. A basic overview of types. Future concepts will cover polymorphic types, user defined types and classes.
- Not My Type exercise for Types concept.

Edits track config.json to:
- Add entries for new concept and exercise.
- Move exercise/concepts next to concepts for easier reference when editing.


ErikSchierboom:
I'm still on vacaction. I'll review this when I'm back next week.

ErikSchierboom (commented):
Thanks for working on this! I've just read the PR and I think there are some issues with the current implementation:

1. The exercise is fairly uninteresting. This is almost entirely due to the fact that there is no story to give some context to the student on why the student should be doing what is expected of them.
2. The exercise uses a lot of different types, most of which will be unknown to the student based on the prerequisites of the exercise. This is confusing and g
3. The concept taught is hard to teach in an exercise without also introducing other concepts.

Looking at the text in the introduction, it looks like there are two things you want to teach:

1. That everything has a type
2. That there is such a thing as type annotations 

Is that correct?

pwadsworth:
Sorry for the delayed response. I've been really busy since the holidays.

To answer your questions, yes I want to teach those two points and how to write type annotations. I get the exercise is not very interesting, but the concept is really dry and I didn't come up with a semi-realistic scenario for it.

ErikSchierboom:
> To answer your questions, yes I want to teach those two points and how to write type annotations. I get the exercise is not very interesting, but the concept is really dry and I didn't come up with a semi-realistic scenario for it.

To me, that is a sign that these concepts are probably best taught together with another concept.

> In regards to issue 2 and 3, I have the numbers concept ready which talks about those types, but requires understanding the idea of types. Both of these concepts are supposed to be at the same level of the learning path since they require each other.

But the numbers concept does not need to know about chars, lists, strings or tuples, does it?

I totally understand the need to teach about types early in the curriculum. What if you'd modify the numbers exercise to also teach about types and do a very basic numbers introduction? In most tracks, the numbers exercise introduces two numeric types: the most common integer type and the most common floating point type. This should be enough to explain about types and type annotations I think.

pwadsworth:
I think this is going to become a bigger issue the further we go with additional topics. Since there is no learning path established for the track, and I don't want to be re-mixing concepts every time I make one, how about we agree on a tried-and-true order like the one followed by [Yet Another Haskell Tutorial](http://users.umiacs.umd.edu/~hal/docs/daume02yaht.pdf)? 

YAHT hits the right spot in introducing concepts fast enough not to bore people with other programing experience but new to FP, and it does it in a way that gets you to practical programs fairly soon. Also, agreeing on an order to present concepts would make it easier for others to contribute.

ErikSchierboom:
@pwadsworth Yeah I could see that working well!

PR #1052, opened 2022-03-04
workflows: use reusable workflows for configlet and sync-labels

petertseng:
individual commit messages speak

for the Haskell track, the org-wide-files sync of https://github.com/exercism/org-wide-files/issues/155 must take the form of this PR, instead of #1051 .

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -138,11 +138,6 @@
   description: "Make this PR count for hacktoberfest"
   color: "ff7518"
 
-# This label can be added to accept PRs as part of Hacktoberfest
-- name: "hacktoberfest-accepted"
-  description: "Make this PR count for hacktoberfest"
-  color: "ff7518"
-
 # This Exercism-wide label is added to all automatically created pull requests that help migrate/prepare a track for Exercism v3
 - name: "v3-migration 🤖"
   description: "Preparing for Exercism v3"
--- a/.github/workflows/configlet.yml
+++ b/.github/workflows/configlet.yml
@@ -1,0 +1,15 @@
+name: Configlet
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  configlet:
+    uses: exercism/github-actions/.github/workflows/configlet.yml@main
--- a/.github/workflows/sync-labels.yml
+++ b/.github/workflows/sync-labels.yml
@@ -2,20 +2,18 @@ name: Tools
 
 on:
   push:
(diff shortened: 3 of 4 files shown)


petertseng:
I misunderstood. Track config doesn't need to go into its own workflow! It's fine to stay in the tests workflow, because it's a separate job in that workflow. Removing that commit.

.github/labels.yml
```diff
@@ -138,11 +138,6 @@
   description: "Make this PR count for hacktoberfest"
   color: "ff7518"
 
-# This label can be added to accept PRs as part of Hacktoberfest
-- name: "hacktoberfest-accepted"
-  description: "Make this PR count for hacktoberfest"
-  color: "ff7518"
-
```
ErikSchierboom, line 145: The contents of this file are automatically generated. See https://github.com/exercism/haskell/blob/c1cfe920481669fd079e3f04b1fa4fa3c7ef66b5/.github/labels.yml#L1-L14
Removing this here will only cause it to be re-added later, due to this being in https://github.com/exercism/org-wide-files/blob/main/global-files/.github/labels.yml

petertseng: See https://github.com/exercism/org-wide-files/issues/155 which also removes these lines. I am doing nothing differently from https://github.com/exercism/org-wide-files/issues/155.

petertseng: See https://github.com/exercism/haskell/pull/1051 which also removes these lines. I am doing nothing differently from https://github.com/exercism/haskell/pull/1051.

petertseng: See https://github.com/exercism/haskell/pull/1051 which also removes these lines. This PR is exactly the same as https://github.com/exercism/haskell/pull/1051, except for two changes:

1. it splits the commits apart and adds a commit message to each (the commit message for this change is https://github.com/exercism/haskell/pull/1052/commits/bf7f280122f302c086d626a0e1424833d2a12f0a)
2. adds https://github.com/exercism/haskell/pull/1052/commits/c1cfe920481669fd079e3f04b1fa4fa3c7ef66b5 to remove the duplicate configlet run

petertseng: These lines are **duplicate lines**. Removing them is correct and in line with https://github.com/exercism/org-wide-files/blob/main/global-files/.github/labels.yml, because the latter file has those lines only **once**, whereas the Haskell track file currently has them **twice**. We must remove the duplicate instance of them.

ErikSchierboom: Aha, I missed that. Thanks for the clarification.

ErikSchierboom (approved):

PR #1068, opened 2022-05-30
word-count: Less ambiguous punctuation rule

adql:
With "ignored", it's unclear if punctuation separates the words or is just removed (relevant for the test case [handles cramped lists](https://github.com/exercism/haskell/blob/e04fc88d1027f72e37a0f140ffc6161be05051eb/exercises/practice/word-count/test/Tests.hs#L56)).


petertseng:
I see that you're right; we need to be clear about which of the two it is.

The change should be made in https://github.com/exercism/problem-specifications/blob/main/exercises/word-count/description.md first. Once it is merged there, then this track will do the same.

kotp:
It is OK for a track to diverge from Problem Specifications.  So if this is fine for Haskell Track, it could come in now.

petertseng (approved):

petertseng:
From all of us at the Haskell track, thank you for being willing to submit improvements to the track!

PR #1071, opened 2022-06-21
space-age: correct typo (algebric -> algebraic)

k4rtik:


petertseng (approved):
with the caveat that the commit message must be (and therefore will be) changed, thank you from us at the Haskell track, for improving the track!

PR #1072, opened 2022-07-01
correction to temperature test case

nicmollel:
the calculation
```
50.1 * 1.8 + 32 = 122.18
```

### DIFF
--- a/exercises/concept/temperature/test/Tests.hs
+++ b/exercises/concept/temperature/test/Tests.hs
@@ -25,7 +25,7 @@ toFCases = [ Case { description = "negative Celsius to negative Fahrenheit"
                   }
            , Case { description = "positive Celsius to positive Fahrenheit"
                   , celsius     = 50.10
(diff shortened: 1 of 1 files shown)


petertseng (changes requested):
since the instructions are very specific in saying you have to round up, if you want to change the expected answer to 122, you must also make the following changes:

* https://github.com/exercism/haskell/blob/main/exercises/concept/temperature/.docs/hints.md
* https://github.com/exercism/haskell/blob/main/exercises/concept/temperature/.docs/instructions.md
* https://github.com/exercism/haskell/blob/main/exercises/concept/temperature/.meta/exemplar/src/Temperature.hs

Personally I think it would be less work to leave it as-is... but, if there is a good reason why `round` is better educationally than `ceiling`, then it'd be worth it to make the change.

nicmollel:
i don't think `round` is better than `ceiling` here and it's me who reasoned about the problem incorrectly.

PR #1073, opened 2022-07-13
Adapt F# concepts

MatthijsBlom:
To do:

1. [x] delete irrelevant concepts
2. [x] translate remaining concepts
3. [ ] add missing concepts

### DIFF
--- a/reference/concepts.md
+++ b/reference/concepts.md
@@ -6,7 +6,7 @@ The Haskell concept exercises are based on concepts. The list below contains the
 
 ### Language-unique
 
-- Strictness vs. nonstrictenss ('laziness')
+- Strictness vs. nonstrictness ('laziness')
 - Type classes
   - `Functor`, `Applicative`, `Monad`
   - `Semigroup`, `Monoid`
--- a/reference/concepts.md
+++ b/reference/concepts.md
@@ -1,35 +1,35 @@
-# F&#35; reference
+# Haskell reference
 
 ## Concepts
 
-The F# concept exercises are based on concepts. The list below contains the concepts that have been identified for the F# language.
+The Haskell concept exercises are based on concepts. The list below contains the concepts that have been identified for the Haskell language.
 
 ### Language-unique
 
-- Active patterns
-- Computation expressions
+- ~~Active patterns~~ Pattern synonyms? That seems a stretch.
+- ~~Computation expressions~~ do-notation
 - Dependency order
 
 ### Functional
 
-- Expression-oriented (not statement oriented)
+- Expression-oriented (no statements at all)
 - Functions
   - Anonymous functions
   - Currying
   - Function composition
   - Higher-order functions
-  - Implicit returns
+  - ~~Implicit returns~~ (no returns at all)
   - Nested functions
   - Partial application
   - Pure functions
   - Recursion
     - Tail recursion
-  - Type annotations
+  - Type declarations & annotations
 - Immutability
 - Pattern matching
-  - Discards
+  - ~~Discards~~ wildcards
   - Guard clauses
-- Pipelines
+- ~~Pipelines~~ (just composition)
 - Type inference
   - Automatic generalisation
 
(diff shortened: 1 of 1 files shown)


reference/concepts.md
```diff
@@ -0,0 +1,105 @@
...
+- Values
+  - Definitions
+- Concurrency
+
+### Types
+
+- Booleans
+- Characters
+- Collections
+  - Arrays 🤔
```
MatthijsBlom, line 88: As far as I know, arrays are rare in Haskell. A multitude of alternative data structures are used instead.

MatthijsBlom:
@petertseng What do you make of this? Did I remove too much ([db5feae](https://github.com/exercism/haskell/pull/1073/commits/db5feae7d754950aea237c25247a89a30668001c))? Are the translations sensible?

ErikSchierboom (commented):
Awesome work @MatthijsBlom!

reference/concepts.md
```diff
@@ -0,0 +1,105 @@
...
+### Functional
+
+- Expression-oriented (no statements at all)
+- Functions
+  - Anonymous functions
+  - Currying
+  - Function composition
+  - Higher-order functions
+  - ~~Implicit returns~~ (no returns at all)
+  - Nested functions 🤔
```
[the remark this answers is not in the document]
MatthijsBlom: Ah, sounds like simply **local definitions** + **functions are values** then, [in-Haskell-universe](https://en.wiktionary.org/wiki/in-universe). I would be hesitant to tell new programmers that a [class distinction](https://en.wikipedia.org/wiki/First-class_citizen) _could_ exists but in fact doesn't, because it might lead to unnecessary confusion.

reference/concepts.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Concepts
+
+The Haskell concept exercises are based on concepts. The list below contains the concepts that have been identified for the Haskell language.
+
+### Language-unique
+
+- ~~Active patterns~~ Pattern synonyms? That seems a stretch.
+- ~~Computation expressions~~ do-notation
+- Dependency order
```
petertseng, line 11: could I hear what you imagine this concept will cover and/or a quick example?

reference/concepts.md
```diff
@@ -0,0 +1,105 @@
...
+  - Imports
+  - Modules
+  - Namespaces
+  - Shadowing
+  - Visibility (export or not)
+- Whitespace significant
+- String formatting 🤔
+- Values
+  - Definitions
+- Concurrency
```
petertseng, line 81: I already see this above, so this one looks like a duplicate. Thus this one can be removed.

reference/concepts.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Types
+
+- Booleans
+- Characters
+- Collections
+  - Arrays 🤔
+  - Lists
+  - Maps
+  - Queues
```
petertseng, line 91: hmm, I think we won't see many of these here, at least not natively. the only queue that comes to mind for me anyway is the queue that's composed of two lists (if I have a need to say that in fewer words in this same review, I may say Okasaki queue for short)

reference/concepts.md
```diff
@@ -0,0 +1,105 @@
...
+  - Ranges (`Enum` + range syntax)
+    - ~~Slicing~~ 🤔
+  - ~~Sequence/for expressions~~ list comprehensions
+- Comparison
+  - Equality (`Eq`)
+  - Ordering  (`Ord`, `Ordering`)
+- Concurrency 🤔 (no expertise here)
+  - Concurrent collections
+  - Locks
+  - Messaging and agents
```
petertseng, line 56: there may be a bit of a risk in leaving these here if we don't have positive confirmation that they translate to this language. How shall we deal with this risk?

petertseng (commented):
Now, I don't see any major problems. That was hopefully going to be true in any case if we were basing off of another language's known-good concepts. Here are a few things for your consideration.

reference/concepts.md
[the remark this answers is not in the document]
ErikSchierboom: Yeah I think those are the same thing when combined. Feel free to remove concepts that you feel don't apply to Haskell.

reference/concepts.md
ErikSchierboom: When in doubt, I'd lean towards removing the concept.

reference/concepts.md
```diff
@@ -0,0 +1,105 @@
...
+  - Recursion
+    - Tail recursion
+  - Type declarations & annotations
+- Immutability
+- Pattern matching
+  - ~~Discards~~ wildcards
+  - Guard clauses
+- ~~Pipelines~~ (just composition)
+- Type inference
+  - Automatic generalisation 🤔
```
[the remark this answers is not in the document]
MatthijsBlom: While I feel like this is not a great term or great way of thinking about it, in days I have not been able to come up with anything better, so I'll keep it as is.

reference/concepts.md
ErikSchierboom: I also think it doesn't

reference/concepts.md
```diff
@@ -0,0 +1,100 @@
+# Haskell reference
+
+## Concepts
+
+The Haskell concept exercises are based on concepts. The list below contains the concepts that have been identified for the Haskell language.
+
+### Language-unique
+
+- Strictness vs. nonstrictenss ('laziness')
```
ErikSchierboom, line 9:
```suggestion
- Strictness vs. nonstrictness ('laziness')
```

ErikSchierboom (approved):

ErikSchierboom (approved):
I would be happy to merge this!

@MatthijsBlom The "add missing concepts" checkbox is unchecked. Do you intend to add more concepts to the list?

MatthijsBlom:
No, it's fine by me to merge it as is.

Undoubtedly there are more concepts to add, but we do not need them now.

PR #1077, opened 2022-07-21
Corrected the expected result of "-3 plus 7 multiplied by -2" from "-8" to "-17"

dduehr:
The "wordy" test case "addition and multiplication" should expect "Just (-17)" instead of "Just (-8)".
The "*" operator has a higher precedence than "+".


dduehr:
> The "wordy" test case "addition and multiplication" should expect "Just (-17)" instead of "Just (-8)". The "*" operator has a higher precedence than "+".

Please discard this pull request. The "wordy" instructions state "... Since these are verbal word problems, evaluate the expression from left-to-right, ignoring the typical order of operations.". Sorry.

kotp:
Thanks @dduehr.

PR #1078, opened 2022-07-23
concept/numbers: Update introduction.md

lerontonge:
Improved readability of the first paragraph.

### DIFF
--- a/concepts/numbers/introduction.md
+++ b/concepts/numbers/introduction.md
@@ -1,6 +1,6 @@
 # Introduction
 
-`Integer`s hold any number no matter how big, up to the limit of your machine's memory, while `Int` corresponds to the set of positive and negative integers that can be expressed in 32 or 64 bits (guaranteed at least -2^29 to 2^29).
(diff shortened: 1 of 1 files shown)


petertseng (commented):
I'm not sure. English isn't my first language, so I have a difficult time telling the difference between "`Integer`s hold any number" and "An `Integer` holds any number". I need some help to understand what makes the latter clearer.

If the change is better, it should also be made to the following files which all have the same content. I don't like that multiple files have the same content and I have to ask contributors to make the same change in multiple places but apparently that is how it is.

* https://github.com/exercism/haskell/blob/main/exercises/concept/temperature/.docs/introduction.md
* https://github.com/exercism/haskell/blob/main/concepts/numbers/about.md

kytrinyx:
This change feels clearer to me, @petertseng. My vote is to move forward with this PR.

@lerontonge Would you have the time to make the change in the other files mentioned as well?

tofische (approved):

IsaacG:
@lerontonge Ping! Are you still interested in working on this?

Issue #1092, opened 2022-08-19
Implement Concept Exercise for pattern matching: Guessing Game


kytrinyx:
We'd like to introduce pattern matching, and we think Guessing Game is a suitable story for this:
https://exercism.org/docs/building/tracks/stories/pattern-matching-number-guessing-game

Both Elixir and F# have implemented this exercise, so it's worth taking a look at those implementations to see if there's anything there that is useful to us here.

- F#: https://github.com/exercism/fsharp/tree/main/exercises/concept/guessing-game
- Elixir: https://github.com/exercism/elixir/tree/main/exercises/concept/guessing-game

Here's a sample code snippet courtesy of @MatthijsBlom that might be suitable as a starting point.
 
```
reply 41 = SoClose
reply 42 = Correct
reply 43 = SoClose
reply n  = if n < 42 then TooLow else TooHigh
```

We will introduce pattern matching in more depth over time, so in this first exercise we'd like to explain _just enough_ to let students complete the exercise. We should also explain enough about types so that `SoClose`, `Correct` and friends aren't mystifying.

To start out, let's just get a sample solution, test suite, and stub together so that we can discuss (both in terms of Haskell and in terms of Exercism Concept Exercises).

After that in follow-on PRs we can develop the instructions (which can then be tested and discussed), and then finally the introduction.

The concept will also need to be worked on. We can do that at the same time as the introduction, or we can start on it earlier. I'm not opening an issue for it yet, but if you want to work on that drop a note in the discussion below so that we know it's being worked on.

Note that we're still working out how to best work with contributors on these concept exercises. If you have questions, please holler! We'll do our best to help figure out how to proceed.

pwadsworth:
Starting step 1 (exemplar, tests and stub). I will put it in a draft PR today.

pwadsworth:
If the solution expects use of custom types we should make that concept first. 
An alternative is to use `case` statements to introduce pattern matching and return a String like the F# track does. 

Preference?

pwadsworth:
@kytrinyx step 1 draft PR is ready. Do you want to create a separate branch to send it to, or should I use main?

pwadsworth:
Ok, I got inpatient and did the instructions and introduction before any feedback ¯\\_(ツ)_/¯
Also changed my mind about using `case` statements and refactored the exemplar to use constant patterns and guards. This aligns closer with how it was done for the F# track.

kytrinyx:
[...]

PR #1094, opened 2022-08-24
Concept/exercise/guessing-game 

pwadsworth:
### Adds:
STEP 1:
- [x] Dir structure + placeholder files
- [x] Exemplar  -- > Opted for a `case` statement solution to avoid custom types since that concept is not built yet. 
- [x] Stub
- [x] Tests

STEP 2:
- [x] Instructions
- [x] Test and discuss
- [x] Introduction

STEP 3:
- [x] Concept about + introduction, links.json
- [x] Add to haskell/config.json

PR #1095, opened 2022-08-27
Pattern Matching concept

pwadsworth:
Completes #1094 by adding Pattern Matching concept.

### DIFF
--- a/concepts/pattern-matching/about.md
+++ b/concepts/pattern-matching/about.md
@@ -1,0 +1,53 @@
+# Introduction
+
+While `if/else` expressions can be used to execute conditional logic, Haskell also has a more powerful way to execute conditional logic: pattern matching.
+With pattern matching, a value can be tested against one or more _patterns_.
+An example of such a pattern is the _constant pattern_, which matches a value against a constant (e.g. `1` or `"hello"`).
+
+When defining functions, you can define separate function bodies for different patterns.
+This leads to clean code that is simple and readable.
+You can pattern match on any data type — numbers, characters, lists, tuples, etc.
+
+For example, a trivial function that takes a whole number (`Int`) and makes it _1_ closer to _0_ could be expressed like this:
+
+```haskell
+closerToZero :: Int -> Int
+closerToZero 0 = 0
+closerToZero 1 = 0
+```
+
+Pattern matching starts to shine when used together with other patterns, for example the _variable pattern_:
+
+```haskell
+closerToZero :: Int -> Int
+closerToZero 0 = 0
+closerToZero n = n - 1
+```
+
+The above example treats all inputs other than _0_ the same, and would produce incorrect results for negative numbers.
+This can be solved using conditional patterns, known as _guards_, which are expressed with the `|` symbol:
+
+```haskell
+closerToZero :: Int -> Int
+closerToZero n
+    | n < 0 = n + 1
+    | n > 0 = n - 1
+```
+
+In the above examples not all possible inputs have a matching pattern.
+The compiler will detect this and output a warning.
+This is a very useful feature of Haskell that helps ensure all possible paths are covered to avoid run-time errors.
+It is known as _exhaustive checking_.
+To solve the warning, you have to handle all cases.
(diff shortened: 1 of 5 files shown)


concepts/pattern-matching/about.md
```diff
@@ -0,0 +1,52 @@
...
+describe 0 = "Zero"
+describe 1 = "One"
+```
+
+Pattern matching starts to shine when used together with other patterns, for example the _variable pattern_:
+
+```haskell
+describe :: Int -> String
+describe 0 = "Zero"
+describe n = "Not zero"
```
petertseng, line 24: I wonder if it would be better to use an example that could make use of the `n` here. Is that not important?

MatthijsBlom: Not a direct answer to your question, but: we're still talking about it on Slack.

pwadsworth: The only way I can think of using the `n` is something like:
```
describe n = show n ++ " is not zero
```
But we haven't introduced `show` or `++` yet. I think not using the variable is preferable to using functions not previously introduced. Also, that example is an intermediate step to the next one where `n` is used for the conditional. Thoughts?

ErikSchierboom: I agree with @pwadsworth's reasoning here. I had the same thought as @petertseng initially, but then I read on and saw that it was used in the next example, so I'm fine with it. And yes, using `show` or `++` is not advisable.

petertseng: Given that `show` and `++` are inappropriate, that implies that the result of the function should be something other than String, rather than discarding the idea of using the `n` entirely!

For example, one could present a function that brings a number closer to zero by subtracting 1 from it, if it isn't already zero:

```haskell
closerToZero 0 = 0
closerToZero 1 = 0
closerToZero 2 = 1
```

and then, one could introduce the ability to bind, replacing the 1 and 2 cases

```haskell
closerToZero 0 = 0
closerToZero n = n - 1
```

Then, guards could be used to say that negative numbers should instead have 1 added to them, because that's the way to bring them closer to zero.

This doesn't necessarily have to be the idea that is actually used, but you must know that it is my duty to ensure that the idea is considered, and when I see no evidence that the idea has been considered I am required to bring it up.

pwadsworth: I can get onboard with that. It makes use of the variable and doesn't distract from the point of patter matching. @ErikSchierboom / @kytrinyx  do you agree ?

kytrinyx: I really like the `closerToZero` suggestion. I'd say ship it!

pwadsworth:
This is ready to merge with #1094 to close #1092

kytrinyx:
@petertseng I'm thinking we can merge this to main, and then rebase the guessing game PR onto main in order to get the checks passing. Does that seem like a good approach to you?

petertseng (approved):
about.md was changed to use `closerToZero`. shouldn't we use `closerToZero` in introduction.md too, then? I made that happen. What should be the difference in about.md and introduction.md? If they have the same content, how about symlinking them? That way nobody will forget to make a change in one place and not the other. We are having this problem in other pull requests too, where I have to ask well-meaning contributors to do extra work to update the same line in three different places and it is causing friction for them.

As you can see, I squashed commits.

If there are no objections, let's go.

kytrinyx:
> If they have the same content, how about symlinking them? That way nobody will forget to make a change in one place and not the other. We are having this problem in other pull requests too, where I have to ask well-meaning contributors to do extra work to update the same line in three different places and it is causing friction for them.

I'm pretty sure the answer to this is to use a `.docs/introduction.md.tpl` file and generate the introduction.md per the instructions here: https://exercism.org/docs/building/tracks/concept-exercises

I could take a stab at doing this next week (I'll be away from the computer for a few days starting tomorrow).

petertseng:
It sounds like there are no objections to merging this. The ideas to save contributors the headache of forgetting to make a change in a necessary place can be applied later. I look forward to those.

Issue #1021, opened 2021-11-25
proverb: add missing "List here the packages you want to use in your solution"


petertseng:
All exercises' package.yaml have lines under `library` section:

```yaml
  # dependencies:
  # - foo       # List here the packages you
  # - bar       # want to use in your solution.
```

such as in https://github.com/exercism/haskell/blob/main/exercises/practice/accumulate/package.yaml.

However, https://github.com/exercism/haskell/blob/main/exercises/practice/proverb/package.yaml does not.

To fix this issue, add those lines to `proverb`'s package.yaml.

PR #1097, opened 2022-08-27
proverb: add commented dependencies section to package.yaml

pwadsworth:
Adds
   
```
# dependencies:
  # - foo       # List here the packages you
  # - bar       # want to use in your solution.
```

to package.yaml to close issue #1021


petertseng (approved):
good. please note proper PR title and first line of commit message should say what was changed. It can't just say the issue number because then those without an internet connection don't know what the issue is, and it can't just say the file that was changed because that information is already available via `log --stat`

PR #1100, opened 2022-08-28
ocr-numbers: fix spelling mistake in Tests.hs

tobeannouncd:
Minor spelling fix


petertseng (approved):
ah, indeed. Thank you. I have modified the PR title (and will update the commit message when merging) to be informative, which commit messages should be, but obviously the change to be made is sound, and we are grateful for it

PR #1101, opened 2022-08-29
Update installation instructions to use GHCup

MatthijsBlom:
These days the recommended installation method is to use GHCup.

The previous instructions (based on Stack) were not wrong, but the GHCup method is more general and should make it easier for students to do Haskell stuff outside of Exercism. I don't think the instructions are much more complicated.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,65 +1,82 @@
 # Installation
 
-## Installing Stack
 
-If you don't have a recent Stack version installed in your system, follow the
-[installation instructions](http://docs.haskellstack.org/en/stable/install_and_upgrade/)
-for your operating system. After completing the installation, test that you
-have installed it correctly by running this command in a terminal:
+## Get GHCup
 
-```bash
-stack --version
-```
+The recommended way of installing Haskell tools is through [GHCup][ghcup].
 
-If it outputs a Stack version equal to or greater than 1.1.2, then you have it
-installed correctly.
+With GHCup you can manage your installed compilers, project/package managers, and language server.
 
-## Installing GHC
 
-Stack will automatically download and install GHC (the Glasgow Haskell
-compiler) when it is required. You can start the installation and test
+## Install Stack
+
+For doing Exercism exercises you only need Stack.
+
+To install Stack, simply run the GHCup installation command or script and follow the defaults but specify that you would like to have Stack installed as well.
+
+Did you miss the opportunity? No worries: just run `ghcup install stack` afterwards.
+
+GHCup will install
+
+- **A compiler** (GHC: the Glasgow Haskell Compiler)
+  
+  Strictly speaking you do not need this for Exercism, but it comes in handy when you want to play with GHCi outside of exercise directories.
+
+- **Cabal**
+
+  A project/package manager. You do not need this for Exercism.
+
+- **Stack**
+
+  A project/package manager that you do need for Exercism. It will make sure you have the right compiler and libraries for each individual project (exercise).
+  
+To check that installation was successful, run the command `ghcup list` and/or `stack --version`.
+
+
+## Get the right GHC version
+
+Stack will automatically download and install GHC when it is required. You can start the installation and test
 by running the following command in a terminal:
 
 ```bash
-stack ghc
+# move into exercise directory
+cd /exercism/haskell/hello-world
+# start interactive environment
+stack ghc -- --version
 ```
 
-If this outputs something like `ghc: no input files`, GHC is installed
-correctly.
+If this outputs something like
 
-Note that there are other ways to install GHC on a system (like the
-[Haskell platform](https://www.haskell.org/platform/)), but for
-running the Exercism tests, a regular stack installation is most
-likely to work correctly.
-
-## Installing HLint (optional)
-
-HLint is a tool for suggesting possible improvements to Haskell code. These
-suggestions include ideas such as using alternative functions, simplifying
-code and spotting redundancies.
-
-You can use Stack to install HLint:
-
-```bash
-stack install hlint
+```
+The Glorious Glasgow Haskell Compilation System, version 8.8.4
 ```
 
-Check that it was installed correctly with:
+then GHC is installed correctly.
 
-```bash
-hlint --version
-```
 
-If you get the version in your terminal, you're done, otherwise, you
-probably need to add a missing directory to your path:
+## Set up your editor to work with Haskell
 
-- `%APPDATA%\local\bin`, if you are running *Windows*
-- `$HOME/.local/bin`, for other operating systems.
+Haskell support is provided by the Haskell Language Server (HLS).
 
-## Haskell IDEs and Plugins
 
-Using an <abbr title="Integrated Development Environment">IDE</abbr> can make it easier to factor your Haskell programs. There are several free and commercial IDEs available
-- [`atom-ide-haskell`](https://atom-haskell.github.io/) for Github's Atom editor
-- There are [a number of Haskell packages](https://medium.com/@dogwith1eye/setting-up-haskell-in-vs-code-on-macos-d2cc1ce9f60a) for Microsoft's Visual Studio Code editor.
-- [IntelliJ Haskell Plugin](https://github.com/rikvdkleij/intellij-haskell)
-- [Other Plugins](https://wiki.haskell.org/IDEs) referenced on the Haskell Wiki.
+### Visual Studio Code
+
+Just get [this extension][vscode-haskell]. It will use GHCup to install HLS automatically, after which you'll be all set.
+
+**Note:** with freshly downloaded exercises, HLS may have trouble starting up. If this happens, just run `stack test` and restart the language server.
+
+
+### Other editors
+
+See the [HLS documentation on editor configuration][hls-editor-config].
+
+
+[ghcup]:
+    https://www.haskell.org/ghcup/
+    "Get GHCup"
+[hls-editor-config]: 
+    https://haskell-language-server.readthedocs.io/en/latest/configuration.html#configuring-your-editor 
+    "Editor configuration for Haskell Language Server"
+[vscode-haskell]: 
+    https://marketplace.visualstudio.com/items?itemName=haskell.haskell 
+    "Haskell extension for VS Code"


docs/INSTALLATION.md
```diff
@@ -1,65 +1,82 @@
...
+
+
+## Set up your editor to work with Haskell
+
+Haskell support is provided by the Haskell Language Server (HLS).
+
+
+### Visual Studio Code
+
+Just get [this extension][vscode-haskell]. It will use GHCup to install HLS automatically, after which you'll be all set.
```
petertseng, line 64:
```suggestion
Just get the [Haskell for Visual Studio Code extension][vscode-haskell]. It will use GHCup to install HLS automatically, after which you'll be all set.
```

By my reading, a link like "this extension" is like "click here", which I was taught by https://webaccess.berkeley.edu/ask-pecan/click-here to avoid, so I have decided to suggest this.

docs/INSTALLATION.md
```diff
@@ -1,65 +1,82 @@
...
-```bash
-stack ghc
-```
+For doing Exercism exercises you only need Stack.
+
+To install Stack, simply run the GHCup installation command or script and follow the defaults but specify that you would like to have Stack installed as well.
 
-If this outputs something like `ghc: no input files`, GHC is installed
-correctly.
+Did you miss the opportunity? No worries: just run `ghcup install stack` afterwards.
```
petertseng, line 17: I would prefer one sentence per line. With one sentence per line diffs will be more appropriately scoped when only one sentence is changed.

docs/INSTALLATION.md
```diff
@@ -1,65 +1,82 @@
...
+**Note:** with freshly downloaded exercises, HLS may have trouble starting up. If this happens, just run `stack test` and restart the language server.
+
 
-If you get the version in your terminal, you're done, otherwise, you
-probably need to add a missing directory to your path:
+### Other editors
 
-- `%APPDATA%\local\bin`, if you are running *Windows*
-- `$HOME/.local/bin`, for other operating systems.
+See the [HLS documentation on editor configuration][hls-editor-config].
```
petertseng, line 71: I'm conflicted. On the one hand, the student may get confused about "what is Haskell Language Server and do I need this?". See how a lot of the install instructions also include instructions to install a language server protocol implementation, which is not in line with the stated purpose of this link (other editors). On the other hand, I wanted to suggest https://wiki.haskell.org/Editors, but that provides no commentary about each editor. Thus I don't think either of these suggestions is perfect. I would like to figure out which will be most helpful to the students.

petertseng (commented):
although I am a dinosaur who didn't keep up with the times and didn't know about GHCup, it's clear that GHCup gained recognition because GHCup made it to https://www.haskell.org that it has gained recognition, so that is good.

docs/INSTALLATION.md
```diff
@@ -1,65 +1,82 @@
...
+- **Cabal**
 
-You can use Stack to install HLint:
+  A project/package manager. You do not need this for Exercism.
+
+- **Stack**
+
+  A project/package manager that you do need for Exercism. It will make sure you have the right compiler and libraries for each individual project (exercise).
+  
+To check that installation was successful, run the command `ghcup list` and/or `stack --version`.
```
petertseng, line 33: Is it self-evident what the student should be looking for as a result of the `ghcup list` and/or `stack --version` commands? If it is not, we will want to explain it here

docs/INSTALLATION.md
```diff
@@ -1,65 +1,82 @@
...
 ```
+The Glorious Glasgow Haskell Compilation System, version 8.8.4
+```
+
+then GHC is installed correctly.
+
+
+## Set up your editor to work with Haskell
+
+Haskell support is provided by the Haskell Language Server (HLS).
```
petertseng, line 59: I'm just a dinosaur who never learned about this language server. But, I would not want to imply that you need the language server to have an editor that can work with Haskell.

petertseng, line 59: I'm just a dinosaur who never learned about this language server. But, I would not want to imply that you need the language server to have an editor that can work with Haskell, which "support is provided by..." seems to imply. It is okay if the docs *recommend* using the language server if it materially improves student experience, but I want it to be clear that it is a recommendation, not a requirement.

docs/INSTALLATION.md
MatthijsBlom: Throughout all of the instructions, we already implicitly assume that the student is familiar with the command line. So for `stack --version` I think we're good. As for `ghcup list`&hellip; see screenshot below. I do not know what failure might look like (other than the usual _command not found_).

<details><summary>Screenshot of <code>ghcup list</code></summary>

image
</details>

If you still think we need to be more explicit, I'll add explanations.

petertseng: It appears I need remember better what sort of background knowledge we assume for those who are joining a track. As you can see, that greatly affects what sort of detail we need to provide. In light of the assumption you've now informed me of, I now see that we can leave it as is.

petertseng (approved):
I think we're ready.

docs/INSTALLATION.md
```diff
@@ -1,65 +1,90 @@
...
+
 
-If you get the version in your terminal, you're done, otherwise, you
-probably need to add a missing directory to your path:
+### Other editors
 
-- `%APPDATA%\local\bin`, if you are running *Windows*
-- `$HOME/.local/bin`, for other operating systems.
+See the [HLS documentation on editor configuration][hls-editor-config].
+You might need to install HSL manually.
```
petertseng, line 78:
```suggestion
You might need to install HLS manually.
```

Of course, HSL would be hue-sat-lum, the notation for specifying colours. We need HLS (Haskell Language Server) here

docs/INSTALLATION.md
```diff
@@ -1,65 +1,90 @@
...
-compiler) when it is required. You can start the installation and test
-by running the following command in a terminal:
+## Install Stack
 
-```bash
-stack ghc
-```
+For doing Exercism exercises you only need Stack.
+
+To install Stack, simply run the GHCup installation command or script and follow the defaults but specify that you would like to have Stack installed as well.
```
kotp, line 15:
```suggestion
To install Stack, run the GHCup installation command or script and follow the defaults but specify that you would like to have Stack installed as well.
```

This is a complex sentence and "simply" is not implied, as we have to make decisions here.  Often "simply" is insulting to see when and if we have a problem.  We can avoid the "insult" by removing this one word.  Otherwise it does encourage one to believe that they have not understood the "simple" and are somehow the problem.

docs/INSTALLATION.md
```diff
@@ -1,65 +1,90 @@
...
-[Haskell platform](https://www.haskell.org/platform/)), but for
-running the Exercism tests, a regular stack installation is most
-likely to work correctly.
+To check that installation was successful, run the command `ghcup list` and/or `stack --version`.
 
-## Installing HLint (optional)
 
-HLint is a tool for suggesting possible improvements to Haskell code. These
-suggestions include ideas such as using alternative functions, simplifying
-code and spotting redundancies.
```
petertseng, line 39: Discuss about the removal of HLint install instructions. Is it implicitly included with the editor support/Haskell Language Server? Is it deemed no longer helpful to students?

docs/INSTALLATION.md
```diff
@@ -1,65 +1,90 @@
...
+## Install Stack
 
-```bash
-stack ghc
-```
+For doing Exercism exercises you only need Stack.
+
+To install Stack, simply run the GHCup installation command or script and follow the defaults but specify that you would like to have Stack installed as well.
+
+Did you miss the opportunity?
```
kotp, line 17: The opportunity for what?

docs/INSTALLATION.md
```diff
@@ -1,65 +1,90 @@
...
+GHCup will install
+
+- **A compiler** (GHC: the Glasgow Haskell Compiler)
+
+  Strictly speaking you do not need this for Exercism, but it comes in handy when you want to play with GHCi outside of exercise directories.
+
+- **Cabal**
+
+  A project/package manager.
+  You do not need this for Exercism.
```
kotp, line 29: Then do we need to mention this section?

docs/INSTALLATION.md
```diff
@@ -1,65 +1,90 @@
...
+To install Stack, simply run the GHCup installation command or script and follow the defaults but specify that you would like to have Stack installed as well.
+
+Did you miss the opportunity?
+No worries: just run `ghcup install stack` afterwards.
+
+GHCup will install
+
+- **A compiler** (GHC: the Glasgow Haskell Compiler)
+
+  Strictly speaking you do not need this for Exercism, but it comes in handy when you want to play with GHCi outside of exercise directories.
```
kotp, line 24: Then, we probably should not mention it here for installation for Exercism.

Maybe point to resources about "learning more" elsewhere?

docs/INSTALLATION.md
```diff
@@ -1,65 +1,90 @@
...
 
-If you get the version in your terminal, you're done, otherwise, you
-probably need to add a missing directory to your path:
+### Other editors
 
-- `%APPDATA%\local\bin`, if you are running *Windows*
-- `$HOME/.local/bin`, for other operating systems.
+See the [HLS documentation on editor configuration][hls-editor-config].
+You might need to install HSL manually.
+To do so, simply run `ghcup install hls`.
```
kotp, line 79:
```suggestion
To do so, run `ghcup install hls`.
```

For the same reasons mentioned earlier.  Placing this one word makes one feel like they can not do the simple things, and can make one reluctant to ask for help, because it is already indicated that this is simple, so maybe if they can not understand the simple things, they have no business doing the more complicated.

docs/INSTALLATION.md
```diff
@@ -1,65 +1,90 @@
...
-If you don't have a recent Stack version installed in your system, follow the
-[installation instructions](http://docs.haskellstack.org/en/stable/install_and_upgrade/)
-for your operating system. After completing the installation, test that you
-have installed it correctly by running this command in a terminal:
+## Get GHCup
 
-```bash
-stack --version
-```
+The recommended way of installing Haskell tools is through [GHCup][ghcup].
```
kotp, line 6: We might also reference where this information comes from, the recommendation stated at https://www.haskell.org/downloads/

This can mean that when updates and changes happen, we are less likely to be outdated.

docs/INSTALLATION.md
```diff
@@ -1,65 +1,90 @@
...
+- **Cabal**
+
+  A project/package manager.
+  You do not need this for Exercism.
+
+- **Stack**
 
-If this outputs something like `ghc: no input files`, GHC is installed
-correctly.
+  A project/package manager that you do need for Exercism.
```
kotp, line 33: By being here, implies that we do need it, I think, if we do not mention things that we do not need, in this document.  I am not sure that this instills confidence in other places if we mention it is needed, and do not mention explicitly the need in other sections.

It is easier to state what is needed, and to state and then declare what is not needed.

Or perhaps put optional information in a complete section for those things that are nice but not necessary?

kotp (changes requested):
Some things I noted while reading the instructions for installation.

MatthijsBlom:
@petertseng I made a few changes in response to feedback and tests. I think it is now ready to be merged.

@kotp, Johannes and Kevin: thanks for the testing and feedback!

kotp (approved):

docs/INSTALLATION.md
kotp: I approved, even though I do not like the "simply run" combination.  It should not block.

docs/INSTALLATION.md
kotp: Still pending…

docs/INSTALLATION.md
kotp: Should we move the "not needed" things to a "Recommended" section if it makes things nicer, but is not necessary?

docs/INSTALLATION.md
petertseng: other words to watch out for are "just", as in someone suggesting "why don't you just _____?"

PR #1104, opened 2022-08-31
Link to updated fork of Learn You A Haskell

MatthijsBlom:
The original [learnyouahaskell.com](http://learnyouahaskell.com/) has become a bit outdated, and moreover had some long standing issues. Therefore, an updated fork is maintained at [learnyouahaskell.github.io](https://learnyouahaskell.github.io/).

I changed every link to the original LYAH in the repo to point to the updated fork instead.

### DIFF
--- a/concepts/basics/about.md
+++ b/concepts/basics/about.md
@@ -65,7 +65,7 @@ So we'll start with this.
 module Calculator ( add ) where
 ```
 
-http://learnyouahaskell.com/modules#making-our-own-modules
+http://learnyouahaskell.github.io/modules#making-our-own-modules
 
 ## Comments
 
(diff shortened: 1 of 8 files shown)


concepts/basics/about.md
```diff
@@ -65,7 +65,7 @@ So we'll start with this.
 module Calculator ( add ) where
 ```
 
-http://learnyouahaskell.com/modules#making-our-own-modules
+http://learnyouahaskell.github.io/modules#making-our-own-modules
```
petertseng, line 68: consider making all http links https instead. When I attempt to visit the link using http, I am redirected to https already, so we might as well save the redirect.

petertseng (approved):
The discussion at https://www.reddit.com/r/haskell/comments/sogi3s/learn_you_a_haskell_a_community_version/ and list of commits at https://github.com/learnyouahaskell/learnyouahaskell.github.io/commits/main look agreeable. Not everything that needs to be updated has been updated yet, but it is in progress and it is good to get started on linking to this one.

concepts/basics/about.md
MatthijsBlom: I did this by accident but reverted it. Should I replace all `http`s with `https`, even in non-LYAH urls?

PR #1105, opened 2022-08-31
Use https everywhere

MatthijsBlom:
This commit replaces all occurrences of `http:` with `https:`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ When changing a test suite, the version number should be updated appropriately s
 This versioning policy was proposed and accepted in https://github.com/exercism/haskell/issues/522.
 
 ## Development Dependencies
-You should have [Stack](http://docs.haskellstack.org/) installed in your system to make contributing to this repository easier.
+You should have [Stack](https://docs.haskellstack.org/) installed in your system to make contributing to this repository easier.
 
 ## Stub solution
 The stub solution should be as general as possible in order to not exclude any possible solutions. It should take Haskell specifics into account (for example use `Maybe` instead of a dummy return value). It should not contain any comments (people might forget to remove them), you can use the hints file instead.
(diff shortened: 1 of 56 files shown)


petertseng:
My apologies for sounding negative. I realised that it's my fault for giving a bad instruction. I wanted to say that only learnyouahaskell links should be changed, instead of all http links, but I accidentally said something that made it sound like I wanted all links changed. That is my fault. Well, I won't mind continuing down this path to change as many http links to https as possible, but as you can see from the many URLs that have problems, care will be required to check which ones are possible. I'll leave it to your judgment how you want to proceed with it. If you want to keep doing all the possible ones, that is okay, or if you only want to do certain domains, that's okay too.

MatthijsBlom:
No offence taken. I learned something, and slept just fine 🙂

All urls in the new commit have been tested automatically; all return 200 except for

- `https://www.wolframalpha.com/examples/NumberBases.html`, but the old one does not work anymore either. I kept it in for consistency with the other WolframAlpha urls.
- `https://papyr.com/hypertextbooks/grammar/ph_noun.htm` gave me 403 but works in the browser.

petertseng (approved):
Excellent. I agree with all assessments (that all URLs return 200s, with the exception of the Wolfram Alpha one that was also 404 on HTTP as well). Let's go.

My way of testing this one was fairly uninspiring (I mention this in case I can be taught about a better way of testing it). I did `git show 4cab0d6396d2b153ea432593bddf2b587ad22277 | grep '^+' | grep -o 'https.*' | sort | uniq`. I probably could have done a little more cleaning by removing a trailing parenthesis or quote, but it was good enough to test the URLs which was the important part.

MatthijsBlom:
I did that basically by hand, and then

```python
# Python
import requests

for url in URLS:
    if requests.get(url).status_code != 200:
        print(url)
```

Issue #1103, opened 2022-08-31
[Learning Mode] Discussion: building up to Pattern Matching


kytrinyx:
@MatthijsBlom suggested the following concept graph, which is a good starting point for building up to a basic understanding of proper pattern matching in Haskell.

```
              |
   Simple Pattern Matching      
   /          |        \
Maybe      Either      ... (tuples, lists?)
   \          |        /
    Algebraic Data Types
      ("custom types")
              |
   Pattern Matching Proper
              |
```

I'd like to start the discussion by looking at some sample code that demonstrates the end point, _Pattern Matching Proper_. It doesn't have to relate to any exercise or story, rather it's a way to make sure that everyone involved understands where we're trying to get to.

@MatthijsBlom I know you've been thinking a lot about this over the past couple of weeks. Would you be willing to write a few (maybe 10 or so?) lines of code to illustrate _Pattern Matching_?

MatthijsBlom:
The solution to Valentine's is feature-complete, so to speak. It matches on a custom ADT (`Activity`), matches on fields (e.g. `Korean`), uses binding patterns (not sure this is the right terminology) (`kilometers`), and it uses guards. All this is very typical of Haskell.

Here are a few variations of a translation of the F# exemplar solution.

<details>
<summary>
Show code
</summary>

```haskell
{-# LANGUAGE LambdaCase #-}
-- The above enables a well-established language extension

module ValentinesDay (rateActivity) where

data Approval = Yes | No | Maybe

data Cuisine = Korean | Turkish

data Genre = Crime | Horror | Romance | Thriller

data Activity = BoardGame | Chill | Movie Genre | Restaurant Cuisine | Walk Integer

-- How I would write it, but requires LambdaCase extension
rateActivity :: Activity -> Approval
rateActivity = \case
  Restaurant Korean -> Yes
  Restaurant Turkish -> Maybe
  Movie Romance -> Yes
  Walk kilometers
    | kilometers < 3 -> Yes
    | kilometers < 5 -> Maybe
  _ -> No

-- Without LambdaCase
rateActivity' :: Activity -> Approval
rateActivity' activity =
  case activity of
    Restaurant Korean -> Yes
    Restaurant Turkish -> Maybe
    Movie Romance -> Yes
    Walk kilometers
      | kilometers < 3 -> Yes
      | kilometers < 5 -> Maybe
    _ -> No

-- Using function definition syntactic sugar
rateActivity'' :: Activity -> Approval
[...]

PR #1106, opened 2022-08-31
Add concept: Maybe

MatthijsBlom:
This is my WIP w.r.t. #1103.

### DIFF
--- a/concepts/maybe/introduction.md
+++ b/concepts/maybe/introduction.md
@@ -1,0 +1,104 @@
+# Introduction
+
+Consider the following two scenarios.
+
+-   A web shop maintains a database of customers.
+    All customers have an e-mail address registered.
+    Also, most have a known postal address.
+    However, some customers do not.
+    How should this possibly-unknown postal address be represented in the database?
+
+-   Eager to become rich, you write a program for scouring Donald Knuth's books for mistakes, so that you may collect [the bounties][wikipedia-knuth-reward-check].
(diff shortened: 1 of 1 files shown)


MatthijsBlom:
@kytrinyx @ErikSchierboom Before I continue with the last bits of `Maybe` and then `Either` and tuples: what do you think of this style and depth?

kytrinyx:
@MatthijsBlom Thank you so much for taking the time to think about this.

I think there's a lot of good stuff in here, but I'm unable to say anything concrete about it until we know what the exercise looks like.

There was a conversation in the Slack a couple of weeks ago which might have gotten lost, as there were about a dozen conversations happening in different places all at once.

Peter Wadsworth asked:

> What process should be followed for developing concepts?

And my response was that there are four steps.

1. Identifying the right concepts to teach at the current point in the concept tree
2. Figuring out the piece of code that will be the basis for the exercise
3. Fleshing out the code part of the exercise
4. Working out the documentation (exercise introduction, instructions, concept introduction)

It is really important that we complete each one before moving to the next.

At the moment we've identified `Maybe` (and `Either` and `tuples`) as concepts we want to tackle (step 1).

In order to avoid rework and wasted effort we need to spend some time on step 2 for all of these concepts: looking at a handful of different code examples that use the concept in simple scenarios. It's useful to have several examples rather than just one, so that we can narrow in on what feels like the right size, shape, and conceptual complexity of the target exercise.

I hope this makes sense. Once we have the code example nailed down it will be much easier to talk about this PR in depth!

PR #630, opened 2017-12-08
Implementing protein-translation

Average-user:
I wanted to use  ```chunksOf``` from [Data.List.Split](https://hackage.haskell.org/package/split-0.2.3.2/docs/Data-List-Split.html#v:chunksOf) but apparently it is not in the cabal configuration

PR #1107, opened 2022-09-06
protein-translation: Add tests for invalid sequences

mhorod:
The result type of the `proteins` function is `Maybe [String]` but all tests expect `Just`.
I added some tests that expect `Nothing` on an invalid input.


petertseng (commented):
hmm, there wasn't discussion in https://github.com/exercism/haskell/pull/630 of why these weren't added, but that must just be because they didn't exist in https://github.com/exercism/problem-specifications/blob/main/exercises/protein-translation/canonical-data.json then. but now that some tests do, we should better use those instead of making up our own

PR #1109, opened 2022-09-11
Concept exercise guessing-game (new)

### LINKED ISSUE #1092: Implement Concept Exercise for pattern matching: Guessing Game

pwadsworth:
Adds concept exercise Guessing Game for concept Pattern Matching.
Closes #1092 and fixes my mess in PR #1094


petertseng:
In this case, it is not a problem, but there are some times where it won't be appropriate to open a new pull request. These are cases where there is significant discussion in the previous pull request and further discussion is required in the new one. Opening a new pull request would split the discussion into two places, which means it is not possible to follow the discussion. That is not the case here, so we are okay.

But this pull request has demonstrated that you were able to create a branch that contains just the two commits you want merged, which is good! In that case, the course of action would to push those two commits to the `guessing-game` branch, the branch being used in https://github.com/exercism/haskell/pull/1094, not to open a new pull request. For example, on the `guessing-game` branch, `git reset --hard 422f001ff84537ca09e357f57f773cdf328b125c`, because https://github.com/exercism/haskell/pull/1109/commits/422f001ff84537ca09e357f57f773cdf328b125c is the head of this pull request's branch.

Finally, I will note that I need to push some commits to this `main` branch, which is the branch being used in this pull request. Please be aware of that if you are going to use this `main` branch in the future.

petertseng (approved):
Okay, I think we're good to go. Thanks for working on it and enduring my numerous requests to change things in the code and in git.

Others have about 24 hour to object, and if not, this is getting merged.

MatthijsBlom:
The exercise is now visible to students, as is the accompanying concept. However, the syllabus is still hidden. Is this the preferred state?

ErikSchierboom:
> The exercise is now visible to students, as is the accompanying concept. However, the syllabus is still hidden. Is this the preferred state?

@MatthijsBlom I think this is actually a bug. Could you perhaps create an issue at https://github.com/exercism/exercism?

ErikSchierboom:
Great work all!

PR #1113, opened 2022-09-26
Update CollatzConjecture.hs

spolisar:
Added n to collatz stub


petertseng (approved):
hmm yes, we should have done this from the start, as other stubs do that. Thank you for improving the track with this change.

PR #1114, opened 2022-10-04
Sync metadata with problem-specifications

kytrinyx:
A number of practice exercises have minor changes
to the metadata. Mostly it's just punctuation in the blurbs, though there are a few other minor changes such as tweaks to source URLs or updating wording for clarity.

Since these should all be straight forward to review, I've taken the liberty of updating them all in a single PR.

### DIFF
--- a/exercises/practice/acronym/.meta/config.json
+++ b/exercises/practice/acronym/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "Convert a long phrase to its acronym",
+  "blurb": "Convert a long phrase to its acronym.",
   "authors": [
     "lpalma"
   ],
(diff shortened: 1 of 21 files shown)


petertseng (commented):
you got in an edit war with https://github.com/exercism/haskell/pull/1105 regarding whether URLs should be http vs https. I think I will prefer to keep them as https. We should get those updated in problem-specifications.

petertseng:
If this http to https change needs to be redone in the future, use commit https://github.com/exercism/haskell/pull/1115/commits/7dd57f1b7d4b6a38364dcbd8375bf7fa46768a4f

petertseng (approved):

kytrinyx:
That's a good catch regarding http vs https, thank you. I will look at these in problem-specifications.

PR #1012, opened 2021-11-01
Update stack.yml to lts-18.14

tejasbubane:
Last updated almost a year ago: https://github.com/exercism/haskell/pull/926

Closes #948

PR #1117, opened 2022-10-05
use resolver 19.27 (from 18.14); add stack.yaml to files.invalidator

petertseng:
https://www.stackage.org/lts-19.27

Last updated in November 2021:
https://github.com/exercism/haskell/pull/1012

https://www.stackage.org/diff/lts-18.14/lts-19.27

### DIFF
--- a/config.json
+++ b/config.json
@@ -31,6 +31,9 @@
     ],
     "exemplar": [
       ".meta/exemplar/src/%{pascal_slug}.hs"
+    ],
+    "invalidator": [
+      "stack.yaml"
     ]
   },
   "exercises": {
--- a/exercises/concept/guessing-game/.meta/config.json
+++ b/exercises/concept/guessing-game/.meta/config.json
@@ -14,6 +14,9 @@
     ],
     "exemplar": [
       ".meta/exemplar/src/GuessingGame.hs"
+    ],
+    "invalidator": [
+      "stack.yaml"
     ]
   }
 }
--- a/exercises/concept/guessing-game/stack.yaml
+++ b/exercises/concept/guessing-game/stack.yaml
@@ -1,1 +1,1 @@
-resolver: lts-18.14
+resolver: lts-19.27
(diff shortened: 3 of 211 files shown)


ErikSchierboom (commented):
Would this require any update to the test runner?

petertseng:
I see that the test runner is already out of date (it uses 16.21). But I think it would be ideal to update it; let's say a student is using some package that got updated between 19.27 and 16.21 and using a new feature that didn't exist in the version in the older resolver. Then it won't run on the test runner!

ErikSchierboom:
@petertseng Would you be willing to also update the test runner?

ErikSchierboom (approved):

petertseng:
aye-aye

MatthijsBlom:
GHCup skipped recommending GHC 9.0 because it has/had issues. (I couldn't find a reference.) The currently recommended GHC version is 9.2.4, but there is not yet a corresponding LTS Stackage snapshot.

petertseng:
If there are documented problems, it is certainly possible to go back. But my reasoning is, if there are problems with GHC 9.0, why would there be a Stackage snapshot for it at all? I hope we can agree that it doesn't really make sense to go back because of undocumented past problems. After all it is possible that they are now fixed.

petertseng:
https://github.com/haskell/vscode-haskell#ghc-abis-dont-match could be a problem, and it specifically mentions 9.0.2. I don't have the ability to firsthand confirm whether it happens because I don't have vscode-haskell though

MatthijsBlom:
> I hope we can agree that it doesn't really make sense to go back because of undocumented past problems.

We can.

> I don't have the ability to firsthand confirm whether it happens because I don't have vscode-haskell though

I tried the `hello-world` exercise and did not encounter this problem.

petertseng:
As a starting point to looking for any potential problems with 9.0, there was a mail thread https://mail.haskell.org/pipermail/haskell-cafe/2021-August/134352.html that cited some major 9.0.1 issues

MatthijsBlom:
That's the one I remember.

Now I remember more: at the time of 9.0, GHC outran tooling support. The latter only caught up by the time of 9.2. Not relevant for us.

PR #1119, opened 2022-10-05
remove unneeded hash from test files

petertseng:
In https://github.com/exercism/haskell/pull/877 we added a hash to all test files, so that when we update stack.yaml we could signal to the website that an exercise has changed and that students should update to the new version (this used to be a problem in
https://github.com/exercism/haskell/issues/864, where students wouldn't get up-to-date versions of the files)

In https://github.com/exercism/haskell/pull/1117, `stack.yaml` was added as an invalidator, so we no longer need to have this hash.
Used by website in:
https://github.com/exercism/website/blob/991fc32d9e0783a043fb0e009aa2ac90686a1172/app/models/git/exercise.rb#L109-L112
https://github.com/exercism/website/blob/991fc32d9e0783a043fb0e009aa2ac90686a1172/app/models/git/exercise.rb#L194-L208


petertseng:
Hmm, a little iffy. I was on nucleotide-count, which I started 3 years ago. I ran `exercism download --exercise=nucleotide-count --track=haskell`, and I got a stack.yaml with 16.21. But, I made a no-op change and submitted the exercise again, and then when I run `exercism download --exercise=nucleotide-count --track=haskell` I get the stack.yaml with 19.27. So something is working but it's only happening at submit time. I'll have to accept that as good enough for now then.

PR #1133, opened 2022-12-25
Exclude `stack.yaml.lock` files

MatthijsBlom:
`stack.yaml.lock` files are generated automatically when exercises are tested and during their development.

Alternatively we could exclude `*.lock` more generally, which I feel makes some sense but could also break in the future.


MatthijsBlom:
@petertseng Do you object to this change?

ErikSchierboom (approved):
@MatthijsBlom LGTM

PR #1138, opened 2023-01-29
saddle-points/test: Do not use head on empty list

BassTony:
The second test case does not play well with `head` on line 21, since using `head` on an empty list yields an error.

Use `head` only when the list is not empty.


petertseng (approved):
it's pretty hard to actually cause the test to actually evaluate `head []`, but just to demonstrate it's possible, you can make it happen with `saddlePoints arr = if snd (snd (bounds arr)) == 3 then [(2, 1)] else []`. Since there is a possible implementation that can make it happen, we should prevent it.

PR #1139, opened 2023-01-31
Add Approaches for Leap

MatthijsBlom:


MatthijsBlom:
Mind: `logical-expression/content.md` is still upcoming. The snippet is present already so at least you'll know what it is about.

exercises/practice/leap/.approaches/conditional-expression/content.md
```diff
@@ -0,0 +1,114 @@
...
+
+Conditional expressions are [syntactic sugar][wikipedia-syntactic-sugar] for certain `case` expressions:
+
+```haskell
+_ = if p then t else f
+-- is an abbreviation of
+_ = case p of
+  True  -> t
+  False -> f
+```
```
ErikSchierboom, line 33: This might be perfect content for an `exercism/note` block.
See https://exercism.org/docs/building/markdown/markdown#h-special-blocks-sometimes-called-admonitions

Feel free to ignore

exercises/practice/leap/.approaches/config.json
```diff
@@ -0,0 +1,37 @@
+{
+    "introduction": {
+        "authors": [
+            "MatthijsBlom"
+        ],
+        "contributors": []
```
ErikSchierboom, line 6:
```suggestion
```

config.json
```diff
@@ -18,6 +18,9 @@
   "test_runner": {
     "average_run_time": 3.0
   },
+  "approaches": {
+    "snippet_extension": "hs"
+  },
```
ErikSchierboom, line 23: We don't yet support this in the website, sorry. The approach snippets must thus all be `snippet.txt`.

Side note: I wonder if making approach snippets be `snippet.md` markdown files like with articles is the more sensible option.

ErikSchierboom (commented):
Fantastic content!

MatthijsBlom:
I will push `logical-expression/content.md` later this evening, at which point I'll think this is ready to merge.

@petertseng I intended for these commits to be squashed. Would you prefer something else?

exercises/practice/leap/.approaches/introduction.md
```diff
@@ -0,0 +1,63 @@
...
+
+[conditional-expression]:
+    https://exercism.org/tracks/python/exercises/leap/approaches/conditional-expression
+    "Approach: a conditional expression"
+[guards]:
+    https://exercism.org/tracks/python/exercises/leap/approaches/guards
+    "Approach: a sequence of guards"
+[logical-expression]:
+    https://exercism.org/tracks/python/exercises/leap/approaches/logical-expression
+    "Approach: a logical expression"
```
kahgoh, line 46: I am getting an HTTP 404 when I try to go to these python track URLs in my browser.

ErikSchierboom (approved):
LGTM!

exercises/practice/leap/.approaches/conditional-expression/content.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+## Conditional expressions
+
+A _conditional expression_ (`if … then … else …`) is a compound expression that uses a test to determine which of two alternatives to evaluate to.
+Many other languages feature a similar construct, often termed 'ternary operator'.
+They are also known as _`if` expressions_.
+
+Concretely, when `p` is some expression of type `Bool` and `t` and `f` are any two expressions of the same type, then `if p then t else f` will
```
kahgoh, line 20: I notice there were a few sentences that began with _Concretely, ..._. Personally, I think dropping this makes the text easier to read. For example:

```suggestion
When `p` is some expression of type `Bool` and `t` and `f` are any two expressions of the same type, then `if p then t else f` will
```

ErikSchierboom:
LGTM! From my viewpoint it is read to be merged

petertseng (approved):
I used to be a purist and for code that I'm personally writing I would only use `rem` for this situation and never `mod`, but

1. I would not have enforced that on code that others are writing, and 
2. I have just read up on https://stackoverflow.com/questions/339719/when-is-the-difference-between-quotrem-and-divmod-useful#comment29716414_342379 and that tells me reason for being purist was mistaken

MatthijsBlom:
I deliberately ignored performance considerations, and instead tried to write very accessible code. Now I'm questioning whether I made the right choice though:

```
 x  y   quotRem   divMod   
 7  3     2  1     2  1
 7 -3    -2  1    -3 -2
-7  3    -2 -1    -3  2
-7 -3     2 -1     2 -1
```

Not that this behavior difference matters here: the formula breaks down for 'negative' years anyway because there was no year 0.

ErikSchierboom:
> I deliberately ignored performance considerations, and instead tried to write very accessible code. 

That's a very sensible thing to do!

The approaches are live: https://exercism.org/tracks/haskell/exercises/leap/dig_deeper
One thing I noticed is that the `introduction.md` document should list the approaches under their own heading and with (an excerpt of) its code.
See https://exercism.org/tracks/csharp/exercises/collatz-conjecture/dig_deeper for an example.

kahgoh:
Just noticed one of the links on the [conditional expression](https://exercism.org/tracks/haskell/exercises/leap/approaches/conditional-expression) isn't rendered properly.
conditional_link

MatthijsBlom:
Yes, #1146 fixes that. Thanks for checking 🙂

PR #1147, opened 2023-02-15
leap approaches: Present the approaches in the overview document

MatthijsBlom:
> One thing I noticed is that the `introduction.md` document should list the approaches under their own heading and with (an excerpt of) its code.
> See https://exercism.org/tracks/csharp/exercises/collatz-conjecture/dig_deeper for an example.

_Originally posted by @ErikSchierboom in https://github.com/exercism/haskell/issues/1139#issuecomment-1430881119_

@ErikSchierboom You meant this, right?

I think this document is now compliant with the (relevant section of the) [approaches building docs](https://exercism.org/docs/building/tracks/approaches#h-approaches-overview).

### DIFF
--- a/exercises/practice/leap/.approaches/introduction.md
+++ b/exercises/practice/leap/.approaches/introduction.md
@@ -1,10 +1,51 @@
 # Introduction
 
-There are various idiomatic approaches to solve Leap, such as
+There are various idiomatic approaches to solve Leap.
+All approaches listed below check for divisibility by 4, 100, and 400.
+However, they differ in the ways in which they combine these checks.
 
-- constructing [a logical expression][logical-expression],
-- using [guards][guards], and
-- using [a conditional expression][conditional-expression].
+
+## Approach: a logical expression
+
+```haskell
+isLeapYear :: Integer -> Bool
+isLeapYear year = divisibleBy 4 && (not (divisibleBy 100) || divisibleBy 400)
+  where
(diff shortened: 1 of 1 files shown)


exercises/practice/leap/.approaches/introduction.md
```diff
@@ -1,10 +1,51 @@
...
+## Approach: a logical expression
+
+```haskell
+isLeapYear :: Integer -> Bool
+isLeapYear year = divisibleBy 4 && (not (divisibleBy 100) || divisibleBy 400)
+  where
+    divisibleBy d = year `mod` d == 0
+```
+
+[Read more about this approach][logical-expression].
```
ErikSchierboom, line 17: Whilst it doesn't really make much sense here, something to consider for other approaches would be to add a one-liner before the  "Read more" sentence that describes the approach taken. But the approach is short enough I think to not need that.

ErikSchierboom (approved):
Excellent!

exercises/practice/leap/.approaches/introduction.md
MatthijsBlom: Yes, I do write these for other exercises.

I could only think of something for one of the three but not the other two, so I left it out.

ErikSchierboom: Perfect!

petertseng (approved):
for immediate distribution

PR #1148, opened 2023-02-26
don't call  String variable hello a function

das-g:
Although function values and non-function values (and the variables bound to them) have a lot in common in Haskell, we shouldn't call all of them "functions".

See discussion [Is a string value in Haskell a function? ](https://forum.exercism.org/t/is-a-string-value-in-haskell-a-function/4034) on the Exercism forum.


exercises/practice/hello-world/.docs/instructions.append.md
```diff
@@ -1,6 +1,6 @@
 # Hints
 
-To complete this exercise, you need to implement the `hello` function.
+To complete this exercise, you need to define the variable `hello`.
```
MatthijsBlom, line 3: It isn't exactly a _variable_ either, as it cannot change, i.e. _vary_.

```suggestion
To complete this exercise, you need to define the constant `hello`.
```

Alternatives: _name_, _value_. The former might confuse; the latter conflicts with line 6.

exercises/practice/hello-world/.docs/instructions.md
```diff
@@ -8,7 +8,7 @@ or environment.
 
 The objectives are simple:
 
-- Write a function that returns the string "Hello, World!".
+- Define a variable containing the string "Hello, World!".
```
petertseng, line 11: the content of instructions.md must be exactly the same as those in https://github.com/exercism/problem-specifications/blob/main/exercises/hello-world/description.md. Therefore, this must be the following:

```suggestion
- Modify the provided code so that it produces the string "Hello, World!".
```

We have no choice in this matter.

exercises/practice/hello-world/.docs/instructions.append.md
```diff
@@ -1,6 +1,6 @@
 # Hints
 
-To complete this exercise, you need to implement the `hello` function.
+To complete this exercise, you need to define the constant `hello`.
 
 You will find the type signature for `hello` already in place,
```
petertseng, line 5: type signature is a term only applicable to functions.

```suggestion
You will find the type for `hello` already in place,
```

MatthijsBlom: Do you have a source on this? I tentatively disagree.

petertseng: I was going to say https://en.wikipedia.org/wiki/Type_signature, but for Haskell specifically, the report supersedes, so it is correct to call it a signature. But I suspect (without proof) that it may be seen as incorrect in common language. Is "type annotation" better according to either criterion?

exercises/practice/hello-world/.docs/instructions.md
das-g: > the content of instructions.md must be exactly the same as those in https://github.com/exercism/problem-specifications/blob/main/exercises/hello-world/description.md. […]
> 
> We have no choice in this matter.

Might be worth mentioning that in a `CONTRIBUTING.md` or other place where new contributors are likely to look. Even though I tried to find and review this project's rules before filing this PR, I wasn't at all aware of this requirement.

As the current instructions (before this PR) already didn't match that specification, would it thus have been necessary to change them sooner or later, anyway?

> Therefore, this must be the following:
> ```suggestion
> - Modify the provided code so that it produces the string "Hello, World!".
> ```

Should we thus change the exercise and test to expect a function? Claiming that code with just a constant "produces" anything may not be completely wrong, but seems a bit far-fetched and not very intuitive.

petertseng: I don't think we would want to have the tests pass an argument that the function just ignores (since the function would in all likelihood just be `const "Hello, World!"`), so I don't think we should have `hello` be a function.

MatthijsBlom: 1. There is something to be said for `hello` actually really producing something (laziness).
2. I do not expect confusion to arise from this word choice. Do you?

exercises/practice/hello-world/.docs/instructions.append.md
MatthijsBlom: I understand «type annotation» to mean what the report calls «[expression type-signature](https://www.haskell.org/onlinereport/haskell2010/haskellch3.html#x8-560003.16)». (In Haskell. E.g. in Python it's different.)

I do not know much about common use of «type signature» outside of Haskell. But 'Inside' Haskell it seems unnatural to me to reserve it for declarations of the form `name :: type -> …` and to use a different term for the other such declarations. Also, I cannot remember ever having seen a distinction being made before, hence my surprise.

exercises/practice/hello-world/.docs/instructions.md
MatthijsBlom: > I actually think it's a rather strong one.

A strongly weakening factor is that a beginner is very unlikely to understand this argument, much less come up with it.

das-g: > A strongly weakening factor is that a beginner is very unlikely to understand this argument, much less come up with it.

That may depend heavily on their previous knowledge of other programming languages and of other fields. Those not (yet) much aware of a compile-time vs. run-time distinction won't have much problems accepting it, I guess. Similarly, I imagine that to someone coming to programming from, say, linguistics (or even just theoretical computer science, where grammars _produce_ stuff), that argument might also seem quite natural. :shrug:

PR #1178, opened 2023-07-10
Sync exercise docs for most exercises

IsaacG:
Sync exercise docs for most exercises where the sync does not change content meaning significantly.

Some exercises have content which would change meaning a bit with a sync and were excluded from this PR.

### DIFF
--- a/exercises/practice/accumulate/.docs/instructions.md
+++ b/exercises/practice/accumulate/.docs/instructions.md
@@ -1,9 +1,6 @@
 # Instructions
 
-Implement the `accumulate` operation, which, given a collection and an
-operation to perform on each element of the collection, returns a new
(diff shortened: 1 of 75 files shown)


IsaacG:
Forum topic: https://forum.exercism.org/t/instructions-in-erlang-saddle-points-are-unclear-or-corrupted/6419/5

MatthijsBlom (approved):
Looks entirely good to me 👍

ErikSchierboom (approved):

iHiD:
FYI, this shouldn't have been merged until we add the support for skip test runs, as it's caused every haskell solution to rerun.

MatthijsBlom:
Is there a way I could have known?

PR #1191, opened 2024-01-15
Added reverse-string exercise

cdimitroulas:
As part of preparation for 48in24, I saw in the Exercism forums that the `reverse-string` exercise will be coming up on the 23rd of January and the Haskell track did not have this exercise implemented.

This PR implements the exercise. Apologies if I have missed anything, this is my first time contributing to an Exercism language track.


cdimitroulas:
Started a new topic in the forums for this: https://forum.exercism.org/t/added-reverse-string-exercise-to-haskell-track/9213

exercises/practice/reverse-string/stack.yaml.lock
```diff
@@ -0,0 +1,12 @@
+# This file was autogenerated by Stack.
```
MatthijsBlom, line 1: This file should not be committed.

I have proposed to have these files be automatically ignored in #1133.

ErikSchierboom (commented):
@cdimitroulas Once the review suggestions are addressed, I'll merge this

ErikSchierboom:
> Curious to know how you folks feel about this?

The goal of the example solution is just to prove that there is _a_ solution that passed the tests. In other words: to verify that students will be able to solve the exercise. So a simple example is perfectly fine.

exercises/practice/reverse-string/src/ReverseString.hs
```diff
@@ -0,0 +1,3 @@
+module ReverseString (reverseString) where
+
+reverseString :: String -> String
```
ErikSchierboom, line 3:
```suggestion
reverseString :: String -> String
reverseString :: str = error "You need to implement this function."
```

ErikSchierboom (commented):
One final, tiny nit

exercises/practice/reverse-string/src/ReverseString.hs
cdimitroulas: Weird, the accepted suggestion from Matthijs' review comment seems to have deleted the line instead of updating it 🤔 
Good catch, will fix.

PR #1193, opened 2024-01-17
Added Rational Numbers exercise

tofische:
Added a new Haskell practice exercise Rational Numbers based on the available problem specification.

### DIFF
--- a/exercises/practice/rational-numbers/src/RationalNumbers.hs
+++ b/exercises/practice/rational-numbers/src/RationalNumbers.hs
@@ -1,0 +1,46 @@
+module RationalNumbers
+(Rational,
+ abs,
+ reduce,
+ add,
+ sub,
+ mul,
+ div,
+ exprational,
+ expreal,
+ rational) where
+
+import Prelude hiding (div, abs, Rational)
+
+-- Data definition -------------------------------------------------------------
+data Rational a = Dummy deriving(Eq, Show)
+
+rational :: (a, a) -> Rational a
+rational = error "You need to implement this function"
+
+-- unary operators -------------------------------------------------------------
+abs :: Integral a => Rational a -> Rational a
+abs = error "You need to implement this function"
+
+reduce :: Integral a => Rational a -> Rational a
+reduce = error "You need to implement this function"
+
+-- binary operators ------------------------------------------------------------
+add :: Integral a => Rational a -> Rational a -> Rational a
+add = error "You need to implement this function"
+
+sub :: Integral a => Rational a -> Rational a -> Rational a
+sub = error "You need to implement this function"
+
+mul :: Integral a => Rational a -> Rational a -> Rational a
+mul = error "You need to implement this function"
+
+div :: Integral a => Rational a -> Rational a -> Rational a
+div = error "You need to implement this function"
+
+exprational :: Integral a => Rational a -> a -> Rational a
+exprational = error "You need to implement this function"
+
+expreal :: Floating a => Integral b => a -> Rational b -> a
+expreal = error "You need to implement this function"
+
--- a/exercises/practice/rational-numbers/src/RationalNumbers.hs
+++ b/exercises/practice/rational-numbers/src/RationalNumbers.hs
@@ -1,13 +1,15 @@
 module RationalNumbers
 (Rational,
  abs,
- reduce,
+ numerator,
+ denominator,
  add,
  sub,
  mul,
  div,
- exprational,
- expreal,
+ pow,
+ expRational,
+ expReal,
  rational) where
 
 import Prelude hiding (div, abs, Rational)
@@ -15,15 +17,18 @@ import Prelude hiding (div, abs, Rational)
 -- Data definition -------------------------------------------------------------
 data Rational a = Dummy deriving(Eq, Show)
 
-rational :: (a, a) -> Rational a
+rational :: Integral a => (a, a) -> Rational a
 rational = error "You need to implement this function"
 
 -- unary operators -------------------------------------------------------------
 abs :: Integral a => Rational a -> Rational a
 abs = error "You need to implement this function"
 
-reduce :: Integral a => Rational a -> Rational a
-reduce = error "You need to implement this function"
+numerator :: Integral a => Rational a -> a
+numerator = error "You need to implement this function"
+
+denominator :: Integral a => Rational a -> a
+denominator = error "You need to implement this function"
 
 -- binary operators ------------------------------------------------------------
 add :: Integral a => Rational a -> Rational a -> Rational a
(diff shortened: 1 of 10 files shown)


MatthijsBlom:
I don't think `reduce` should be exported. Going by the instructions there shouldn't be a need for that either:

> Your implementation of rational numbers should always be reduced to lowest terms.

I understand this to mean that it should be impossible for a user of this module to construct unreduced fractions.

To still be able to test proper reduction, we could require export of projections `numerator` and `denominator`.

exercises/practice/rational-numbers/src/RationalNumbers.hs
```diff
@@ -0,0 +1,46 @@
+module RationalNumbers
+(Rational,
```
MatthijsBlom, line 2: As there is a smart constructor `rational`, perhaps the 'dumb' `Rational` should not be exported.

exercises/practice/rational-numbers/src/RationalNumbers.hs
```diff
@@ -0,0 +1,46 @@
...
+ abs,
+ reduce,
+ add,
+ sub,
+ mul,
+ div,
+ exprational,
+ expreal,
```
MatthijsBlom, line 10: Name suggestions:

- `expRational`, or plain `exp`
- `expReal`

exercises/practice/rational-numbers/src/RationalNumbers.hs
```diff
@@ -0,0 +1,46 @@
...
+ rational) where
+
+import Prelude hiding (div, abs, Rational)
+
+-- Data definition -------------------------------------------------------------
+data Rational a = Dummy deriving(Eq, Show)
```
MatthijsBlom, line 16: Do we really want to parameterize this type? We test no types other than `Integer`.

exercises/practice/rational-numbers/test/Tests.hs
```diff
@@ -0,0 +1,305 @@
...
+                   , number1C     :: (Integer, Integer)
+                   , number2C     :: Integer
+                   , expectedC    :: (Integer, Integer)
+                   }
+
+data CaseD = CaseD { descriptionD :: String
+                   , number1D     :: Float
+                   , number2D     :: (Integer, Integer)
+                   , expectedD    :: Float
+                   }
```
MatthijsBlom, line 71: These four data types _could_ be combined into two (`CaseUnary`, `CaseBinary a b c`). Is the separation into four cases deliberate?

ErikSchierboom:
@MatthijsBlom Ping me when you've approved!

MatthijsBlom:
We now have `rational :: (a, a) -> Rational a`, but this allows creating fractions with 0 as denominator. Seemingly, having `rational :: (a, a) -> Maybe (Rational a)` would be more idiomatic. However, then `div` would still allow us to commit division by zero, so I think we might as well keep the present `rational :: (a, a) -> Rational a`.

exercises/practice/rational-numbers/src/RationalNumbers.hs
tofische: Having Rational exported allows pattern matching which is beneficial for simple data with defined internal structure (cf. Complex).

Not exporting data constructor directly it is beneficial if a smart constructor shall assure some kind of internal consistency. This might be a case also here if we wanted to assure correct rationals - see also [this comment](https://github.com/exercism/haskell/pull/1193#issuecomment-1898003150), but this goes beyond the scope of this exercise.

exercises/practice/rational-numbers/src/RationalNumbers.hs
tofische: I've taken these identifiers from `canonical-data.schema.json` from problem description. I'm unsure to which extend am I allowed to deviate from that file. If I could choose identifiers, I'd use `pow` and `exp` ;-)

I have also a different question in this context and I'm not sure how to proceed:

The instructions require the support for three types of exponentiation:

a) Exponentiation of a rational number to an integer power with two sub-cases (positive and negative exponent)
b) Exponentiation of a rational number to a real number
c) Exponentiation of a real number to a rational number

There are however no tests for the case b) in the problem specification. If I want to cover this case as well, shall I add new tests also there, or only in the Haskell track?

exercises/practice/rational-numbers/src/RationalNumbers.hs
tofische: But we want to demonstrate this - cf. Complex Numbers.

exercises/practice/rational-numbers/test/Tests.hs
tofische: No it isn't, I just didn't pay attention to the economics of test cases.

tofische:
> I don't think `reduce` should be exported. Going by the instructions there shouldn't be a need for that either:
> 
> > Your implementation of rational numbers should always be reduced to lowest terms.
> 
> I understand this to mean that it should be impossible for a user of this module to construct unreduced fractions.
> 
> To still be able to test proper reduction, we could require export of projections `numerator` and `denominator`.

You are correct, I'll update the exercise.

exercises/practice/rational-numbers/src/RationalNumbers.hs
MatthijsBlom: Parameterizing `Complex` was first suggested in https://github.com/exercism/haskell/pull/601#issuecomment-335623224. Having read the discussion, I still don't see much point to it. But given the `Integral` constraints all over the API I do not foresee any problems either, so I'm fine with keeping it parameterized.

exercises/practice/rational-numbers/src/RationalNumbers.hs
MatthijsBlom: > I'm unsure to which extend am I allowed to deviate from that file.

There is not set limit. Names are routinely tailored to language tracks.

> The instructions require the support for three types of exponentiation: [&hellip;]

I think we are allowed to deviate from the canonical instructions, just like we are allowed to deviate from the canonical tests. @ErikSchierboom is this correct / what do you think?

ErikSchierboom: That is correct

ErikSchierboom:
@MatthijsBlom could you review again?

tofische:
I'm now a little bit confused - shall I do something in order to enable this PR to be merged?

MatthijsBlom:
Sorry, I haven't been able to find the energy to review properly again.

tofische:
> Sorry, I haven't been able to find the energy to review properly again.

No problem, I was just unsure because one check (although not required) failed.

ErikSchierboom:
@MatthijsBlom if you haven't found the time in the next couple of days, I'll look into it

ErikSchierboom (approved):
LGTM! @MatthijsBlom If you find time and find things to improve, we can do that as a follow-up PR.

PR #1195, opened 2024-01-20
Affine cipher exercise

tofische:
Added a new Haskell practice exercise Affine Cipher based on the available problem specification.


tofische:
PR is obsolete, no further activity necessary.

PR #1196, opened 2024-01-20
Affine cipher exercise

tofische:
Added a new Haskell practice exercise Affine Cipher based on the available problem specification.

### DIFF
--- a/exercises/practice/affine-cipher/.meta/config.json
+++ b/exercises/practice/affine-cipher/.meta/config.json
@@ -1,0 +1,22 @@
+{
+  "authors": [],
+  "files": {
+    "solution": [
+      "src/AffineCipher.hs",
+      "package.yaml"
+    ],
+    "test": [
+      "test/Tests.hs"
+    ],
+    "example": [
+      ".meta/examples/success-standard/src/AffineCipher.hs",
+      ".meta/examples/success-text/src/AffineCipher.hs"
+    ],
+    "invalidator": [
+      "stack.yaml"
+    ]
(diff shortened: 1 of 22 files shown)


exercises/practice/affine-cipher/.meta/config.json
```diff
@@ -0,0 +1,22 @@
+{
+  "authors": [],
```
BNAndras, line 2: Not a Haskell maintainer here, but this should contain your GitHub username.

tofische: You're right, corrected.

MatthijsBlom:
It looks like you included `rational-numbers` by accident.

tofische:
> It looks like you included `rational-numbers` by accident.

As I explained [here](https://forum.exercism.org/t/affine-cipher-exercise/9320/3), it was my mistake because I created the affine-cipher-exercise branch from the not-yet-merged rational-numbers-exercise branch and not from the main branch.

If I understand git properly, this PR can be merged without problems after Rational Numbers is merged, so no corrective action is necessary - is this assumption correct?

MatthijsBlom:
I suspect so. [`git rebase --onto`](https://git-scm.com/docs/git-rebase) should also work, I think.

tofische:
> I suspect so. [`git rebase --onto`](https://git-scm.com/docs/git-rebase) should also work, I think.

It shall be corrected now.

tofische:
PR is obsolete.

PR #1207, opened 2024-02-06
Decrease all Dig Deeper header levels by one

MatthijsBlom:
Witness: [Bob &ndash; Dig Deeper](https://exercism.org/tracks/haskell/exercises/bob/dig_deeper), the header «Beware partial functions».

It seems the recent page format change broke some rendering. In particular, `<h4>` (`###` in the source material) is now visually indistinguishable from plain text. If this is intentional, then please merge this. If it is not, then merging is probably unnecessary/undesirable.

@ErikSchierboom

### DIFF
--- a/exercises/practice/bob/.approaches/introduction.md
+++ b/exercises/practice/bob/.approaches/introduction.md
@@ -1,12 +1,10 @@
-# Introduction
-
 This problem requires checking against a series of conditions, each associated with a desired output.
 This suggests using guards.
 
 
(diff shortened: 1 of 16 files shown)


ErikSchierboom:
> If it is not, then merging is probably unnecessary/undesirable.

This was not our goal. @dem4ron could you look into this?

MatthijsBlom:
( poking @dem4ron in case this fell off the radar )

PR #1210, opened 2024-02-16
Added Satellite exercise

tofische:
Added a new Haskell practice exercise Satellite based on the available problem specification.

### DIFF
--- a/config.json
+++ b/config.json
@@ -501,6 +501,17 @@
         ]
       },
       {
+        "slug": "satellite",
+        "name": "Satellite",
+        "uuid": "22136812-8160-49f0-8a0d-f76a2086a34c",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 6,
+        "topics": [
+          "maybe"
+        ]
+      },
+      {
         "slug": "go-counting",
         "name": "Go Counting",
         "uuid": "8f1854d7-b8cb-4dfa-9e6a-8811b0891798",
--- a/exercises/practice/satellite/package.yaml
+++ b/exercises/practice/satellite/package.yaml
@@ -3,7 +3,6 @@ version: 1.0.0.0
 
 dependencies:
   - base
-  - containers
 
 library:
   exposed-modules:
--- a/exercises/practice/satellite/package.yaml
+++ b/exercises/practice/satellite/package.yaml
@@ -1,0 +1,22 @@
+name: satellite
+version: 1.0.0.0
+
+dependencies:
+  - base
+  - containers
+
+library:
+  exposed-modules: Satellite
+  source-dirs: src
+  ghc-options: -Wall
+  # dependencies:
+  # - foo       # List here the packages you
+  # - bar       # want to use in your solution.
+
+tests:
+  test:
+    main: Tests.hs
+    source-dirs: test
+    dependencies:
+      - satellite
+      - hspec
--- a/exercises/practice/satellite/package.yaml
+++ b/exercises/practice/satellite/package.yaml
@@ -6,7 +6,9 @@ dependencies:
   - containers
 
 library:
-  exposed-modules: Satellite
+  exposed-modules:
+  - Satellite
+  - BinaryTree
   source-dirs: src
   ghc-options: -Wall
   # dependencies:
--- a/exercises/practice/satellite/src/Satellite.hs
+++ b/exercises/practice/satellite/src/Satellite.hs
@@ -2,5 +2,5 @@ module Satellite (treeFromTraversals) where
 
 import Data.Tree (Tree)
 
-treeFromTraversals :: String -> String -> Maybe (Tree Char)
+treeFromTraversals :: Ord a => [a] -> [a] -> Maybe (Tree a)
 treeFromTraversals preorder inorder = error "You need to implement this function."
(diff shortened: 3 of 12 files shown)


config.json
```diff
@@ -500,6 +500,17 @@
...
+      {
+        "slug": "satellite",
+        "name": "Satellite",
+        "uuid": "22136812-8160-49f0-8a0d-f76a2086a34c",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 6,
+        "topics": [
+          "maybe"
+        ]
```
MatthijsBlom, line 512: @ErikSchierboom I am not familiar with this feature, so I cannot judge this.

exercises/practice/satellite/src/Satellite.hs
```diff
@@ -0,0 +1,6 @@
+module Satellite (treeFromTraversals) where
+
+import Data.Tree (Tree)
+
+treeFromTraversals :: String -> String -> Maybe (Tree Char)
```
MatthijsBlom, line 5:
```suggestion
treeFromTraversals :: Ord a => [a] -> [a] -> Maybe (Tree a)
```

Or is this unworkable?

tofische: Good proposal, done.

MatthijsBlom:
Hmm. This exercise is about binary trees, but `Data.Tree.Tree` is a rose tree. Should we be more precise and use some dedicated `BinaryTree` type instead?

tofische:
> Hmm. This exercise is about binary trees, but `Data.Tree.Tree` is a rose tree. Should we be more precise and use some dedicated `BinaryTree` type instead?

I agree with you, but I couldn't find a ready-to-use binary tree in any common package. I didn't want to use some exotic package, and I also didn't want to let the students to create their own tree implementation.

The closest I could find is [https://hackage.haskell.org/package/tree-traversals-0.1.2.0/docs/Data-BinaryTree.html](Data.BinaryTree) - what do you think about the usage of this package?

MatthijsBlom:
Neat!

If we add a pre-defined type in this way we should probably add a note about it in `instructions.append.md`.

```haskell
-- Which type to choose?

data BinaryTree a = Leaf    | Node a (BinaryTree a) (BinaryTree a)
-- or
data BinaryTree a = Leaf a | Node a (BinaryTree a) (BinaryTree a)
```

tofische:
> Neat!
> 
> If we add a pre-defined type in this way we should probably add a note about it in `instructions.append.md`.
> 
> ```haskell
> -- Which type to choose?
> 
> data BinaryTree a = Leaf   | Node a (BinaryTree a) (BinaryTree a)
> -- or
> data BinaryTree a = Leaf a | Node a (BinaryTree a) (BinaryTree a)
> ```

I have to admit that I'm not keen on the idea of providing a custom-made basic data structure - one of the purposes of the exercises is to point people towards suitable existing libraries. At least for me it was a big benefit that I got from the Exercism, so I don't want to withhold it from others.

What are the arguments against using `Data.BinaryTree`?

MatthijsBlom:
> What are the arguments against using `Data.BinaryTree?`

To me: that it is not [small](https://www.parsonsmatt.org/2018/10/02/small_types.html). Many rose trees are not binary trees. Haskell allows and encourages expressing one's intent precisely. Defining new types even for single use is not unusual.

tofische:
> > What are the arguments against using `Data.BinaryTree?`
> 
> To me: that it is not [small](https://www.parsonsmatt.org/2018/10/02/small_types.html). Many rose trees are not binary trees. Haskell allows and encourages expressing one's intent precisely. Defining new types even for single use is not unusual.

Are we talking about the same thing? Source of `Data.BinaryTree`:

```
data BinaryTree a = Leaf | Branch (BinaryTree a) a (BinaryTree a)
  deriving (Show, Functor, Eq)
```

MatthijsBlom:
No, sorry. I thought we were talking about `Data.Tree` from `containers`.

`tree-traversals` seems not a commonly used package. In any case it would have to be added to the test runner. I have no definite opinion on whether we should use it.

tofische:
> No, sorry. I thought we were talking about `Data.Tree` from `containers`.
> 
> `tree-traversals` seems not a commonly used package. In any case it would have to be added to the test runner. I have no definite opinion on whether we should use it.

This is a valid argument.

exercises/practice/satellite/package.yaml
```diff
@@ -0,0 +1,24 @@
+name: satellite
+version: 1.0.0.0
+
+dependencies:
+  - base
+  - containers
```
MatthijsBlom, line 6: `containers` isn't needed anymore.

MatthijsBlom (approved):

exercises/practice/satellite/package.yaml
tofische: You are correct, it is need only in the example solution which uses Data.Set.

ErikSchierboom:
> tree-traversals seems not a commonly used package. In any case it would have to be added to the test runner. I have no definite opinion on whether we should use it.

When possible, we'd prefer not to add libraries to the test runner. 

>  one of the purposes of the exercises is to point people towards suitable existing libraries

I'd argue that this isn't generally true, although it can be in individual cases. Our idea is that students solve things _without_ using libraries, as Exercism is all about teaching fluency in a language, but not necessarily everything that goes with it (like tooling and libraries).

config.json
ErikSchierboom: Topics are deprecated and are not used by the website. They are allowed to be in the `config.json` as they used to be used and they could be useful to help determine the `practices` and `prerequisites` fields.

ErikSchierboom (approved):
LGTM!

PR #1212, opened 2024-02-18
 Fixed indentation error in yacht/package.yaml which made the tests not run on the website. 

Jaffenheimer:
Approved in the forum by iHiD

### DIFF
--- a/exercises/practice/yacht/package.yaml
+++ b/exercises/practice/yacht/package.yaml
@@ -1,21 +1,21 @@
-name:                yacht
-version:             1.2.0.0
+name: yacht
+version: 1.2.0.0
 
 dependencies:
-  - base
+    - base
 
 library:
-  exposed-modules: Yacht
-  source-dirs: src
-  ghc-options: -Wall
-  # dependencies:
-  # - foo       # List here the packages you
-  # - bar       # want to use in your solution.
+    exposed-modules: Yacht
+    source-dirs: src
+    ghc-options: -Wall
+    # dependencies:
+    # - foo       # List here the packages you
+    # - bar       # want to use in your solution.
 
 tests:
-  test:
-    main:        Tests.hs
-    source-dirs: test
-    dependencies:
-    - yacht
-    - hspec
+    test:
+        main: Tests.hs
+        source-dirs: test
+        dependencies:
+            - yacht
+            - hspec


MatthijsBlom:
Please link to relevant forum discussions explicitly.

What exactly is the problem that this patch attempts to solve? Please post all information you have.

This patch introduces lots of whitespace changes. Are they all necessary?

Jaffenheimer:
Here is a link to the forum discussion: https://forum.exercism.org/t/haskell-track-error-fix/9853

When trying to run the tests on the website the following error occurs: 
image
This is because that the lines 20 and 21 is not correctly indented. After indenting them the tests run fine.

I have made another commit with less whitespace changes (my code editor formatted it with tabs set to 4 spaces instead of 2).
The lines 1,2 and 8 have only been changed to match .yaml files from other exercises where there is not that much whitespace before the value of the key.

MatthijsBlom:
I can confirm that this patch seems to solve the problem.

Three thoughts:

- What exactly went wrong here? The other dependency lists do not seem to require the additional indentation. Is there a faulty parser in play, or a morally wrong spec (i.e. that of YAML), or something else? I am confused.

- I am unpleasantly surprised that this was not caught by CI. @ErikSchierboom maybe something to look into? I have no ideas here.

- Are there other exercises with the same problem? We'd probably want to fix those as well.

Jaffenheimer:
I compared this .yaml file, to other .yaml files from other exercises and at least the ones I looked at, had the same indentation under the dependencies and the rest of the specifications, and you can also see the same indentation under the dependencies in line 5. So i think the only thing that went wrong was a syntactical error i.e the missing indentation that the cabal file couldn't be generated.

In regards to other tests, i have attempted around 50 exercises in the haskell track, and this is the only issue i have encountered, i have however not checked the remaining exercises, so it would probably be a good idea to look into.

iHiD:
> I can confirm that this patch seems to solve the problem.

Merging 🙂 

---

Thanks for the PR and the review. Would be good to address the CI etc - I'll leave that for @ErikSchierboom when he's in next.

Jaffenheimer:
You are very welcome, happy to help :)

ErikSchierboom:
I don't think it's the C but the test runner. When I run `./bin/test-example $PWD/exercises/practice/yacht/.meta/examples/success-standard` locally, it's building fine.

My guess it that when https://github.com/exercism/haskell-test-runner/pull/104 was merged, this broke. @cdimitroulas would you mind taking a look?

ErikSchierboom (approved):

cdimitroulas:
@ErikSchierboom could you clarify how the test runner could affect the formatting of a `package.yaml` file committed to the Haskell track repo?

cdimitroulas:
Ah, I think what you're saying is that even with the bad indentation the test runner was able to run the tests. Whereas with the new test runner it fails.

This might be because of the code injection which adds extra dependencies to `package.yaml` and assumes the correct indentation level. Perhaps having dependencies with different indentation levels results in the error?

E.g.
```
dependencies:
- yacht
- hspec
    - hspec-core
    - bytestring
```

ErikSchierboom:
> Ah, I think what you're saying is that even with the bad indentation the test runner was previously able to run the tests. Whereas with the new test runner it fails.

That is indeed what I was saying. 

Is this fixable in the test runner? If not, we should add an extra step to CI in this repo to verify the correct indentation.

cdimitroulas:
@ErikSchierboom for now I reviewed all the practice exercise `package.yaml` files and they all looked good with the exception of one which had some benchmarks in it as well - PR for fixing that one here -> https://github.com/exercism/haskell/pull/1213

In order to fix this in the test runner we would need to make the code injection smarter. There are two things to make it more robust:
1. Identify the level of indentation currently used for `tests.dependencies` items and use the same when injecting code
2. Do not assume that `tests.dependencies` is the last item in the YAML. Instead identify it's location and inject the code at the right place (this avoids issues like the one fixed in https://github.com/exercism/haskell/pull/1213
 where there was a `benchmarks` section below `tests` in the YAML)

cdimitroulas:
@ErikSchierboom is it worthwhile to pursue item either (or both) of these improvements in the test runner implementation? For sure item `1.` is more valuable than `2.` given that I only found a single practice exercise package.yaml which didn't have `tests.dependencies` last due to some benchmarks in the project as well.

cdimitroulas:
I would need to look into that. Not sure off the top of my head what would be a good way to do that since the other indentation which caused an error with the test-runner is technically valid YAML.

ErikSchierboom:
Maybe use a yaml formatter?

cdimitroulas:
Sorry haven't had much time recently, I will look into if we can get a YAML formatter added to CI to check that this formatting is respected

PR #1223, opened 2024-03-27
Add Word Search exercise

tofische:
Added a new Haskell practice exercise Word Search based on the available problem specification.

### DIFF
--- a/config.json
+++ b/config.json
@@ -416,6 +416,17 @@
         ]
       },
       {
+        "slug": "word-search",
+        "name": "Word Search",
+        "uuid": "6d406931-7173-4fc0-b897-798227d09b5e",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 5,
+        "topics": [
+          "strings"
+        ]
+      },
+      {
         "slug": "wordy",
         "name": "Wordy",
         "uuid": "0869734e-0629-4db9-8e62-af246e5f91ff",
--- a/exercises/practice/word-search/.docs/instructions.md
+++ b/exercises/practice/word-search/.docs/instructions.md
@@ -1,0 +1,24 @@
+# Instructions
+
+In word search puzzles you get a square of letters and have to find specific words in them.
+
+For example:
+
+```text
+jefblpepre
+camdcimgtc
+oivokprjsm
+pbwasqroua
+rixilelhrs
+wolcqlirpc
+screeaumgr
+alxhpburyi
+jalaycalmp
+clojurermt
+```
+
+There are several programming languages hidden in the above square.
+
+Words can be hidden in all kinds of directions: left-to-right, right-to-left, vertical and diagonal.
+
+Given a puzzle and a list of words return the location of the first and last letter of each word.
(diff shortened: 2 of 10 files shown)


ErikSchierboom (approved):
Lgtm. @MatthijsBlom ?

MatthijsBlom:
I'm debating what the ideal / a good signature would be. Keeping in mind both idiomaticity and interestingness/scenario diversity.

```haskell
search :: Grid -> Needles -> Output
```

Should `Grid` be a single (`\n`-separated) `String`, a list of (line) `Strings`, a 2d `Array` of `Chars`, ...?

I'm fine with `Needles ~ [Word]`, and `Word ~ String` isn't too bad.

The tests do not cover cases in which the same word occurs several times, so `Output` could be a `Map`.

---

What do you two think about putting the `data`s into an `editor` file?

tofische:
> I'm debating what the ideal / a good signature would be. Keeping in mind both idiomaticity and interestingness/scenario diversity.
> 
> ```haskell
> search :: Grid -> Needles -> Output
> ```
> 
> Should `Grid` be a single (`\n`-separated) `String`, a list of (line) `Strings`, a 2d `Array` of `Chars`, ...?
> 
> I'm fine with `Needles ~ [Word]`, and `Word ~ String` isn't too bad.

The conversion between a single String and a list of Strings is trivial, and I deliberately avoided some more complicated data structure in order not to restrict the creativity of the students - very often they build up their solution based on the given data type instead of converting the data so some more appropriate representation (as also my example solution does ;-) ). I'm however open - if they are any advantages in using another data type, I'll implement the change.


> The tests do not cover cases in which the same word occurs several times, so `Output` could be a `Map`.

This is a valid point, but does the increase in the attractiveness of the problem justify additional complexity? The difference is in searching the first occurrence or searching all occurrences. I doubt it, but the example solution finds all occurrences (as a List, not a Map) and only then converts to Maybe.

> What do you two think about putting the `data`s into an `editor` file?

I'm sorry, I don't understand what do you mean.

ErikSchierboom:
> This is a valid point, but does the increase in the attractiveness of the problem justify additional complexity? The difference is in searching the first occurrence or searching all occurrences. I doubt it, but the example solution finds all occurrences (as a List, not a Map) and only then converts to Maybe.

I don't think these puzzles ever have a word occuring multiple times, so I would be fine not accounting for that fact.

tofische:
One check has failed, but I don't know how to correct it (if I cat correct it at all) - do you know what can be done now?

ErikSchierboom:
Nothing to do with this PR. Ci is broken

tofische:
> I've finally been able to fix CI and it turns out this exercise is broken: https://github.com/exercism/haskell/actions/runs/8720539792/job/23922275342?pr=1235 Would you mind looking into it @tofische?

Done (on the same branch). Do I have to create a new PR?

ErikSchierboom:
Yes please.

tofische:
https://github.com/exercism/haskell/pull/1236

PR #1241, opened 2024-04-23
Update package.yaml

karginoleg:
Added tabs to fix an invalid dependency error.


MatthijsBlom:
This is not the first time these problems have arisen. Should this not be fixed in `haskell-test-runner` instead, @ErikSchierboom?

ErikSchierboom:
> Should this not be fixed in haskell-test-runner instead, @ErikSchierboom?

Ideally: yes. I asked how hard that would be, but I was informed that it would not be easy.

PR #1279, opened 2024-10-30
Add Eliud's eggs exercise

c0m3tx:
Added Haskell practice exercise "Eliud's eggs" based on the available problem specification.


config.json
```diff
@@ -1443,15 +1438,15 @@
...
-    "platform/mac",
     "platform/linux",
+    "platform/mac",
+    "platform/windows",
     "runtime/standalone_executable",
+    "typing/static",
+    "typing/strong",
     "used_for/backends",
     "used_for/financial_systems"
   ]
```
iHiD, line 1452: A lot of this file has changed, which seems unrelated to the exercise being added.

BNAndras: Running `configlet create --practice-exercise <slug>` formats the track config as if the user had also run `configlet fmt`. So if this was the first Haskell exercise added since that functionality was added, the noise makes sense. The next Haskell exercise PR though shouldn't have this noise since the config would be already correctly formatted.

c0m3tx: Yeah I was also a bit confused when I ran the `configlet` cmd to be honest :sweat_smile:

iHiD: @BNAndras I defer to you! :)

BNAndras (approved):

exercises/practice/eliuds-eggs/.meta/examples/success-standard/src/EliudsEggs.hs
```diff
@@ -0,0 +1,6 @@
+module EliudsEggs (eggCount) where
+
+eggCount :: Int -> Int
+eggCount 0 = 0
+eggCount x | odd x = 1 + eggCount (x `div` 2)
+eggCount x = eggCount (x `div` 2)
```
kotp:
```suggestion
eggCount x = eggCount (x `div` 2)

```

exercises/practice/eliuds-eggs/src/EliudsEggs.hs
```diff
@@ -0,0 +1,4 @@
+module EliudsEggs (eggCount) where
+
+eggCount :: Int -> Int
+eggCount = error "You need to implement this function"
```
kotp:
```suggestion
eggCount = error "You need to implement this function"

```

kotp (approved):
A couple of missing EOL characters for files that are text files.

iHiD:
(cc @MatthijsBlom @tofische)

PR #1291, opened 2025-07-30
Add `square-root`

quintuple-mallard:
Adds the Square Root practice exercise.

I'm not sure if the version on package.yaml is correct, saw that different exercises list different versions.


quintuple-mallard:
Some other exercises' config files were changed by configlet.
This should be undone once the PR is reopened.

PR #1294, opened 2025-12-22
Correct failed build.

tofische:


IsaacG:
@tofische With the test runner updated, it might be worth rerunning the failed check here

IsaacG (approved):

IsaacG:
@tofische did you accidentally fix something that was meant to be broken?

tofische:
> @tofische did you accidentally fix something that was meant to be broken?

No, but it is not the first time that I see this error message. It seems that not every workflow respect STUB-ALLOWED-TO-NOT-COMPILE (if I understand the setup properly). Can't is be because the exemplar does compile?

To be honest, I don't see a point of having an exercise that is intentionally broken. IMHO a language track shall teach a programming language, not a very basic programming concepts, like write down a definition etc.

MatthijsBlom:
The reason for the failing Leap example appears to be here: https://github.com/exercism/haskell/pull/477#issuecomment-276259246

As for the concept exercise: the point is not the concept of giving names to values, but rather the syntax for doing so.

IsaacG:
The [test-stub](https://github.com/exercism/haskell/blob/main/bin/test-stub#L46) script checks if the exercise contains a `.meta/STUB-ALLOWED-TO-NOT-COMPILE` file. If the exercise does contain that file, the `test-stub` _expects_ and _requires_ that the exercise stub not compile; if the stub _does_ compile, `test-stub` will fail.

If you want this stub to be valid code that can compile, you need to also delete the file `exercises/concept/lucians-luscious-lasagna/.meta/STUB-ALLOWED-TO-NOT-COMPILE`. (I have no opinion on whether that is desirable or not for this exercise.)

tofische:
> The [test-stub](https://github.com/exercism/haskell/blob/main/bin/test-stub#L46) script checks if the exercise contains a `.meta/STUB-ALLOWED-TO-NOT-COMPILE` file. If the exercise does contain that file, the `test-stub` _expects_ and _requires_ that the exercise stub not compile; if the stub _does_ compile, `test-stub` will fail.
> 
> If you want this stub to be valid code that can compile, you need to also delete the file `exercises/concept/lucians-luscious-lasagna/.meta/STUB-ALLOWED-TO-NOT-COMPILE`. (I have no opinion on whether that is desirable or not for this exercise.)

I assume that STUB-ALLOWED-TO-NOT-COMPILE is used because the `LuciansLusciousLasagna.hs` template located in the `src` directory intentionally does not compile.

The question is however, which source file is used by the test-stub - successful tests of the guessing-game exercise (see lines 13 - 25 in the output of Check exercises step) suggest that instead of sources from the `src` directory (that can't be tested) sources from `.meta/exemplar/src` are used - but this example compiles and runs perfectly for this exercise despite of STUB-ALLOWED-TO-NOT-COMPILE.

I don't know, if my reasoning is correct and unfortunately I don't understand the exercism's infrastructure well enough.

IsaacG:
> The question is however, which source file is used by the test-stub - successful tests of the guessing-game exercise (see lines 13 - 25 in the output of Check exercises step) suggest that instead of sources from the `src` directory (that can't be tested) sources from `.meta/exemplar/src` are used - but this example compiles and runs perfectly for this exercise despite of STUB-ALLOWED-TO-NOT-COMPILE.

[The test runs the following loop](https://github.com/exercism/haskell/blob/main/.github/workflows/tests.yml#L127-L132), interleaving the stub checks and example checks.

```bash
          for exercise in ${{ github.workspace }}/exercises/{concept,practice}/*; do
            time bin/test-stub $exercise
            time bin/test-all-examples $exercise
          done
```

This tests the stub file (`exercises/concept/lucians-luscious-lasagna/src/LuciansLusciousLasagna.hs`) then the example file (`exercises/concept/lucians-luscious-lasagna/.meta/exemplar/src/LuciansLusciousLasagna.hs`). The first test takes into consideration the file `STUB-ALLOWED-TO-NOT-COMPILE` and requires that the stub compile or not compile based on the existence of that file. The second test (the exemplar test) does not use `STUB-ALLOWED-TO-NOT-COMPILE` and requires that the exemplar always compiles.

This PR changes the **stub** file (`exercises/concept/lucians-luscious-lasagna/src/LuciansLusciousLasagna.hs`) from not-compiling to compiling. If the stub's compile status changes, the `STUB-ALLOWED-TO-NOT-COMPILE` file needs to change. Since this PR changes the stub from non-compiling to compiling, `STUB-ALLOWED-TO-NOT-COMPILE` needs to be removed.

The example check is a _separate_ command from the stub check (`bin/test-all-examples` vs `bin/test-stub`) and does _not_ check `STUB-ALLOWED-TO-NOT-COMPILE`. I suspect you're confusing the outputs from `bin/test-all-examples` and the outpus from `bin/test-stub`.

The CI would be less confusing if the stub checks and example checks were not interleaved.

tofische:
> @tofische did you accidentally fix something that was meant to be broken?

Based on the discussion I think that you are correct, I'm going to close this PR without merging.

tofische:
No cleanup is necessary.

PR #1295, opened 2026-04-12
bin/verify-exercises-in-docker script

keiravillekode:
This script is useful for testing a practice exercise against a test runner image.

Example usage

```bash
$ bin/verify-exercises-in-docker -i exercism/haskell-test-runner:latest hello-world
```

PR #1296, opened 2026-05-13
Updated config.json to reference proper package.yaml

tofische:
See https://github.com/exercism/haskell/pull/1295#issuecomment-4416051540

Please check, if this patch solves the problem with bin/verify-exercises-in-docker script.

### DIFF
--- a/exercises/concept/pacman-rules/.meta/config.json
+++ b/exercises/concept/pacman-rules/.meta/config.json
@@ -10,7 +10,7 @@
     "test": [
       "test/Tests.hs"
     ],
-    "example": [
+    "exemplar": [
       ".meta/exemplar/src/PacmanRules.hs",
       ".meta/exemplar/package.yaml"
     ],
--- a/exercises/concept/pacman-rules/.meta/config.json
+++ b/exercises/concept/pacman-rules/.meta/config.json
@@ -10,8 +10,9 @@
     "test": [
       "test/Tests.hs"
     ],
-    "exemplar": [
-      ".meta/exemplar/src/PacmanRules.hs"
+    "example": [
+      ".meta/exemplar/src/PacmanRules.hs",
+      ".meta/exemplar/package.yaml"
     ],
     "invalidator": [
       "stack.yaml"
--- a/exercises/practice/affine-cipher/.meta/config.json
+++ b/exercises/practice/affine-cipher/.meta/config.json
@@ -12,7 +12,7 @@
     ],
     "example": [
       ".meta/examples/success-standard/src/Affine.hs",
-      ".meta/examples/success-text/src/Affine.hs"
+      ".meta/examples/success-standard/package.yaml"
     ],
     "invalidator": [
       "stack.yaml"
--- a/exercises/concept/guessing-game/.meta/config.json
+++ b/exercises/concept/guessing-game/.meta/config.json
@@ -10,8 +10,9 @@
     "test": [
       "test/Tests.hs"
     ],
-    "exemplar": [
-      ".meta/exemplar/src/GuessingGame.hs"
+    "example": [
+      ".meta/exemplar/src/GuessingGame.hs",
+      ".meta/exemplar/package.yaml"
     ],
     "invalidator": [
       "stack.yaml"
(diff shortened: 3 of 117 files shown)


tofische:
@IsaacG @keiravillekode  please have a look.

exercises/concept/pacman-rules/.meta/config.json
```diff
@@ -10,8 +10,9 @@
     "test": [
       "test/Tests.hs"
     ],
-    "exemplar": [
-      ".meta/exemplar/src/PacmanRules.hs"
+    "example": [
```
IsaacG: Concept exercises need an `exemplar`, not an `example`

keiravillekode (approved):

tofische:
Build has failed with an error that I cannot correct - @IsaacG can you help?

IsaacG:
I don't have write access to this repo. As a maintainer, I think you should be able to rerun that test. That step succeeded on the last PR that got merged here.

keiravillekode:
cc @MatthijsBlom

MatthijsBlom:
I may be missing some context. Is there still a problem? At present I do not know where to look; I'd have to study the test setups in general.

tofische:
> I may be missing some context. Is there still a problem? At present I do not know where to look; I'd have to study the test setups in general.

The purpose of this PR was to include `.meta/example/package.yaml` in the `config.json` file, section `example`. As far as I'm aware this part is not used by any tests, change was made to enable https://github.com/exercism/haskell/pull/1295 only.

IsaacG:
> I may be missing some context. Is there still a problem? At present I do not know where to look; I'd have to study the test setups in general.

The `config.json`'s `.files.example` is supposed to reference all the files that are part of the example solution. It ought to roughly mirror the .files.solution` list, which contains the `package.yaml` file. This PR corrects the files to the `.files.example` is complete.

I believe the `.files.example` is not used by the Exercism infrastructure (test runner, analyzer, representer) directly. It gets used by helper tools (like the one referenced in the original description) that tries to verify that the example solution passes the test. In order to test the example solution, verify tools rely on the `.files.example` list to locate files it needs. It copies files from `.files.example` to corresponding files in `.files.solution`. With the `packages.yaml` file missing, those tools do not copy that file and the examples fail to pass tests.

This PR corrects that issue by adding `packages.yaml` to the `.files.example`, making example-verification tools able verify that the examples pass the tests.

exercises/practice/affine-cipher/.meta/config.json
```diff
@@ -12,7 +12,7 @@
     ],
     "example": [
       ".meta/examples/success-standard/src/Affine.hs",
-      ".meta/examples/success-text/src/Affine.hs"
```
MatthijsBlom, line 15: This shouldn't be removed, right?
Should'nt also `success-text/package.yaml` be included?

IsaacG: [The purpose of the example file](https://exercism.org/docs/building/tracks/practice-exercises/add#h-add-example-implementation): 

> To ensure that it is possible to write code that passes the tests, an example implementation needs to be added.

Most (all?) other tracks have a single solution per exercise. Having multiple `Affine.hs` files as part of the example solution breaks the semantics used across the rest of Exercism. The purpose of the example is to demonstrate the problem is solve-able. This is done with a single solution.

If there's a desire to show different approaches to solving an exercise, that should be done via [approaches docs](https://exercism.org/docs/building/tracks/approaches).

IsaacG: And, yes, this should probably have the `packages.yaml` listed

tofische: > This shouldn't be removed, right? Should'nt also `success-text/package.yaml` be included?

I would disagree, I think that listing two independent solution (with two `packages.yaml` files) in one list is not the intent of this entry. Therefore I chose only only (main) solution to list here.

IsaacG (approved):

exercises/practice/affine-cipher/.meta/config.json
MatthijsBlom: Right, so it wasn't a mistake.

If I had to choose between the two test cases, I'd go with `succes-text`, as that is the one more likely to fail. But I do not have strong feelings about this.

MatthijsBlom (approved):

PR #1297, opened 2026-05-15
Use `Test.Hspec.Runner` to run concept tests, matching what is done for practice tests

IsaacG:


IsaacG:
+cc @exercism/haskell maintainers

tofische (approved):

PR #1298, opened 2026-05-16
[list-ops] Make `foldl'` call unambiguous.

IsaacG:
```
/solution/test/Tests.hs:80:9: error: [GHC-87543]
    Ambiguous occurrence ‘foldl'’.
    It could refer to
       either ‘Prelude.foldl'’,
              imported from ‘Prelude’ at test/Tests.hs:(8,1)-(16,5)
              (and originally defined in ‘ghc-internal-9.1003.0:GHC.Internal.Data.Foldable’),
           or ‘ListOps.foldl'’,
              imported from ‘ListOps’ at test/Tests.hs:22:7-12.
   |
80 |         foldl' (flip (:)) [] "asdf" `shouldBe` "fdsa"
   |         ^^^^^^
```

### DIFF
--- a/exercises/practice/list-ops/.meta/examples/success-standard/src/ListOps.hs
+++ b/exercises/practice/list-ops/.meta/examples/success-standard/src/ListOps.hs
@@ -4,13 +4,13 @@ module ListOps
   , map
   , filter
   , foldr
-  , ListOps.foldl'
+  , foldl'
   , (++)
   , concat
   ) where
 
 import Prelude hiding
-  ( length, reverse, map, filter, foldr, (++), concat )
+  ( length, reverse, map, filter, foldl', foldr, (++), concat )
 
 foldl' :: (b -> a -> b) -> b -> [a] -> b
 foldl' f = go
@@ -26,10 +26,10 @@ foldr f x0 = go
     go (x : xs) = x `f` go xs
 
 length :: [a] -> Int
-length = ListOps.foldl' (\acc _ -> 1 + acc) 0
+length = foldl' (\acc _ -> 1 + acc) 0
 
 reverse :: [a] -> [a]
-reverse = ListOps.foldl' (flip (:)) []
+reverse = foldl' (flip (:)) []
 
 map :: (a -> b) -> [a] -> [b]
 map f = foldr (\x acc -> f x : acc) []
--- a/exercises/practice/list-ops/.meta/examples/success-standard/src/ListOps.hs
+++ b/exercises/practice/list-ops/.meta/examples/success-standard/src/ListOps.hs
@@ -4,7 +4,7 @@ module ListOps
   , map
   , filter
   , foldr
-  , foldl'
+  , ListOps.foldl'
   , (++)
   , concat
   ) where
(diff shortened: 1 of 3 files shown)


MatthijsBlom (changes requested):
I think a `[no important files changed]` would be fitting.

---

Prior to `base-4.20.0.0` (bundled with GHC 9.10), `foldl'` was not re-exported by the `Prelude`. But from `base-4.20.0.0` onward it is, hence the error that the proposed change fixes.

All other names defined in `ListOps` were already exported by the `Prelude`. This conflict was solved by `hiding` these while importing the `Prelude`. Rather than disambiguating all its uses, add `foldl'` to this import list.

---

The solution stub file should be patched as well.

IsaacG:
> I think a `[no important files changed]` would be fitting.

I completely agree. That needs to be in the merge message when the track maintainer merges this PR.

> The solution stub file should be patched as well.

Done.

Thanks for the explanation and details!

MatthijsBlom:
Both `…/list-ops/.meta/…/ListOps.hs` and `…/list-ops/test/Tests.hs` are good now.

By "the solution stub file" I meant `…/list-ops/src/ListOps.hs`. That one should receive the same patch as the other two.

---

Hiding a nonexistent export gives rise to merely a [GHC-61689 -Wdodgy-imports](https://errors.haskell.org/messages/GHC-61689/) warning, so it looks like this change is backwards-compatible.

---

...Except apparently the tests have `-Werror=dodgy-imports`. I guess as a consequence of a `-Werror=all` or something somewhere.

IsaacG:
> By "the solution stub file" I meant `…/list-ops/src/ListOps.hs`. That one should receive the same patch as the other two.

Woops. Missed that one. Added now.

IsaacG:
Eh. This PR was meant in service of the test runner upgrade. If that lands first, a test rerun ought to pass here.

Unrelated, there's a bunch of other PRs in this repo worth a look, too!

MatthijsBlom:
I think the tests fail because of

https://github.com/exercism/haskell/blob/9e4f3858a403caa42db2e83cf84afc28a18cdbfa/bin/test-example#L58

I guess they could be made to pass again by adding `--ghc-options="-Wwarn=dodgy-imports"` to that `stack` call.
But we could also just wait until the track moves on to use `base-4.20+` (GHC 9.10+).

IsaacG:
@tofische With the test runner updated, it might be worth rerunning the failed check here

IsaacG:
@exercism/haskell  Could you rerun the failed test now that the test runner has been updated?

MatthijsBlom:
The exercises check failure is expected. This PR adds a nonexistent item to import lists. GHC raises a warning about this, and the exercises check elevates all warnings to errors. The warning will go away as soon as the test runner is moved on to use GHC 9.10 / base 4.20.

The configuration check failure confuses me.

tofische:
> The configuration check failure confuses me.

The path doesn't contain the `flower-field` exercise that was added later. @IsaacG rebase and push again, it shall IMHO help.

MatthijsBlom:
Ah, so it is this line that causes `flower-field` to show up: https://github.com/exercism/haskell/blob/35284ffdeb7195e8d61d76914c791189aaa863b8/bin/ensure-practice-exercise-descriptions-are-synced.sh#L18

MatthijsBlom:
GHCup has moved on to recommend GHC 9.10.3. After the track has been moved over to [lts-24.48](https://www.stackage.org/lts-24.48), this PR can be merged.

PR #1303, opened 2026-06-18
verify-exercises-in-docker: avoid cleanup failure

keiravillekode:
Docker runs the test runner as root. 

Previously in a directory like `/tmp/exercism-verify-*-#####/` we were creating a root-owned subdirectory `.stack-work/` that the script failed to cleanup. Now that subdirectory is no longer being created under `/tmp/exercism-verify-bob-#####/`  and cleanup succeeds.


tofische (approved):

keiravillekode:
(I don't have merge privileges.)

PR #1307, opened 2026-07-08
Leap exercise, fix misprint

flinguenheld:
Hello,
Here a misprint fix.


MatthijsBlom:
What made you think this is a mistake? Did you find it confusing?

The original is as intended. The list is part of a sentence:
By happenstance, &bullet; A does X, and &bullet; B does Y.
If people find this construction confusing, maybe I should refrain from using it in the future.

flinguenheld:
Oh I see.
I was confused by the bullet list.
I'm used to read each line separately so it feels that it misses something.
And maybe also because the sentence is about *or* and *and*.

PR #1275, opened 2024-10-05
Add pattern matching and tuple concept

meatball133:

PR #1309, opened 2026-08-09
Clean-up lint errors and test failures.

tofische:
Correct problems after merging #1273 and #1275

### DIFF
--- a/config.json
+++ b/config.json
@@ -1434,6 +1434,11 @@
       "name": "Booleans"
     },
     {
+      "uuid": "9f819455-c3bd-4957-93a9-da70cd29afe5",
+      "slug": "guards",
+      "name": "Guards"
+    },
+    {
       "uuid": "ef484ab5-d4d2-4d7c-b4b5-abb5920f916a",
       "slug": "lists",
       "name": "Lists"
--- a/exercises/concept/guessing-game/.meta/exemplar/src/GuessingGame.hs
+++ b/exercises/concept/guessing-game/.meta/exemplar/src/GuessingGame.hs
@@ -3,7 +3,7 @@ module GuessingGame (reply) where
 reply :: Int -> Int -> String
 reply n guess
   | guess == n = "Correct"
-  | guess + 1 == n || guess - 1 == n = "So close!"
+  | guess + 1 == n || guess - 1 == n = "So close"
   | guess < n = "Too low"
   | guess > n = "Too high"
   | otherwise = "Incorrect" -- catch-all clause to make pattern matching exhaustive
--- a/exercises/concept/guessing-game/.meta/exemplar/src/GuessingGame.hs
+++ b/exercises/concept/guessing-game/.meta/exemplar/src/GuessingGame.hs
@@ -6,4 +6,4 @@ reply n guess
   | guess + 1 == n || guess - 1 == n = "So close!"
   | guess < n = "Too low"
   | guess > n = "Too high"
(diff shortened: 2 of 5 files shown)


tofische:
Test fails on lucians-luscious-lasagna, but this stub contains the file .meta/STUB-ALLOWED-TO-NOT-COMPILE - why?

@IsaacG could you have a look?

IsaacG:
> Test fails on lucians-luscious-lasagna, but this stub contains the file .meta/STUB-ALLOWED-TO-NOT-COMPILE - why?
> 
> @IsaacG could you have a look?

The test shows,

> testing /home/runner/work/haskell/haskell/exercises/concept/kitchen-calculator/.meta/exemplar - should succeed
> lucians-luscious-lasagna> configure (lib + test)
> Configuring lucians-luscious-lasagna-1.0.0.0...
> lucians-luscious-lasagna> build (lib + test) with ghc-9.6.7
> Preprocessing library for lucians-luscious-lasagna-1.0.0.0..
> Error: Cabal-simple_w2MFVN35_3.10.3.0_ghc-9.6.7: can't find source for
> LuciansLusciousLasagna in src,
> .stack-work/dist/x86_64-linux-tinfo6/ghc-9.6.7/build/autogen,
> .stack-work/dist/x86_64-linux-tinfo6/ghc-9.6.7/build/global-autogen
> 
> 
> Error: [S-7282]
>        Stack failed to execute the build plan.

The exemplar solution should always work. It looks like the exemplar can't compile for some reason. Are you expecting the exemplar to compile?

IsaacG:
Weird. That exemplar passes just fine when I test it ... but others fail. Maybe double check it's not transient by rerunning the check.

IsaacG:
```bash
# Test the exemples/exemplars:

~/exercism/haskell » docker run -v $PWD:/haskell -w /haskell --entrypoint bash exercism/haskell-test-runner -c 'for i in ./exercises/concept/* ./exercises/practice/* ; do ./bin/test-all-examples "$i" ; done'
```

Failures:

```
» grep -e 'Test suite test fa' -e 'While building package' [...]
       While building package guessing-game-1.0.0.0 (scroll up to its section to see the error) using:
accumulate> Test suite test failed
       While building package leap-0.0.0 (scroll up to its section to see the error) using:
robot-name> Test suite test failed
strain> Test suite test failed
```

tofische:
Yes, exemplar shall always compile - and locally it also does.

The error message says:

`testing /home/runner/work/haskell/haskell/exercises/concept/kitchen-calculator/.meta/exemplar - should succeed`

but then a couple of lines later

`can't find source for LuciansLusciousLasagna in src`

as if it tried to use sources from `src` and not from `.meta/exemplar/src`.

It looks like `.meta/STUB-ALLOWED-TO-NOT-COMPILE` isn't completely respected, but I'd appreciate if somebody who really understands the pipeline have a look.

IsaacG:
I wasn't involved in any part of the pipeline building and don't have a lot of Haskell track context, unfortunately. I'm not a Haskell track maintainer.

I do see the test runner uses Haskell 9.10.3 on Debian while [the GHA](https://github.com/exercism/haskell/blob/main/.github/workflows/tests.yml) is using ghc-9.6.7 on Ubuntu 24.04.

One thing I've seen others do is use the Docker test runner image to run the GHA tests, reducing the number of versions/platforms the code needs to work across. That might be an option worth exploring here.

